PR #10, opened 2019-11-18
Watch OSM contracts and transform LogValue events

Gslaughl:


transformers/events/log_value/converter.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (c Converter) ToModels(abi string, logs []core.HeaderSyncLog) ([]event.InsertionModel, error) {
+	entities, entityErr := c.toEntities(abi, logs)
+	if entityErr != nil {
+		return nil, fmt.Errorf("converter couldn't convert logs to entities: %v", entityErr)
+	}
+
+	var models []event.InsertionModel
+	for _, logValueEntity := range entities {
+		bigIntVal := new(big.Int).SetBytes(logValueEntity.Val[:]) // TODO: verify this is a good type
```
Gslaughl: I wasn't quite sure what type to make the `val` value. It's not obvious from the contract what it's being used for, but since the OSM contracts are oracles (presumably about collateral evaluations), I figured an int would be a good choice, and the resulting `bigint` values don't seem too objectionable.

m0ar: Contract says `val uint128` so `big.Int` is fine :)

environments/docker.toml
```diff
@@ -273,6 +274,13 @@
         migrations = "db/migrations"
         contracts = ["CDP_MANAGER"]
         rank = "0"
+    [exporter.log_value]
+        path = "transformers/events/log_value/initializer"
+        type = "eth_event"
+        repository = "github.com/makerdao/vdb-mcd-transformers"
+        migrations = "db/migrations"
+        contracts = ["OSM_ETH", "OSM_REP", "OSM_ZRX", "OSM_OMG", "OSM_BAT", "OSM_DGD", "OSM_GNT"]
```
elizabethengelman: Do these contracts correspond to the **PIP**\_* addresses or the **VAL**\_ *addresses in the [changelog](https://changelog.makerdao.com/releases/kovan/0.2.15/contracts.json)? They're mostly the same, except that there is a PIP_SAI address, but not a VAL_SAI one, and I'm wondering if we should include one for Sai.

Gslaughl: Good question-- it took me a while to realize that the addresses I wanted were even in the changelog, so I actually got them by looking in etherscan and just grabbing every `pip_` value that was passed to the [`Spotter.file`](https://github.com/makerdao/dss/blob/master/src/spot.sol#L77) method. So I guess these addresses are more closely linked with the PIP_* addresses.

I was also curious about the Sai situation, but in the changelog the `PIP_SAI` address is `0x0000000000000000000000000000000000000001` (not a real address ofc), and interestingly that's also what's passed to `Spotter.file`. So I assume there's not really an oracle in the system for Sai.

environments/docker.toml
```diff
@@ -501,3 +509,31 @@
...
         deployed = 14374540
+    [contract.OSM_ETH]
+        address  = "0x75dd74e8afe8110c8320ed397cccff3b8134d981"
```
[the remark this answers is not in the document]
Gslaughl: Conveniently they don't change between versions, although all of them besides ETH and BAT are removed from the 0.2.16 changelog. It sounds like we might want to remove all the collateral types besides those 2 anyway.

transformers/shared/constants/signature_test.go
```diff
@@ -86,6 +86,10 @@ var _ = Describe("Signature constants", func() {
 		Expect(NewCdpSignature()).To(Equal("0xd6be0bc178658a382ff4f91c8c68b542aa6b71685b8fe427966b87745c3ea7a2"))
 	})
 
+	It("generates log note signature", func() {
```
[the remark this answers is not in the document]
Gslaughl: Yeah, nice catch! So used to working with `LogNote` events that I kept typing that on accident :P

environments/docker.toml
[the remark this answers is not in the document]
rmulhol: Still true for 0.2.17?

environments/docker.toml
rmulhol: Yeah iirc we just assume sai/dai always == $1. Not sure why an address would be declared for its spotter, though, if that's the case... 🤔

environments/docker.toml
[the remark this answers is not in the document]
Gslaughl: Yep :D

transformers/events/log_value/converter.go
rmulhol: This is true but in general I'm a fan of using https://github.com/makerdao/vdb-mcd-transformers/blob/staging/transformers/shared/utilities.go#L67 or https://github.com/makerdao/vdb-mcd-transformers/blob/staging/transformers/shared/utilities.go#L62 depending on whether it's an `int` or `uint`, since that distinction can cause sneaky 🐛

Gslaughl: Thanks 🤦‍♂ I always forget we have these nice conversion utils!

PR #11, opened 2019-11-19
Fix imports

m0ar:


scripts/reset_db
```diff
@@ -11,7 +11,7 @@ fi
 
 db=$1
 dir=$(basename "$(pwd)")
-if [ $dir != "mcd_transformers" ]
+if [ $dir != "vdb-mcd-transformers" ]
```
elizabethengelman: nice catch!

m0ar: :female_detective:

PR #14, opened 2019-11-20
VDB-975 Move dent

m0ar:


transformers/events/dent/converter.go
```diff
@@ -20,23 +20,36 @@ import (
...
 
-type DentConverter struct{}
+type Converter struct {
+	db *postgres.DB
+}
 
 const (
-	logDataRequired   = true
-	numTopicsRequired = 4
+	logDataRequired                    = true
```
yaoandrew: :100:

transformers/test_data/dent.go
```diff
@@ -60,20 +61,18 @@ var DentHeaderSyncLog = core.HeaderSyncLog{
...
+	ColumnValues: event.ColumnValues{
+		dent.Id:            dentBidId,
+		dent.Lot:           dentLot,
+		dent.Bid:           dentBid,
 		constants.HeaderFK: DentHeaderSyncLog.HeaderID,
 		constants.LogFK:    DentHeaderSyncLog.ID,
-	},
-	ForeignKeyValues: shared.ForeignKeyValues{
-		constants.AddressFK: DentHeaderSyncLog.Log.Address.Hex(),
+		// constants.AddressColumn
```
yaoandrew: Do we need this?

transformers/test_data/dent.go
```diff
@@ -60,20 +61,18 @@ var DentHeaderSyncLog = core.HeaderSyncLog{
...
 	},
-	ColumnValues: shared.ColumnValues{
-		"bid_id":           dentBidId,
-		"lot":              dentLot,
-		"bid":              dentBid,
+	ColumnValues: event.ColumnValues{
+		dent.Id:            dentBidId,
+		dent.Lot:           dentLot,
+		dent.Bid:           dentBid,
 		constants.HeaderFK: DentHeaderSyncLog.HeaderID,
```
yaoandrew: Not sure when to use one vs the other or if we need both but could we potentially use `event.HeaderFK` from `vulcanizedb/libraries/shared/factories`

transformers/events/dent/repository.go
```diff
@@ -17,19 +17,18 @@
...
 
 import (
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
 	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
-
-	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
 )
 
-type DentRepository struct {
+type Repository struct {
```
rmulhol: Why are we keeping a transformer specific repository at all? Seems like we could just initialize the shared `factories/transformer` with a database connection and pass that directly to `event.Create`. Then we could remove the `Repository` from the initializer entirely, right?

transformers/test_data/dent.go
m0ar: Sanity preservation, it should be there but we can't assert against it since it depends on the DB state when the tests run :)

transformers/events/dent/repository.go
m0ar: yep.

m0ar: Available for whoever wants to shepherd the dual PRs :) We just decided to do it in steps a while back, moving stuff over, then removing the repos.

PR #9, opened 2019-11-14
VDB-1020 - 2.16 release

elizabethengelman:
- note: `MCD_FLIP_ETH_B`, `MCD_FLIP_ETH_C`, `MCD_FLIP_REP_A`, `MCD_FLIP_ZRX_A`, `MCD_FLIP_OMG_A`, `MCD_FLIP_DGD_A` and `MCD_FLIP_GNT_A` weren't included in the addresses for 2.16, so I remove them from the regular and docker config files, but left them in `testing.toml` as a way to document what we have. I'd be happy to remove them from there as well if folks thing it's cleaner.

PR #16, opened 2019-11-20
Update vDB dep to v0.0.10

rmulhol:


rmulhol:
Closing because this will be addressed on #9

PR #25, opened 2019-11-25
VDB-966 terminate container

elizabethengelman:
- terminates the container if one of the processes exits for some reason
- we probably want to switch to having a different container for each process at some point, but i think this may be helpful in the short term
- also added some error messaging/refactored testing if required arguments have been passed to docker run

### DIFF
--- a/dockerfiles/startup_script.sh
+++ b/dockerfiles/startup_script.sh
@@ -37,11 +37,29 @@ then
 fi
 
 
+echo "Starting headerSync and executing the transformers..."
 # Fire up the services
 if [ $? -eq 0 ]; then
   # Fire up the services
   ./vulcanizedb headerSync --config config.toml -s $STARTING_BLOCK_NUMBER &
   ./vulcanizedb execute --config config.toml &
 fi
+
+
+# Check every 60 seconds to see if either process has excited.
+# If grepping for process names finds something, they exit with 0 status. If they are not both 0, then one of the processes has already excited.
+
+while sleep 60; do
+  ps aux | grep headerSync | grep -q -v grep
+  HEADER_SYNC_STATUS=$?
+
+  ps aux | grep execute | grep -q -v grep
+  EXECUTE_STATUS=$?
+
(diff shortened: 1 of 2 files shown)


dockerfiles/startup_script.sh
```diff
@@ -37,11 +37,27 @@ then
...
   ./vulcanizedb headerSync --config config.toml -s $STARTING_BLOCK_NUMBER &
   ./vulcanizedb execute --config config.toml &
 fi
 
-wait
+
+# Check every 60 seconds to see if either process has excited.
+# If grepping for process names finds something, they exit with 0 status. If they are not both 0, then one of the processes has already excited.
+
+while sleep 60; do
```
elizabethengelman: one thing to note is that when stopping the docker container with `CTRL+C`, it seems to finish the current sleep, and won't actually terminate for another 60 seconds. This doesn't seem ideal. I suppose we could pass in a sleep time, so that we can set it to 1 for development purposes. I'd love to hear other thoughts about this.

m0ar: I mean, this is so lightweight anyway that checking every 10-15 seconds would be fine too, and eliminate that problem.

The long term solution to this is IMO to figure out how to call vDB with a _set_ of commands and do it's own concurrency with goroutines, since the current system is bad practice anyway.. :)

rmulhol: Agree that this is no problem and potentially cutting down the interval could make sense.

I think enabling vdb to be invoked with a set of commands is a cool idea but I'd also be in favor of isolating each command in its own container (and breaking up `execute` so that we can have separate containers extracting logs vs extracting diffs vs transforming logs vs transforming diffs). Thinking that should make it easier to restart an isolated process that fails without needing to restart everything (rather than the current approach which means we have to restart header sync due to an unrelated fatal error in the event transformer execution)

elizabethengelman: cool, thanks for the feedback guys! i'll drop the checking interval to 10 seconds. 

and yep, i definitely agree that we need to rework how we're kicking off vdb. I was also thinking along the lines of breaking things up into different containers. but i like the idea of breaking our current commands down even further as well. 👍

PR #26, opened 2019-11-25
(VDB-969) Add triggers for re-org safety in ilk history table

Gslaughl:
When a diff record is deleted, these triggers wipe out that diff's value from the `historical_ilk_state` table.

### DIFF
--- a/db/migrations/20190919141224_create_ilk_state_history_table_and_triggers.sql
+++ b/db/migrations/20190919141224_create_ilk_state_history_table_and_triggers.sql
@@ -26,12 +26,14 @@ COMMENT ON COLUMN api.historical_ilk_state.ilk_identifier IS '@name id';
 CREATE FUNCTION ilk_rate_before_block(ilk_id INTEGER, header_id INTEGER) RETURNS NUMERIC AS
 $$
 WITH passed_block_number AS (
-    SELECT block_number FROM public.headers WHERE id = header_id
+    SELECT block_number
+    FROM public.headers
+    WHERE id = header_id
 )
 
 SELECT rate
 FROM maker.vat_ilk_rate
-    LEFT JOIN public.headers ON vat_ilk_rate.header_id = headers.id
+         LEFT JOIN public.headers ON vat_ilk_rate.header_id = headers.id
 WHERE vat_ilk_rate.ilk_id = ilk_rate_before_block.ilk_id
   AND headers.block_number < (SELECT block_number FROM passed_block_number)
 ORDER BY block_number DESC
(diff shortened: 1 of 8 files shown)


db/migrations/20190919141224_create_ilk_state_history_table_and_triggers.sql
```diff
@@ -176,6 +176,52 @@ WHERE vat_init.ilk_id = ilk_time_created.ilk_id
 $$
     LANGUAGE sql;
 
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION maker.delete_redundant_ilk_state(ilk_id INTEGER, block_number BIGINT) RETURNS api.historical_ilk_state
```
Gslaughl: This function is responsible for determining whether a record in `historical_ilk_state` is the same as the one before it, and deleting it if so. It's necessary because a single record in `historical_ilk_state` can have data from multiple storage tables, so we can't know whether it's safe to delete a row until we've verified that the only new data it held was from a re-orged diff.

If we can guarantee that re-orged diffs will always be deleted before new diffs are added, we can probably get rid of this function.

db/migrations/20190919141224_create_ilk_state_history_table_and_triggers.sql
```diff
@@ -247,23 +293,28 @@ $$
...
 BEGIN
-    PERFORM maker.insert_new_rate(NEW);
-    PERFORM maker.update_later_rates(NEW);
+    IF (TG_OP IN ('INSERT', 'UPDATE')) THEN
+        PERFORM maker.insert_new_rate(NEW);
+        PERFORM maker.update_rates_until_next_diff(NEW, NEW.rate);
+    ELSIF (TG_OP = 'DELETE') THEN
+        PERFORM maker.update_rates_until_next_diff(OLD, ilk_rate_before_block(OLD.ilk_id, OLD.block_number));
+        PERFORM maker.delete_redundant_ilk_state(OLD.ilk_id, OLD.block_number);
+    END IF;
```
Gslaughl: Not sure what the best practices are for triggers (i.e. whether it's better to have multiple triggers on the same table or use conditional logic like this), but this seemed preferable to me. Happy to change if we want to.

db/migrations/20190919141224_create_ilk_state_history_table_and_triggers.sql
```diff
@@ -26,12 +26,14 @@ COMMENT ON COLUMN api.historical_ilk_state.ilk_identifier IS '@name id';
 CREATE FUNCTION ilk_rate_before_block(ilk_id INTEGER, header_id INTEGER) RETURNS NUMERIC AS
 $$
 WITH passed_block_number AS (
-    SELECT block_number FROM public.headers WHERE id = header_id
```
Gslaughl: In hindsight I probably should've put the formatting changes into a separate commit :-/

PR #28, opened 2019-11-27
(VDB-1036) Add Pot File Vow event transformer

rmulhol:


transformers/events/pot_file/vow/repository.go
```diff
@@ -0,0 +1,18 @@
...
+	db *postgres.DB
+}
+
+func (repository Repository) Create(models []event.InsertionModel) error {
+	return event.Create(models, repository.db)
+}
+
+func (repository *Repository) SetDB(db *postgres.DB) {
+	repository.db = db
+}
```
m0ar: @rmulhol Does this really work after the removal of the transformer-specific repository from vdb core? :confused:

PR #29, opened 2019-11-27
VDB-923 Remove repo

m0ar:
:scissors: :tada:


db/schema.sql
```diff
@@ -502,7 +502,6 @@ CREATE FUNCTION api.all_bites(ilk_identifier text, max_results integer DEFAULT '
     LANGUAGE sql STABLE STRICT
```
rmulhol: Mind popping this off of this PR? I think for some reason our schema formatters are not in agreement about whether these spaces should be here, so we'll probably be going back and forth regularly. But at least maybe we can limit that to cases where we're also making a substantive change to the schema

PR #30, opened 2019-11-27
(VDB-1037) Add Pot Cage event transformer

Gslaughl:


db/migrations/20191127095017_create_pot_cage.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
```
Gslaughl: There are no cage events in kovan, which I think makes sense since it's only invoked during global settlement. But as a result, there aren't any tests that require the existence of this table in the db. Might be worth brainstorming a way to test-drive db changes without having an event in kovan

Gslaughl:
@rmulhol will do, many thanks!

PR #33, opened 2019-12-03
VDB-1002: use VDB InsertionModel for vow file

elizabethengelman:


transformers/shared/constants/label.go
```diff
@@ -16,6 +16,12 @@
 
 package constants
 
+const (
```
m0ar: nice

transformers/test_data/vow_file.go
```diff
@@ -51,17 +52,16 @@ var VowFileHeaderSyncLog = core.HeaderSyncLog{
...
-var VowFileModel = shared.InsertionModel{
-	SchemaName: "maker",
-	TableName:  "vow_file",
-	OrderedColumns: []string{
-		constants.HeaderFK, "what", "data", constants.LogFK,
+var VowFileModel = event.InsertionModel{
+	SchemaName: constants.MakerSchema,
+	TableName:  constants.VowFileLabel,
+	OrderedColumns: []event.ColumnName{
+		constants.HeaderFK, constants.WhatColumn, constants.DataColumn, constants.LogFK,
```
m0ar: Should use the constants from `event` in core for non domain specific things :)
```go
const LogFK ColumnName = "log_id"
const AddressFK ColumnName = "address_id"
const HeaderFK ColumnName = "header_id"
```

They still exist in both repos to allow for step-by-step migration of the transformers. No biggie either, we'll notice breakage when the `vdb-mcd-transformers` code is nuked anyway :tada:

transformers/test_data/vow_file.go
```diff
@@ -51,17 +52,16 @@ var VowFileHeaderSyncLog = core.HeaderSyncLog{
...
 	},
-	ColumnValues: shared.ColumnValues{
-		"what":             "wait",
-		"data":             "100000000000000000000000",
-		constants.HeaderFK: VowFileHeaderSyncLog.HeaderID,
-		constants.LogFK:    VowFileHeaderSyncLog.ID,
+	ColumnValues: event.ColumnValues{
+		constants.WhatColumn: "wait",
```
m0ar: I'm actually more inclined to declaring these locally in all transformers, to make them as self-sufficient as possible. What's your take? :thinking:

elizabethengelman: 🤔 My thought was to reduce duplication, since in this case we'd end up defining a WhatColumn constant in several converters. I understand the instinct to make the transformers self-sufficient, but I'm not immediately seeing a downside to using the constants package in these, since we're also including the shared package as well. That said, I don't feel super strongly about it either way, so I'm up for changing it back if folks think that's best.

rmulhol: I'm fine with either approach, but I think I would lean toward having shared constants for terminology that's reused across contracts in the dss repo. My thinking is that even if we break down the code in this repo further along any number of abstractions, we'd probably keep event transformers targeting dss together - so we might as well take advantage of its naming scheme.

transformers/events/vow_file/converter.go
```diff
@@ -19,41 +19,38 @@ package vow_file
...
-	numTopicsRequired = 4
-)
-
-func (VowFileConverter) ToModels(_ string, logs []core.HeaderSyncLog) ([]shared.InsertionModel, error) {
-	var models []shared.InsertionModel
+func (Converter) ToModels(_ string, logs []core.HeaderSyncLog, _ *postgres.DB) ([]event.InsertionModel, error) {
+	var models []event.InsertionModel
 	for _, log := range logs {
-		err := shared.VerifyLog(log.Log, numTopicsRequired, logDataRequired)
+		err := shared.VerifyLog(log.Log, shared.FourTopicsRequired, shared.LogDataRequired)
```
rmulhol: Doesn't need to happen in this PR, but I'm wondering whether `VerifyLog` and the constants for fields required should potentially be moved over to vulcanizedb

transformers/events/vow_flog/converter.go
```diff
@@ -40,7 +40,7 @@ func (VowFlogConverter) ToModels(_ string, logs []core.HeaderSyncLog) ([]shared.
 		era := log.Log.Topics[2].Big()
 
 		model := shared.InsertionModel{
-			SchemaName: "maker",
+			SchemaName: constants.MakerSchema,
 			TableName:  "vow_flog",
```
rmulhol: Also doesn't need to happen in this PR, but could be cool to reuse the labels from the constants for all of the table names. Perhaps even renaming `Label` to `Table`

PR #35, opened 2019-12-04
Mainnet 1.0.1

rmulhol:
We should be able to build an image from here to run vdb-lite against a fresh db with transactions for logs.

Main blockers:
- `headerSync` needs to continue running even if `execute` fails, otherwise we don't replace the invalid header unless it's still within the validation window (most recent 15 blocks) when we restart. This is prevented by our current setup, which kills `headerSync` if `execute` fails.
- may want to change the `vat_fold` table to persist an `ilk_id` and `u`, rather than an `urn_id`
- may want to cut a release in vdb so that we aren't pointing at a branch (`staging`) that may change before future builds

Note: probably don't want to merge this until we've cut Kovan 0.2.17 (and perhaps Mainnet 1.0.0 🤔), but we can build images from here in the meantime


m0ar:
What did we want to resolve by making `execute` and `headerSync` dependent on each other to begin with? (ping @elizabethengelman )

rmulhol:
The idea is to facilitate automated restarts by shutting everything down if one process is down. I think we do want that behavior anytime we're running multiple processes in a container, but I think `headerSync` probably deserves its own dedicated container.

m0ar:
@rmulhol Would you mind doing a rundown of the remaining blockers here?

rmulhol:
@m0ar core things I'd like to see taken care of before this is merged:
- merge PRs that fix bugs in the existing vdb-lite setup (e.g. retry on error, separate containers with automatic restarts, etc)
- cut 0.2.17 release

nice to haves:
- new PR to inline addresses/abis/deployment blocks so that reading config is less likely to cause errors
- modify `vat_fold` to enable querying by ilk

m0ar:
@rmulhol Hrm, after some more thinking, only inlining of ABIs make sense, since addresses/deployment blocks change depending on chain. Testnet, mainnet, local, etc. We'll be shooting ourselves in the foot with this I think :( 

I'd be happy to pair up to figure the config reading shit out though!

m0ar:
@rmulhol Took care of `vat_fold` :ok_man:

rmulhol:
Closing in favor of https://github.com/makerdao/vdb-lite-mcd-transformers/pull/1

PR #37, opened 2019-12-05
remove repo from vat_suck

yaoandrew:


transformers/shared/constants/keys.go
```diff
@@ -54,6 +54,9 @@ const (
 	DataColumn    event.ColumnName = "data"
 	InkColumn     event.ColumnName = "ink"
 	ArtColumn     event.ColumnName = "art"
+	RadColumn     event.ColumnName = "rad"
 	TabColumn     event.ColumnName = "tab"
+	UColumn       event.ColumnName = "u"
```
yaoandrew: @elizabethengelman Are we adding these here now? Was looking at your PR...

transformers/test_data/vat_suck.go
```diff
@@ -53,18 +53,17 @@ var VatSuckHeaderSyncLog = core.HeaderSyncLog{
 	Transformed: false,
 }
 
-var VatSuckModel = shared.InsertionModel{
+var VatSuckModel = event.InsertionModel{
 	SchemaName: "maker",
```
yaoandrew: @elizabethengelman I'll rebase when your PR with the new constants gets merged

transformers/events/vat_suck/converter.go
```diff
@@ -20,18 +20,20 @@ import (
...
 	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
 )
 
-type VatSuckConverter struct{}
+type Converter struct{}
 
 const (
 	logDataRequired   = false
 	numTopicsRequired = 4
```
elizabethengelman: There are some shared constants for these vars in https://github.com/makerdao/vdb-mcd-transformers/blob/staging/transformers/shared/converter.go#L28.

transformers/shared/constants/keys.go
elizabethengelman: Yep!

PR #38, opened 2019-12-05
(VDB-970) urn state history

Gslaughl:
Creates a trigger-table exposing all historical states of all urns. Automatically responds to data coming in out of order as well as data being deleted.

### DIFF
--- a/db/migrations/20191202091203_create_historical_urn_state_table_and_triggers.sql
+++ b/db/migrations/20191202091203_create_historical_urn_state_table_and_triggers.sql
@@ -1,0 +1,318 @@
+-- +goose Up
+CREATE TABLE api.historical_urn_state
+(
+    urn_identifier TEXT,
+    ilk_identifier TEXT,
+    block_height   BIGINT,
+    ink            NUMERIC   DEFAULT NULL,
+    art            NUMERIC   DEFAULT NULL,
+    created        TIMESTAMP DEFAULT NULL,
+    updated        TIMESTAMP NOT NULL,
+    PRIMARY KEY (urn_identifier, ilk_identifier, block_height)
+);
+
+
+CREATE FUNCTION urn_ink_before_block(urn_id INTEGER, header_id INTEGER) RETURNS NUMERIC AS
+$$
+WITH passed_block_number AS (
+    SELECT block_number
+    FROM public.headers
+    WHERE id = header_id)
+SELECT ink
+FROM maker.vat_urn_ink
+         LEFT JOIN public.headers ON vat_urn_ink.header_id = headers.id
+WHERE vat_urn_ink.urn_id = urn_ink_before_block.urn_id
+  AND headers.block_number < (SELECT block_number FROM passed_block_number)
+ORDER BY block_number DESC
+LIMIT 1
+$$
+    LANGUAGE sql;
+
+
+CREATE FUNCTION urn_art_before_block(urn_id INTEGER, header_id INTEGER) RETURNS NUMERIC AS
+$$
+WITH passed_block_number AS (
+    SELECT block_number
+    FROM public.headers
+    WHERE id = header_id)
+SELECT art
+FROM maker.vat_urn_art
+         LEFT JOIN public.headers ON vat_urn_art.header_id = headers.id
+WHERE vat_urn_art.urn_id = urn_art_before_block.urn_id
+  AND headers.block_number < (SELECT block_number FROM passed_block_number)
+ORDER BY block_number DESC
+LIMIT 1
+$$
+    LANGUAGE sql;
+
+
+CREATE FUNCTION urn_time_created(urn_id INTEGER) RETURNS TIMESTAMP AS
+$$
+SELECT api.epoch_to_datetime(MIN(block_timestamp))
+FROM maker.vat_urn_ink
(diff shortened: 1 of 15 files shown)


transformers/storage/vat/repository_test.go
```diff
@@ -37,15 +39,14 @@ import (
 
 var _ = Describe("Vat storage repository", func() {
 	var (
-		db           *postgres.DB
+		db           = test_config.NewTestDB(test_config.NewTestNode())
```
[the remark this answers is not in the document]
rmulhol: Nice! That seems like a change worth making everywhere, though not necessarily as part of this PR

db/migrations/20191203104431_create_historical_urn_state_computed_columns.sql
```diff
@@ -0,0 +1,46 @@
+-- +goose Up
+-- SQL in this section is executed when the migration is applied.
+
+CREATE FUNCTION api.historical_urn_state_ilk(state api.historical_urn_state)
+    RETURNS api.ilk_state AS
+$$
+SELECT *
+FROM api.get_ilk(state.ilk_identifier, state.block_height)
```
m0ar: What's the complexity of this call? And the rest really...

Might be a trade-off to be done here, fast access to the "native" data, but having to do a second query to grab the `ilk`/`frobs`/`bites` for a certain block... Not sure actually.

m0ar: But my point is that if these slow down the queries by a lot, maybe it's not worthwhile to include them even...

transformers/component_tests/queries/all_urn_states_query_test.go
```diff
@@ -46,12 +46,12 @@ var _ = Describe("Urn history query", func() {
 	})
 
 	It("returns a reverse chronological history for the given ilk and urn", func() {
-		urnSetupData := helper.GetUrnSetupData(headerOne)
+		urnSetupData := helper.GetUrnSetupData()
 		urnMetadata := helper.GetUrnMetadata(helper.FakeIlk.Hex, fakeUrn)
-		helper.CreateUrn(urnSetupData, urnMetadata, vatRepo)
+		helper.CreateUrn(urnSetupData, headerOne.Id, urnMetadata, vatRepo)
```
m0ar: If you cba I think golint prefers `headerOne.ID` ;)

transformers/storage/vat/repository_test.go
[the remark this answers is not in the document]
m0ar: Created VDB-1058 for that

db/migrations/20191202091203_create_historical_urn_state_table_and_triggers.sql
```diff
@@ -0,0 +1,319 @@
...
+SELECT api.epoch_to_datetime(MIN(block_timestamp))
+FROM maker.vat_urn_ink
+         LEFT JOIN public.headers ON vat_urn_ink.header_id = headers.id
+WHERE vat_urn_ink.urn_id = urn_time_created.urn_id
+$$
+    LANGUAGE sql;
+
+
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION maker.delete_redundant_urn_state(urn_id INTEGER, header_id INTEGER) RETURNS api.historical_urn_state
```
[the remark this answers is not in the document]
Gslaughl: Haha yeah for sure 😅 

It sounds like you got the gist of it, but it just deletes a `historical_urn_state` row if it's identical to the previous one. It's necessary because 1 row can have data from both an `ink` diff and an `art` diff, so deleting 1 kind of diff doesn't mean you can delete the associated row in this table. First we have to check that the row doesn't contain data from any other diffs, and only then can we delete it.

I agree that it's not very clear what's happening. I can at least add a comment and maybe find a clearer way to accomplish the same thing.

db/migrations/20191202091203_create_historical_urn_state_table_and_triggers.sql
```diff
@@ -0,0 +1,319 @@
...
+                 LEFT JOIN api.historical_urn_state
+                           ON urn_identifier = urns.identifier AND ilk_identifier = ilks.identifier
+        WHERE urns.id = urn_id
+          AND historical_urn_state.block_height < urn_state_block_number
+        ORDER BY historical_urn_state.block_height DESC
+        LIMIT 1);
+BEGIN
+    DELETE
+    FROM api.historical_urn_state
+        USING maker.urns LEFT JOIN maker.ilks ON urns.ilk_id = ilks.id
```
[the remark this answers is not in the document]
Gslaughl: `USING` is basically just a read-only `JOIN ` for delete statements.

db/migrations/20191203104431_create_historical_urn_state_computed_columns.sql
Gslaughl: Yeah that's a good point. I just included these computed columns because they're included in the original `all_urn_states` query, and I wanted to match that functionality. It's probably worth doing some performance analysis on these columns, bc just from looking at the `get_ilk` query it seems pretty complex.

m0ar: Yeah, `get_ilk` is what usually locks the endpoint completely, so it's gonna nuke this query too :)

PR #45, opened 2019-12-06
(VDB-1045) Isolate execute container

rmulhol:
- Remove headerSync so it can run in a separate container

This includes a change to point at vdb staging and makes associated changes to the deps, which is necessary to use the new storage repository interface. I'd be fine with cutting a new version of vdb whenever we're ready (potentially awaiting open PRs over there) and then updating things here accordingly.

When I run this locally with a config file that includes storage transformers but without a valid path to the CSV, "Waiting for  to appear" spams the logs. Thinking that shouldn't be an issue when actually running this because you'd either include a valid path or turn off the storage transformers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,18 @@
 FROM golang:alpine as builder
 
 RUN apk --update --no-cache add make git g++ linux-headers
-# DEBUG
-RUN apk add busybox-extras
 
-ENV GO111MODULE on
-
-# Build statically linked vDB binary (wonky path because of Dep)
-WORKDIR /go/src/github.com/makerdao/vdb-mcd-transformers
-ADD . .
+# Build migration tool
+WORKDIR /go
+RUN GO111MODULE=auto go get -u -d github.com/pressly/goose/cmd/goose
+WORKDIR /go/src/github.com/pressly/goose/cmd/goose
+RUN GO111MODULE=auto go build -a -ldflags '-s' -tags='no_mysql no_sqlite' -o goose
 
 ARG VDB_VERSION=staging
+ENV GO111MODULE on
+
+WORKDIR /go/src/github.com/makerdao/vdb-mcd-transformers
+ADD . .
 
 WORKDIR /go/src/github.com/makerdao
 RUN git clone https://github.com/makerdao/vulcanizedb.git
@@ -20,20 +22,12 @@ RUN go build
 
 # build mcd with local vdb
 WORKDIR /go/src/github.com/makerdao/vdb-mcd-transformers
-
 RUN go mod edit -replace=github.com/makerdao/vulcanizedb=/go/src/github.com/makerdao/vulcanizedb/
 RUN make plugin PACKAGE=github.com/makerdao/vdb-mcd-transformers
-
-# Build migration tool
-WORKDIR /go
-RUN GO111MODULE=auto go get -u -d github.com/pressly/goose/cmd/goose
-WORKDIR /go/src/github.com/pressly/goose/cmd/goose
-RUN GO111MODULE=auto go build -a -ldflags '-s' -tags='no_mysql no_sqlite' -o goose
 
 
 # app container
 FROM golang:alpine
-# workdir needs to match gopath for building file to correct path
 WORKDIR /go/src/github.com/makerdao/vulcanizedb
 
 # add certificates for node requests via https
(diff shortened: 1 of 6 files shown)


plugins/transformerExporter.go
```diff
@@ -58,56 +61,5 @@ type exporter string
...
..., spot_file_mat.EventTransformerInitializer, spot_poke.EventTransformerInitializer, tick.EventTransformerInitializer, vat_frob.EventTransformerInitializer, vat_flux.EventTransformerInitializer, vat_fold.EventTransformerInitializer, cat_file_chop_lump.EventTransformerInitializer, spot_file_pip.EventTransformerInitializer, tend.EventTransformerInitializer, vat_init.EventTransformerInitializer, vat_slip.EventTransformerInitializer, flip_kick.EventTransformerInitializer, jug_file_ilk.EventTransformerInitializer, pot_file_vow.EventTransformerInitializer, vat_file_debt_ceiling.EventTransformerInitializer, vow_flog.EventTransformerInitializer, jug_init.EventTransformerInitializer, vat_suck.EventTransformerInitializer}, []interface1.StorageTransformerInitializer{cdp_manager.StorageTransformerInitializer, flop_storage.StorageTransformerInitializer, spot.StorageTransformerInitializer, jug.StorageTransformerInitializer, vow.StorageTransformerInitializer, bat_flip.StorageTransformerInitializer, eth_flip_a.StorageTransformerInitializer, vat.StorageTransformerInitializer, cat.StorageTransformerInitializer, flap_storage.StorageTransformerInitializer}, []interface1.ContractTransformerInitializer{}
```
m0ar: ugh?

environments/docker.toml
```diff
@@ -31,6 +31,7 @@
         "log_value",
         "pot_cage",
         "pot_file_dsr",
+        "pot_file_vow",
```
m0ar: What's this got to do with this PR? :thinking:

elizabethengelman: I think this PR just needs a rebase - these pot_file_vow changes were merged into staging already

plugins/transformerExporter.go
elizabethengelman: I'm kind of in favor of having each of these transformer initializers on a new line so it's easier to see what has been included, though it does make the file a lot longer. 🤷‍♀ But, don't feel super strongly either way.

Dockerfile
```diff
@@ -1,24 +1,20 @@
 FROM golang:alpine as builder
 
 RUN apk --update --no-cache add make git g++ linux-headers
-# DEBUG
-RUN apk add busybox-extras
```
elizabethengelman: just out of curiosity - do you know what we need this package for in headerSync?

Dockerfile
```diff
@@ -1,24 +1,20 @@
...
 ENV GO111MODULE on
 
-# Build statically linked vDB binary (wonky path because of Dep)
 WORKDIR /go/src/github.com/makerdao/vdb-mcd-transformers
 ADD . .
 
 WORKDIR /go/src/github.com/makerdao
 RUN git clone https://github.com/makerdao/vulcanizedb.git
 WORKDIR /go/src/github.com/makerdao/vulcanizedb
-RUN git checkout v0.0.10
```
elizabethengelman: I'm trying to think if we'd need any sort of change to our workflow with pointing at VDB staging, in case we make a change to VDB that's incompatible with mcd transformers. e.g. the situation we'd be in if https://github.com/makerdao/vulcanizedb/pull/8 gets merged in before https://github.com/makerdao/vdb-mcd-transformers/pull/52

Dockerfile
```diff
@@ -46,36 +41,17 @@ RUN apk add --update --no-cache git g++ linux-headers
 
 ARG USER
 ARG config_file=environments/docker.toml
-ARG vdb_connect
-
-# setup environment
-ENV GOPATH $HOME/go
-ENV GO111MODULE on
```
elizabethengelman: how come we're able to remove these ENV variables? Don't we still need to make sure that the GOPATH is set properly, and go mods are on?

Dockerfile
```diff
@@ -46,36 +41,17 @@ RUN apk add --update --no-cache git g++ linux-headers
...
-RUN adduser -Su 5000 $USER
-# container needs to be writable for plugin execution
-RUN chown 5000:5000 /go/src/github.com/makerdao/vulcanizedb
-
-USER $USER
-
-# chown first so dir is writable
-# note: using $USER is merged, but not in the stable release yet
-COPY --chown=5000:5000 --from=builder /go/src/github.com/makerdao/vulcanizedb .
-COPY --chown=5000:5000 --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers /go/src/github.com/makerdao/vdb-mcd-transformers
```
elizabethengelman: also wondering how we're able to remove these lines, specifically why we're able to remove the `--chown`

plugins/transformerExporter.go
rmulhol: Yeah happy to add line breaks- do y'all know a way to automate that? I just copied over the generated file

Dockerfile
rmulhol: Which package are you referring to? I removed `busybox-extras` bc it didn't seem necessary

Dockerfile
rmulhol: yeah good point. I think this should probably be pointing at a release once a new one is cut. Maybe it would make sense to add an arg that you pass at build time to specify the branch/release of vdb you want to target?

Dockerfile
rmulhol: Don't need them since we built the plugin in the builder container, so this container doesn't need to peruse the GOPATH or load modules

Dockerfile
rmulhol: With the plugin already built, we don't need write access to create an .so file or assemble migrations

Dockerfile
elizabethengelman: 👍  I like that idea.

Dockerfile
elizabethengelman: Yeah, `busybox-extras` - I just wasn't sure what it was even. But I'll up for ✂️ it if it's not needed.

elizabethengelman:
Wondering if this may require an update to the Docker README too?

elizabethengelman:
Wondering if this may require an update to the Docker README too?

elizabethengelman:
https://github.com/makerdao/vdb-mcd-transformers/pull/45

Dockerfile
rmulhol: I think it's [helpful tools](https://hub.docker.com/_/busybox) for tinkering inside the container. May be worth re-adding if we're `sh`ing in there a lot to mess around, but I'm hoping we can set-it-and-forget-it with minimal interference

rmulhol:
> Wondering if this may require an update to the Docker README too?

Done 🎉

plugins/transformerExporter.go
m0ar: Oh right it's the generated exporter, no idea. No biggie either.

PR #47, opened 2019-12-09
Add docker compose file with automatic restart

m0ar:
Maybe this repo is the wrong place for it to live :woman_shrugging: 

Anyway:
* `makerdao/vdb-headersync` & `makerdao/vdb-execute` @ docker hub are now at our disposal

* Since we've separate the containers we can start them as a cluster and use the restart policies from `docker-compose`

* They read from their own respective `.env` file (they could share one if we want to)

* We can pass additional/overwrite environment variables by passing them to the `docker-compose up -d`

* We could keep one `mcd.env` file in the repo, and add an additional `secrets.env` for sensitive things, living directly on the host (DB creds, RPC endpoint etc)

Opinions? : )


m0ar:
OK so this is fixed as far as I can go now. @elizabethengelman is up to date on the remaining blockers.

The `docker-compose.yml` depends on `wait-for-it.sh` to be included in the docker hub images that are referenced.

Dockerfile
```diff
@@ -52,6 +54,7 @@ COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/dockerfiles
 COPY --from=builder /go/src/github.com/pressly/goose/cmd/goose/goose goose
 COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/db/migrations/* db/migrations/
 COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/plugins/transformerExporter.so plugins/transformerExporter.so
+COPY --from=builder /go/src/github.com/makerdao/vulcanizedb/dockerfiles/wait-for-it.sh .
```
rmulhol: probably want to copy this from `vdb-mcd-transformers`?

m0ar: oops yes, this probably would not build :hear_no_evil: Mind fixing?

rmulhol: Fair enough - the reason this stood out to me is because I initially had trouble building this image against an older version of vdb, but I suppose that shouldn't be an issue moving forward.

Still feels a little odd adding the file to this repo and then not using it - any chance we'd want to modify it here without doing the same thing in vdb?

Thinking I may go ahead and rebase + either remove or reference the file, if that works for you?

m0ar: Yeah, feel free! Maybe let's use the file in this repo indeed then :)

dockerfiles/mainnet-1.0.1.env
```diff
@@ -0,0 +1,57 @@
...
+
+CONTRACT_MCD_FLOP_DEPLOYED=8928163
```
rmulhol: I think this might not be the 1.0.1 flop deployment block

yaoandrew: 👍  this was updated in #96

Dockerfile
```diff
@@ -54,7 +54,7 @@ COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/dockerfiles
 COPY --from=builder /go/src/github.com/pressly/goose/cmd/goose/goose goose
 COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/db/migrations/* db/migrations/
 COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/plugins/transformerExporter.so plugins/transformerExporter.so
-COPY --from=builder /go/src/github.com/makerdao/vulcanizedb/dockerfiles/wait-for-it.sh .
+COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/dockerfiles/wait-for-it.sh .
```
yaoandrew: Just curious, does this just determine what repo this file lives in?

rmulhol: It determines what repo we pull the file from. We could technically continue to grab it from vulcanizedb, but that raises two issues:
- if the file doesn't exist in vdb (either an old version or because we decide to remove it), the build fails
- if we want to change that file, we need to do it in that repo

m0ar:
:clap: :100:

PR #48, opened 2019-12-10
Vdb 995 vat grab

yaoandrew:


transformers/test_data/vat_grab.go
```diff
@@ -97,22 +93,18 @@ var VatGrabHeaderSyncLogWithNegativeDink = core.HeaderSyncLog{
 	Transformed: false,
 }
 
-var VatGrabModelWithNegativeDink = shared.InsertionModel{
+var VatGrabModelWithNegativeDink = event.InsertionModel{
```
rmulhol: I think we should be using `CopyEventModel` here and not exporting the `event.InsertionModel` directly (so that modifications to properties on the var don't pollute other tests)

transformers/events/vat_grab/converter_test.go
```diff
@@ -29,10 +31,11 @@ import (
 )
 
 var _ = Describe("Vat grab converter", func() {
-	var converter vat_grab.VatGrabConverter
+	var converter vat_grab.Converter
+	db := test_config.NewTestDB(test_config.NewTestNode())
```
rmulhol: Liked Gabe's comment on another PR - the `BeforeEach` should clean the test db. Also, I think it'd be good to explicitly declare the `db` a `var`. Could be wrong, but I feel like I've run into scoping issues when something like this is declared implicitly

PR #49, opened 2019-12-10
Move vat_fold & index it by ilk

m0ar:


db/migrations/20181002220302_create_vat_fold.sql
```diff
@@ -4,20 +4,21 @@ CREATE TABLE maker.vat_fold
     id        SERIAL PRIMARY KEY,
     header_id INTEGER NOT NULL REFERENCES headers (id) ON DELETE CASCADE,
     log_id    BIGINT  NOT NULL REFERENCES header_sync_logs (id) ON DELETE CASCADE,
-    urn_id    INTEGER NOT NULL REFERENCES maker.urns (id) ON DELETE CASCADE,
+    ilk_id    INTEGER NOT NULL REFERENCES maker.ilks (id) ON DELETE CASCADE,
+    u         TEXT NOT NULL,
```
rmulhol: Should `u` potentially be an int referencing `public.addresses (id)`? Not sure if it matters if this is always just the address for the `vow`, but thinking that would match our current approach elsewhere

m0ar: Didn't see that for `vat_suck` which is very similar, that's why I didn't :)

rmulhol: 🤔 seems like maybe we should make the change in both places?

m0ar: Hmm maybe that's true.. Would you mind creating tasks for this? I'm quite sure there are more ad-hoc addresses that we don't reference there that we should also catch in that cleanup.

PR #50, opened 2019-12-10
VDB-1058 Speedup tests

m0ar:
:racing_car:

### DIFF
--- a/transformers/component_tests/queries/queries_suite_test.go
+++ b/transformers/component_tests/queries/queries_suite_test.go
@@ -1,15 +1,14 @@
 package queries
 
 import (
-	"github.com/makerdao/vdb-mcd-transformers/test_config"
-	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
 	"io/ioutil"
 	"testing"
 
-	"github.com/sirupsen/logrus"
-
+	"github.com/makerdao/vdb-mcd-transformers/test_config"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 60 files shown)


transformers/component_tests/queries/queries_suite_test.go
```diff
@@ -1,6 +1,8 @@
 package queries
 
 import (
+	"github.com/makerdao/vdb-mcd-transformers/test_config"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
 	"io/ioutil"
 	"testing"
```
rmulhol: Can you remove spaces between imports in this file and then run `goimports -w transformers/component_tests/queries/queries_suite_test.go`

m0ar:
@rmulhol I think the issue has been that we have initialised a completely new connection _per test_, which we hadn't explicitly closed until you added your `AfterEach` things. Now, there is only one connection _per suite_ and it should be fine :)

rmulhol:
👍 definitely no problems here - mostly just noting since the pattern of:
```
var name type
BeforeEach(func() {
     name = type{}
})
```
is usually pretty bulletproof, so all things being equal I wouldn't abandon it without a good reason (which we definitely have here).

I think I've seen issues where:
```
var name = func()
```
can sometimes result in zero-valued vars (though I could be misremembering - thought I had seen that when the function depended on other vars/functions that may not have been evaluated in time to make a successful assignment 🤔).

Also could be that I'm confusing things in a `var` declaration with implicitly defined vars, e.g.
```
Describe("thing", func() {
     name := func()
     It("does something", func() {
          ...
     })
})
```
(with no var declaration)

PR #51, opened 2019-12-10
(VDB-1050) add missing indexes

Gslaughl:
Add the indexes that postgraphile complains about lacking when run with the `--no-ignore-indexes` flag

Also I noticed that when running postgraphile with the `api` and `maker` schemas, a bunch of convenience methods being used by our trigger tables are exposed, so I went ahead and added `omit` tags to those. It might make more sense to exclude them a different way (maybe a different schema?), but I figured this might work as a first pass.

### DIFF
--- a/db/migrations/20190709160237_create_flip_storage_tables.sql
+++ b/db/migrations/20190709160237_create_flip_storage_tables.sql
@@ -11,9 +11,11 @@ CREATE TABLE maker.flip_bid_bid
     UNIQUE (diff_id, header_id, bid_id, address_id, bid)
 );
 
+CREATE INDEX flip_bid_bid_header_id_index
+    ON maker.flip_bid_bid (header_id);
 CREATE INDEX flip_bid_bid_bid_id_index
     ON maker.flip_bid_bid (bid_id);
-CREATE INDEX flip_bid_bid_address_id_index
+CREATE INDEX flip_bid_bid_address_index
     ON maker.flip_bid_bid (address_id);
 
 CREATE TABLE maker.flip_bid_lot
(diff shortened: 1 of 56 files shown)


db/migrations/20190709160237_create_flip_storage_tables.sql
```diff
@@ -129,6 +129,9 @@ CREATE TABLE maker.flip_vat
     UNIQUE (header_id, address_id, vat)
 );
 
+CREATE INDEX flip_vat_address_id_index
```
rmulhol: seems like sometimes above we're calling this same relationship an `address_index` - would be down to choose either that or `address_id_index` and use it everywhere

db/migrations/20190709160237_create_flip_storage_tables.sql
```diff
@@ -189,7 +203,12 @@ COMMENT ON TABLE maker.flip_kicks IS E'@name flipKicksStorage';
 -- SQL in this section is executed when the migration is rolled back.
 DROP INDEX maker.flip_kicks_address_id_index;
 DROP INDEX maker.flip_kicks_kicks_index;
```
rmulhol: Not introduced by this PR but I wonder why we need this index? I guess it's probably used in a join 🤔

db/migrations/20190711091453_create_flip_table.sql
```diff
@@ -29,6 +32,9 @@ LIMIT 1
 $$
     LANGUAGE sql;
 
+COMMENT ON FUNCTION get_latest_flip_bid_guy
+    IS E'@omit';
```
rmulhol: 🎉

db/migrations/20190709160237_create_flip_storage_tables.sql
Gslaughl: Nice catch! Currently it's not getting referenced by any other table, but my guess is we were using it at some point to get a `created` value for bids. Now we're getting `created` from `flip_kick` events instead of this storage table, so it seems safe to delete the index.

PR #53, opened 2019-12-10
Remove repo from vat_flux

yaoandrew:


transformers/events/vat_flux/converter_test.go
```diff
@@ -19,23 +19,41 @@ package vat_flux_test
...
-	It("Converts logs to models", func() {
-		models, err := converter.ToModels(constants.VatABI(), []core.HeaderSyncLog{test_data.VatFluxHeaderSyncLog})
+	var (
+		converter vat_flux.Converter
+		db        *postgres.DB
+	)
+
+	BeforeEach(func() {
+		converter = vat_flux.Converter{}
+		db = test_config.NewTestDB(test_config.NewTestNode())
```
rmulhol: Maybe initialize the db in the `var` declaration so it only happens once per suite

transformers/events/vat_flux/converter_test.go
```diff
@@ -19,23 +19,41 @@ package vat_flux_test
 import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/makerdao/vdb-mcd-transformers/test_config"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
 	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
```
rmulhol: if you remove this space and run `goimports -w transformers/events/vat_flux/converter_test.go` it'll rearrange the deps better

transformers/integration_tests/vat_flux.go
```diff
@@ -17,6 +17,7 @@
 package integration_tests
 
 import (
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
 	"strconv"
```
rmulhol: ditto here

transformers/test_data/vat_flux.go
```diff
@@ -53,20 +53,17 @@ var VatFluxHeaderSyncLog = core.HeaderSyncLog{
 	Transformed: false,
 }
 
-var VatFluxModel = shared.InsertionModel{
+var VatFluxModel = event.InsertionModel{
```
rmulhol: I think we want to make the actual `event.InsertionModel` private and export a `VatFluxModel()` function that calls `CopyEventModel` (so that we don't modify properties on the var directly.

yaoandrew: Ok.. I think I'll need to do this in a few other places as well.

PR #55, opened 2019-12-10
Remove repo from vat_fork

yaoandrew:


db/schema.sql
```diff
@@ -5698,6 +5728,7 @@ WITH passed_block_number AS (
     FROM public.headers
     WHERE id = header_id
 )
+
```
rmulhol: Would be cool to remove this from the diff if it's just whitespace. I think there's some difference in the way the schema gets generated bw mac vs linux :/

transformers/events/vat_fork/converter_test.go
```diff
@@ -19,31 +19,54 @@ package vat_fork_test
...
-	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
 	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
 	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
 )
 
 var _ = Describe("VatFork converter", func() {
-	converter := vat_fork.VatForkConverter{}
+	var (
+		converter vat_fork.Converter
+		db        *postgres.DB
```
rmulhol: can init the db here to have it only happen once for the suite

transformers/test_data/vat_fork.go
```diff
@@ -53,22 +53,19 @@ var VatForkHeaderSyncLogWithNegativeDinkDart = core.HeaderSyncLog{
 	Transformed: false,
 }
 
-var VatForkModelWithNegativeDinkDart = shared.InsertionModel{
+var VatForkModelWithNegativeDinkDart = event.InsertionModel{
```
rmulhol: should probably use `CopyEventModel` here

transformers/test_data/vat_fork.go
```diff
@@ -17,6 +17,7 @@
 package test_data
 
 import (
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
 	"math/rand"
```
rmulhol: can you remove whitespace in the imports and then `goimports -w` these files

PR #57, opened 2019-12-11
(VDB-986) remove spot event repos

Gslaughl:


transformers/events/spot_file/pip/converter.go
```diff
@@ -20,46 +20,51 @@ import (
...
-	numTopicsRequired = 4
-)
-
-func (SpotFilePipConverter) ToModels(_ string, logs []core.HeaderSyncLog) ([]shared.InsertionModel, error) {
-	var models []shared.InsertionModel
+func (Converter) ToModels(_ string, logs []core.HeaderSyncLog, db *postgres.DB) ([]event.InsertionModel, error) {
+	var models []event.InsertionModel
 	for _, log := range logs {
-		err := shared.VerifyLog(log.Log, numTopicsRequired, logDataRequired)
+		err := shared.VerifyLog(log.Log, shared.FourTopicsRequired, shared.LogDataRequired)
```
Gslaughl: Previously `logDataRequired` was false, but I'm pretty sure it should be true since we're getting `pip` from the log data.

rmulhol: Nice catch!

PR #62, opened 2019-12-12
(VDB-1080) Add Spot file par transformer

rmulhol:


db/migrations/20190318000000_create_spot_file.sql
```diff
@@ -16,6 +16,20 @@ CREATE INDEX spot_file_mat_header_index
 CREATE INDEX spot_file_mat_ilk_index
     ON maker.spot_file_mat (ilk_id);
 
+CREATE TABLE maker.spot_file_par
+(
+    id        SERIAL PRIMARY KEY,
+    header_id INTEGER NOT NULL REFERENCES headers (id) ON DELETE CASCADE,
+    log_id    BIGINT  NOT NULL REFERENCES header_sync_logs (id) ON DELETE CASCADE,
```
Gslaughl: Postgraphile basically asks for an index on every FK. So if we want to indulge it, we'll need one on `log_id`.

go.sum
```diff
@@ -410,6 +410,7 @@ google.golang.org/grpc v1.21.0/go.mod h1:oYelfM1adQP15Ek0mdvEgi9Df8B9CZIaU1084ij
 google.golang.org/grpc v1.23.1/go.mod h1:Y5yQAOtifL1yxbo5wqy6BxZv8vAUGQwXBOALyacEbxg=
 gopkg.in/alecthomas/kingpin.v2 v2.2.6/go.mod h1:FMv+mEhP44yOT+4EoQTLFTRgOQ1FBLkstjWtayDeSgw=
 gopkg.in/bsm/ratelimit.v1 v1.0.0-20160220154919-db14e161995a/go.mod h1:KF9sEfUPAXdG8Oev9e99iLGnl2uJMjc5B+4y3O7x610=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
```
Gslaughl: I'm kind of weirded out by how erratic our dependencies are all of a sudden 🤔

transformers/events/spot_file/par/initializer/initializer.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/makerdao/vdb-mcd-transformers/transformers/events/spot_file/par"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/libraries/shared/transformer"
+)
+
+var EventTransformerInitializer transformer.EventTransformerInitializer = event.Transformer{
+	Config:    shared.GetEventTransformerConfig(constants.SpotFileParTable, constants.SpotFileParSignature()),
+	Converter: &par.Converter{},
```
Gslaughl: Are we moving away from using pointers here? Tbh I don't know why we'd use one or the other in this instance, just thinking about consistency

rmulhol: Yeah good catch - I think we should trend toward only using them when we have a good reason, like needing to modify the internal state of the struct. Will remove here since we pass the db connection as an arg to the function (rather than calling `SetDB` on the converter)

PR #63, opened 2019-12-13
Fix jug drip initializer

rmulhol:
Fixes an oversight from my previous PR 😅 

Second commit is not as urgent but felt like a worthwhile simplification - can remove if others aren't on board


rmulhol:
Interesting thought! I'd definitely be in favor of using more of our actual source code in the integration tests 👍 

I think the main reason we aren't doing this now is to make it easier to configure the block number that the transformer executes over. The initializer derives the starting block number from the config, and returns a struct implementing the `EventTransformer` interface (where the config is no longer writable). Going to put a story on the board to consider whether we should enable setting the configured block number so that the integration tests can use the initializers

PR #64, opened 2019-12-13
(VDB-1038) Add Pot drip transformer

rmulhol:


go.sum
```diff
@@ -410,6 +410,7 @@ google.golang.org/grpc v1.21.0/go.mod h1:oYelfM1adQP15Ek0mdvEgi9Df8B9CZIaU1084ij
 google.golang.org/grpc v1.23.1/go.mod h1:Y5yQAOtifL1yxbo5wqy6BxZv8vAUGQwXBOALyacEbxg=
 gopkg.in/alecthomas/kingpin.v2 v2.2.6/go.mod h1:FMv+mEhP44yOT+4EoQTLFTRgOQ1FBLkstjWtayDeSgw=
 gopkg.in/bsm/ratelimit.v1 v1.0.0-20160220154919-db14e161995a/go.mod h1:KF9sEfUPAXdG8Oev9e99iLGnl2uJMjc5B+4y3O7x610=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
```
Gslaughl: Do we want to check this in?

db/migrations/20191212211420_create_pot_drip.sql
```diff
@@ -0,0 +1,18 @@
+-- +goose Up
+CREATE TABLE maker.pot_drip
+(
+    id         SERIAL PRIMARY KEY,
+    header_id  INTEGER NOT NULL REFERENCES headers (id) ON DELETE CASCADE,
+    log_id     BIGINT  NOT NULL REFERENCES header_sync_logs (id) ON DELETE CASCADE,
+    msg_sender INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
```
Gslaughl: Postgraphile will probably ask for indexes on `log_id` and `msg_sender`

transformers/test_data/pot_drip.go
```diff
@@ -0,0 +1,48 @@
+package test_data
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"math/rand"
```
Gslaughl: nbd but `goimports` could help out here

transformers/events/pot_drip/converter.go
```diff
@@ -0,0 +1,43 @@
...
+	for _, log := range logs {
+		err := shared.VerifyLog(log.Log, shared.TwoTopicsRequired, shared.LogDataNotRequired)
+		if err != nil {
+			return nil, err
+		}
+
+		msgSender := common.BytesToAddress(log.Log.Topics[1].Bytes()).Hex()
+		addressId, addressErr := shared.GetOrCreateAddress(msgSender, db)
+		if addressErr != nil {
+			_ = shared.ErrCouldNotCreateFK(addressErr)
```
Gslaughl: We probably want to return `addressErr` here

transformers/test_data/pot_drip.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+var rawPotDripLog = types.Log{
+	Address: common.HexToAddress(PotAddress()),
+	Topics: []common.Hash{
+		common.HexToHash(constants.PotDripSignature()),
+		common.HexToHash("0x00000000000000000000000087e76b0a50efc20259cafe0530f75ae0e816aaf2"),
```
Gslaughl: I like this approach of not including more topics than necessary in our test data. Now that I think about it, would it make sense to omit the `Data` field on line 19 for converters that don't need log data?

transformers/events/pot_drip/converter.go
rmulhol: Nice catch!

go.sum
rmulhol: I think so, but would be happy to remove if others disagree. As I understand it, the purpose of the `go.sum` is to cache the checksums of all deps - so it seems like we should include basically anything that's added other than sensitive/private files.

transformers/test_data/pot_drip.go
rmulhol: tbh I removed them because I saw some error when they were included. But I'm not able to reproduce that and thinking maybe I copy/pasted the empty topics wrong. I'm inclined to re-add them and keep everything to accurately simulate what we'd expect over the RPC

PR #65, opened 2019-12-13
(VDB-1051) Create flop vow storage transformer

Gslaughl:


rmulhol:
LGTM but withholding approval since this shouldn't be merged until 0.2.17 is cut 👍

transformers/storage/bid_storage.go
```diff
@@ -26,6 +26,7 @@ var (
 	Tau    = "tau"
 	Kicks  = "kicks"
 	Live   = "live"
+	Vow    = "vow"
```
elizabethengelman: totally doesn't have to be part of this PR, but I wonder if we should rename this file? 🤔

Gslaughl: Yeah good point. Seems like these vars might be more at home in `constants/keys.go`, with the rest of our key/column strings

PR #68, opened 2019-12-16
(VDB-1039) Add pot join event transformer

Gslaughl:


transformers/test_data/pot_join.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/makerdao/vulcanizedb/pkg/core"
+)
+
+var rawPotJoinLog = types.Log{
+	Address: common.HexToAddress(PotAddress()),
+	Topics: []common.Hash{
+		common.HexToHash("0x049878f300000000000000000000000000000000000000000000000000000000"),
+		common.HexToHash("0x000000000000000000000000e7bc397dbd069fc7d0109c0636d06888bb50668c"),
+		common.HexToHash("0x000000000000000000000000000000000000000000000000417fa3222791bd1a"),
+		common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000000"),
```
Gslaughl: These topics are from a real kovan event (the one used in integration tests), and I just thought it was weird how it has 3 topics even though we're only expecting 2. I'm probably missing some subtlety of solidity, but it seemed worth mentioning

rmulhol: I _think_ this is because `LogNote` always includes 4 topics, regardless of how many topics are actually used by an individual `note` modifier. Hopefully we could (theoretically) omit this field and everything would still behave correctly.

transformers/test_data/pot_join.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+)
+
+var rawPotJoinLog = types.Log{
+	Address: common.HexToAddress(PotAddress()),
+	Topics: []common.Hash{
+		common.HexToHash("0x049878f300000000000000000000000000000000000000000000000000000000"),
```
rmulhol: Totally nbd but I think we conventionally populate topic0 with the signature from the constants. Can't think of any reason that matters other than consistency, though

transformers/events/pot_join/pot_join_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestPotJoin(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "PotJoin Suite")
+}
```
rmulhol: I was going to say we should `logrus.SetOutput(ioutil.Discard)`, but that actually doesn't seem necessary. I wonder if we can delete this from other places if transformers are just converters and converters aren't logging? 🤔

transformers/test_data/pot_join.go
Gslaughl: Yeah everything  does seem to work correctly if we omit the 3rd and 4th topics, but the part that seemed weird to me was that the 3rd topic isn't empty. In other events shaped the same way (e.g. vat heal events), the 3rd and 4th topics are all zeroes, like the 4th one is here.

rmulhol: Oh wait I think `topics[2]` is the `wad` value we want. Pretty sure `topics[1]` is the `msg.sender`: https://github.com/makerdao/dss/blob/master/src/lib.sol#L37

PR #70, opened 2019-12-16
Update vdb version to v0.0.11

elizabethengelman:
updating VDB version to v0.0.11 since that release has been cut

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/howeyc/fsnotify v0.9.0 // indirect
 	github.com/jmoiron/sqlx v0.0.0-20181024163419-82935fac6c1a
 	github.com/magiconair/properties v1.8.0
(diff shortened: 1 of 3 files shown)


rmulhol:
Looks like the plugin tests need to be updated to use the new event transformer initializer (with retry interval + max attempts)

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/ethereum/go-ethereum v1.9.8
 	github.com/jmoiron/sqlx v0.0.0-20181024163419-82935fac6c1a
 	github.com/magiconair/properties v1.8.0
-	github.com/makerdao/vulcanizedb v0.0.11-0.20191212205905-623b53fb346f
+	github.com/makerdao/vulcanizedb v0.0.11
```
rmulhol: 🎉

rmulhol:
Did you copy/paste the migrations from vdb over to here? I don't think there should be a difference but just in case - wouldn't want a mismatch

PR #74, opened 2019-12-16
Use constants for storage table names in repo tests

elizabethengelman:


transformers/storage/cat/repository_test.go
```diff
@@ -229,7 +234,8 @@ var _ = Describe("Cat storage repository", func() {
 				Metadata:      storage.GetValueMetadata(cat.IlkLump, map[storage.Key]string{constants.Ilk: test_helpers.FakeIlk.Hex}, storage.Uint256),
 				PropertyName:  "Lump",
 				PropertyValue: strconv.Itoa(rand.Int()),
-				TableName:     "maker.cat_ilk_lump",
```
elizabethengelman: I was thinking of making a similar change in some of the other places we're hardcoding table names, but I'm not sure if that would reduce the readability of queries... ie. line 204
```			
err = db.Get(&result, `SELECT diff_id, header_id, ilk_id AS key, lump AS value FROM maker.cat_ilk_lump`)
```

could become
```
err = db.Get(&result, `SELECT diff_id, header_id, ilk_id AS key, lump AS value FROM $1`,
     shared.GetFullTableName(constants.MakerSchema, constants.CatIlkLump)
)

```

What do you all think?

yaoandrew: 👍 I don't think it impacts the readability that much. After `FROM` my brain just thinks "..what table is this coming `FROM`?"

transformers/shared/utilities_test.go
```diff
@@ -135,4 +135,13 @@ var _ = Describe("Shared utilities", func() {
 			Expect(actualIlkIdentifier).To(Equal(""))
 		})
 	})
+
+	Describe("GetFullTableName", func() {
+		It("Concatenates a schema and table name", func() {
+			schema := "schema_name"
+			table := "table_name"
+			fullTableName := shared.GetFullTableName(schema, table)
+			Expect(fullTableName).To(Equal("schema_name.table_name"))
```
yaoandrew: ⭐️

transformers/shared/constants/table.go
```diff
@@ -65,3 +66,62 @@ const (
 	VowFlogTable            = "vow_flog"
 	YankTable               = "yank"
 )
+
+// storage tables
+const (
+	CdpManagerVat      = "cdp_manager_vat"
```
rmulhol: Would be a fan of adding `Table` to the end of these names so it's clear what purpose they serve when imported in other packages, though not sure whether that could cause collisions with the event tables?

transformers/test_data/shared_behaviors/storage_repository_behaviors.go
```diff
@@ -3,6 +3,7 @@ package shared_behaviors
 import (
 	"database/sql"
 	"fmt"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
```
rmulhol: 👀

transformers/storage/cat/repository_test.go
rmulhol: I'm a fan of the change - to me the readability concern is outweighed by the scope of change entailed by updating a table name 👍

PR #76, opened 2019-12-17
Move integration tests' db initialization to BeforeSuite

Gslaughl:
~~This is just a proof of concept applied to a couple integration test files-- wanted to get eyes on it before proceeding further.~~

This PR fixes a "too many connections" error when we run our integration tests, and I think it's a solid change for reducing duplication and improving performance as well. The main issue I'm encountering is that GoLand doesn't seem to know `db` and `blockChain` (from `integration_tests_suite_test.go`) are in scope for the integration tests, which is weird because those vars are being declared in the same package.

Interested to hear people's thoughts and suggestions. ~~If this issue is solvable or acceptable, I'll move on with this approach. Otherwise we'll have to find a different way of solving the "too many connections" error.~~ For the time being, I think this approach is the best path forward, although the red squiggles are definitely annoying.


Gslaughl:
Update incoming that fixes red squiggles by renaming the integration test files to end with a `_test`. Not sure why it works but it seems to do the trick

PR #77, opened 2019-12-18
Rename converter transformer

m0ar:
:100:


m0ar:
Hmm, I tried replacing `github.com/makerdao/vulcanizedb => github.com/makerdao/vulcanizedb rename-converter-transformer` but doesn't seem to work :thinking:

m0ar:
Like I said in the call I'm unable to debug this locally since I've got DB troubles, but shouldn't be anything big.

@rmulhol I remember touching some tests where additional parameters were required for maxRetry etc on the new watcher functionality. IDK why this was even missing, maybe I screwed something up? :thinking:

transformers/test_data/deal.go
```diff
@@ -58,15 +57,15 @@ var dealModel = event.InsertionModel{
...
 		event.LogFK,
+		constants.BidIDColumn,
 	},
 	ColumnValues: event.ColumnValues{
-		event.HeaderFK:          DealHeaderSyncLog.HeaderID,
-		constants.BidIDColumn:   "10000000000000000",
-		constants.AddressColumn: rawDealLog.Address.Hex(),
-		event.LogFK:             DealHeaderSyncLog.ID,
+		event.HeaderFK:        DealHeaderSyncLog.HeaderID,
+		event.AddressFK:       rawDealLog.Address.Hex(),
```
elizabethengelman: I think that most of the tests that use these models for assertions, we look up the address id in the test itself rather than relying on this hex string. So maybe we should remove this from the test model since it's not an accurate representation of what a model will look like similar to how we're handling it for [test dent](https://github.com/makerdao/vdb-mcd-transformers/blob/staging/transformers/test_data/dent.go#L80)?

Gslaughl: Nice catch! Yeah that hex definitely doesn't belong there.

PR #79, opened 2019-12-19
(VDB-1026) Pot storage transformer

Gslaughl:
I also had to go back and add `msg.sender` to the `join` and `exit` events, since I forgot to include it when I initially wrote those event transformers. 

Also I don't have any storage diffs handy, so I fudged the component tests, but it should be pretty easy to go back and amend them if we get ahold of any real diffs.


transformers/storage/pot/repository.go
```diff
@@ -0,0 +1,158 @@
...
+	repository.db = db
+}
+
+func (repository PotStorageRepository) insertUserPie(diffID, headerID int64, metadata storage.ValueMetadata, pie string) error {
+	user, err := getUser(metadata.Keys)
+	if err != nil {
+		return err
+	}
+
+	var insertUserPieDiff diffInserter = func(addressID int64, tx *sqlx.Tx) error {
```
rmulhol: 🥇

transformers/storage/pot/repository_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+			createErr := repo.Create(diffID, fakeHeaderID, pot.VatMetadata, fakeAddress)
+			Expect(createErr).NotTo(HaveOccurred())
+
+			var count int
+			query := `SELECT COUNT(*) FROM maker.pot_vat`
+			queryErr := db.Get(&count, query)
+			Expect(queryErr).NotTo(HaveOccurred())
+			Expect(count).To(Equal(1))
+		})
```
rmulhol: Why not uses the shared behaviors here?

transformers/storage/repository.go
```diff
@@ -219,6 +220,19 @@ func (repository *MakerStorageRepository) GetFlipBidIds(contractAddress string)
...
+func (repository *MakerStorageRepository) GetPotPieUsers() ([]string, error) {
+	var userAddresses []string
+	err := repository.db.Select(&userAddresses, `
+		SELECT addresses.address
+		FROM maker.pot_join
+		    LEFT JOIN public.addresses ON pot_join.msg_sender = addresses.id
+		UNION
+		SELECT addresses.address
+		FROM maker.pot_exit
+		    LEFT JOIN public.addresses ON pot_exit.msg_sender = addresses.id`)
```
rmulhol: Seems legit and follows the pattern we've used, but there should never be an address exiting without a prior join - right? Totally fine with leaving things as is to avoid errors from missed events, just curious

transformers/storage/pot/repository_test.go
Gslaughl: The "persists a record" test is more complex than the shared test suite, because it has to deal with the address FK. But the "duplicate record" test is identical to the one in the shared suite, so maybe it would be worth breaking apart the shared test suite to allow us to extract only the "duplicate record" one in cases like these.

transformers/storage/repository.go
Gslaughl: Yeah, totally! I hadn't really thought about that. Hopefully one day we'll have a node so reliable that we won't need to be redundant :)

transformers/storage/pot/repository_test.go
rmulhol: ah, that makes sense - would be interested in refactors that make it easier to use the shared tests for fields including a FK, but definitely doesn't need to be a part of this PR 👍

PR #80, opened 2019-12-19
Vdb 1087 cat rely

yaoandrew:
This is will add the event transformer for the rely method.

### DIFF
--- a/transformers/events/cat_rely/converter_test.go
+++ b/transformers/events/cat_rely/converter_test.go
@@ -1,0 +1,43 @@
+package cat_rely_test
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/makerdao/vdb-mcd-transformers/test_config"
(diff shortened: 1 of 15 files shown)


transformers/events/cat_rely/converter_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Cat Rely converter", func() {
+	var (
+		converter event.Converter
+		db        *postgres.DB
```
Gslaughl: Probably won't really make a difference performance-wise, but do we want to initialize the `db` outside of the `BeforeEach` for consistency's sake?

rmulhol:
Closing in favor of #84

PR #84, opened 2020-01-03
(VDB-1074) rely transformer

Gslaughl:
Builds on top of @yaoandrew's work by generalizing the `Cat.rely` transformer to work across all contracts with `rely` events. Also took it a step further by combining the transformers for `rely` and `deny` events into a single transformer, which I called the `auth` transformer, but we don't have to include that commit depending on how people feel about it.


transformers/events/auth/transformer_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Deny Transformer", func() {
```
rmulhol: :%s/Deny/Auth

PR #87, opened 2020-01-09
Updating README to explain current usage

elizabethengelman:
A couple of things that I think would be helpful to add:
- how to add a new transformer
- how to use/query the graphiql API

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ about transformers.
 These core commands can be run via Docker images which is the preferred method, or they can be built and run via the
 command line interface. In either method, a postgres database will first need to be created:
 1. Install Postgres
-1. Create a superuser for yourself and make sure `psql --list` works without prompting for a password.
+1. Create a user for yourself that is able run migrations and add extensions.
 1. `createdb vulcanize_public`
 1. Migrating the database manually is unnecessary as the commands handle this.
 
--- a/README.md
+++ b/README.md
@@ -2,145 +2,132 @@
 
 [![Build Status](https://travis-ci.com/makerdao/vdb-mcd-transformers.svg?token=yVLfQ4hNQCqvt2qjLkus&branch=staging)](https://travis-ci.com/makerdao/vdb-mcd-transformers)
 
-## About
 
-Vulcanize DB is a set of tools that make it easier for developers to write application-specific indexes and caches for dapps built on Ethereum.
+## Table of Contents
+1. [Background](#background)
+1. [Install](#install)
+1. [Usage](#usage)
+1. [Contributing](#contributing)
+1. [License](#license)
 
-## Dependencies
- - Go 1.11+
- - Postgres 11
+## Background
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,145 +2,132 @@
...
-
-After `dep` finishes, dependencies should be installed within your `GOPATH` at the versions specified in `Gopkg.toml`.
-
-Lastly, ensure that `GOPATH` is defined in your shell. If necessary, `GOPATH` can be set in `~/.bashrc` or `~/.bash_profile`, depending upon your system. It can be additionally helpful to add `$GOPATH/bin` to your shell's `$PATH`.
-
-## Setting up the Database
+   
+### Getting the project
+Download the transformer codebase to your local local `GOPATH` via:
+`go get github.com/makerdao/vdb-mcd-transformers`
```
rmulhol: Does this work? I thought we might not be able to use `go get` as long as we're importing the Geth fork

README.md
```diff
@@ -2,145 +2,132 @@
...
+
+1. `execute` uses the raw Ethereum data that has been synced into Postgres and applies transformations to configured MCD
+contract data via [event](./transformers/events) and [storage](./transformers/storage) transformers. The VulcanizeDB
+repository includes a [general description](https://github.com/makerdao/vulcanizedb/blob/staging/documentation/custom-transformers.md)
+about transformers.
+
+These core commands can be run via Docker images which is the preferred method, or they can be built and run via the
+command line interface. In either method, a postgres database will first need to be created:
 1. Install Postgres
 1. Create a superuser for yourself and make sure `psql --list` works without prompting for a password.
```
rmulhol: Do you know if this is still necessary? Thinking we might be able to avoid it as long as the db user/pw are included in config

README.md
elizabethengelman: Good call, I had forgotten about that. I think `go get` should work for getting the transformers repo - since it doesn't directly depend on geth, but geth through VDB. But you're totally right that we'll probably have to use `git clone` for the vdb repo.

That said, I wonder if we should use the same method for both repos, just for clarity/consistency?

README.md
elizabethengelman: It looks like we still need to create a supersuser for creating the `citext` extension which we're using to allow for case insensitivity for urn identifiers.
```
CREATE EXTENSION IF NOT EXISTS citext;
ERROR:  permission denied to create extension "citext"
HINT:  Must be superuser to create this extension.
```

I wonder if we should invest some time into figuring out a way around this, so we can remove the necessity for superusers.

rmulhol: I might be focused too much on the part that says "make sure `psql --list` works without prompting for a password" - thinking maybe we can avoid that as long as the user/pw has the right permissions to run the migrations

elizabethengelman: gotcha, yeah, that makes sense to me, does the change in f69b277 make sense?

PR #88, opened 2020-01-09
(VDB-110) cat wards storage transformer

Gslaughl:
Transforms `wards` storage mapping for the `Cat` contract. Using `usr` and `msg_sender` of `rely` and `deny` events for key discovery.

### DIFF
--- a/db/migrations/20191219074134_create_rely.sql
+++ b/db/migrations/20191219074134_create_rely.sql
@@ -6,6 +6,7 @@ CREATE TABLE maker.rely
     header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
     log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
     address_id INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    msg_sender INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
     usr        INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
     UNIQUE (header_id, log_id)
 );
@@ -16,6 +17,8 @@ CREATE INDEX rely_log_index
     ON maker.rely (log_id);
 CREATE INDEX rely_address_index
     ON maker.rely (address_id);
+CREATE INDEX rely_msg_sender_index
+    ON maker.rely (msg_sender);
 CREATE INDEX rely_usr_index
     ON maker.rely (usr);
 
@@ -25,6 +28,7 @@ CREATE INDEX rely_usr_index
 DROP INDEX maker.rely_header_index;
 DROP INDEX maker.rely_log_index;
 DROP INDEX maker.rely_address_index;
+DROP INDEX maker.rely_msg_sender_index;
 DROP INDEX maker.rely_usr_index;
 
 DROP TABLE maker.rely;
--- a/db/migrations/20200102160236_create_wards_storage_table.sql
+++ b/db/migrations/20200102160236_create_wards_storage_table.sql
@@ -1,0 +1,26 @@
+-- +goose Up
+CREATE TABLE maker.wards
+(
+    id         SERIAL PRIMARY KEY,
+    diff_id    BIGINT  NOT NULL REFERENCES storage_diff (id) ON DELETE CASCADE,
+    header_id  INTEGER NOT NULL REFERENCES headers (id) ON DELETE CASCADE,
+    address_id INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    usr        INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    wards      INTEGER NOT NULL,
+    UNIQUE (diff_id, header_id, address_id, usr, wards)
+);
+
+CREATE INDEX wards_header_id_index
+    ON maker.wards (header_id);
+CREATE INDEX wards_address_index
+    ON maker.wards (address_id);
+CREATE INDEX wards_usr_index
+    ON maker.wards (usr);
+
+
+-- +goose Down
+DROP INDEX maker.wards_usr_index;
+DROP INDEX maker.wards_address_index;
+DROP INDEX maker.wards_header_id_index;
+
+DROP TABLE maker.wards;
--- a/transformers/component_tests/cat/transformer_test.go
+++ b/transformers/component_tests/cat/transformer_test.go
@@ -111,7 +111,7 @@ var _ = Describe("Executing the transformer", func() {
 			catAddressID, catAddressErr := shared.GetOrCreateAddress(test_data.CatAddress(), db)
 			Expect(catAddressErr).NotTo(HaveOccurred())
 
-			userAddress := "0x13141b8a5e4a82ebc6b636849dd6a515185d6236"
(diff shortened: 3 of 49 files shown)


transformers/storage/repository.go
```diff
@@ -257,8 +258,37 @@ func (repository *MakerStorageRepository) GetFlopBidIds(contractAddress string)
...
+		UNION
+		SELECT addresses.address
+		FROM maker.deny
+		    LEFT JOIN public.addresses ON deny.usr = addresses.id
+		WHERE deny.address_id = $1
+		UNION
+		SELECT addresses.address
+		FROM maker.deny
+		LEFT JOIN public.addresses ON deny.msg_sender = addresses.id
+		WHERE deny.address_id = $1`, contractAddressID)
```
Gslaughl: The only reason this query needs to combine more than just `rely.usr` and `deny.usr` is because we also want to find the `msg_sender` of the transaction that deployed the contract. Since we don't have a direct way of doing that (constructor doesn't emit an event), we have to infer it.

The `msg_sender` of the first `rely` event _must_ be the `msg_sender` who deployed the contract, because at that point no one else is authorized to call `rely` (or any other method with the `auth` modifier). Technically I didn't need to `UNION` with _all_ `rely`/`deny` `msg_sender`s (just the 1st `rely` event's), but I figured we might as well for redundancy's sake.

Also worth noting that these tables will likely be very small, so hopefully there's no reason to worry about performance.

transformers/integration_tests/deny_test.go
```diff
@@ -15,58 +15,48 @@ import (
...
 var _ = Describe("Deny transformer", func() {
-	const (
-		defaultOffset int = 0
-		vatOffset     int = -1
-	)
-
 	Context("Cat deny events", func() {
 		usrAddress := "0x13141b8a5e4a82ebc6b636849dd6a515185d6236"
-		denyIntegrationTest(int64(14764643), test_data.CatAddress(), usrAddress, defaultOffset)
+		denyIntegrationTest(int64(14764643), test_data.CatAddress(), usrAddress, usrAddress)
```
Gslaughl: Kind of weird that for all these integration tests I'm passing in `usrAddress` twice, but it's because I just happened to pick `deny` events where `msg.sender`s are passing their own addresses in as the `usr` parameter. I didn't want to abstract `usrAddressHex` and `msgSenderAddressHex` (on line 59) into a single parameter, because I was worried that would imply that those 2 addresses _have to_ be the same, which they don't.

db/migrations/20200102160236_create_wards_storage_table.sql
```diff
@@ -0,0 +1,26 @@
+-- +goose Up
+CREATE TABLE maker.wards
+(
+    id         SERIAL PRIMARY KEY,
+    diff_id    BIGINT  NOT NULL REFERENCES storage_diff (id) ON DELETE CASCADE,
+    header_id  INTEGER NOT NULL REFERENCES headers (id) ON DELETE CASCADE,
+    address_id INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    usr        INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    wards      INTEGER NOT NULL,
```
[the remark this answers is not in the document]
m0ar: I'd be hesitant to step away from the contract types

db/migrations/20191219074134_create_rely.sql
```diff
@@ -6,6 +6,7 @@ CREATE TABLE maker.rely
     header_id  INTEGER NOT NULL REFERENCES headers (id) ON DELETE CASCADE,
     log_id     BIGINT  NOT NULL REFERENCES header_sync_logs (id) ON DELETE CASCADE,
     address_id INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    msg_sender INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
```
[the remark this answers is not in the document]
yaoandrew: Heads on up these for the rebase... `header_sync_logs` =>  `event_logs` and `addresses` => `public.addresses`

transformers/component_tests/cat/transformer_test.go
```diff
@@ -36,9 +39,9 @@ import (
 var _ = Describe("Executing the transformer", func() {
 	var (
 		db                = test_config.NewTestDB(test_config.NewTestNode())
-		storageKeysLookup = storage.NewKeysLookup(cat.NewKeysLoader(&mcdStorage.MakerStorageRepository{}))
-		repository        = cat.CatStorageRepository{}
-		contractAddress   = "81f7aa9c1570de564eb511b3a1e57dae558c65b5"
+		contractAddress   = test_data.CatAddress()
```
yaoandrew: 💯

transformers/events/auth/transformer_test.go
```diff
@@ -30,41 +30,25 @@ var _ = Describe("Deny Transformer", func() {
 			test_data.RelyHeaderSyncLog.Log.Address.String())
```
yaoandrew: `RelyHeaderSyncLog` => `RelyEventLog`

transformers/storage/test_helpers/maker_storage_repository.go
```diff
@@ -42,18 +43,35 @@ type MockMakerStorageRepository struct {
...
+func (repository *MockMakerStorageRepository) GetFlipBidIds(contractAddress string) ([]string, error) {
+	repository.GetFlipBidIdsCalledWith = contractAddress
+	return repository.FlipBidIds, repository.GetFlipBidIdsError
+}
+
+func (repository *MockMakerStorageRepository) GetFlopBidIds(contractAddress string) ([]string, error) {
+	repository.GetFlopBidIdsCalledWith = contractAddress
+	return repository.FlopBidIds, repository.GetFlopBidIdsError
+}
+
```
yaoandrew: 🥇

transformers/storage/utilities/wards/repository.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/storage"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
+)
+
+var insertWardsQuery = `INSERT INTO maker.wards (diff_id, header_id, address_id, usr, wards) VALUES ($1, $2, $3, $4, $5) ON CONFLICT DO NOTHING`
+
+func InsertWards(diffID, headerID int64, metadata storage.ValueMetadata, contractAddress, wards string, db *postgres.DB) error {
```
yaoandrew: ➕

transformers/integration_tests/deny_test.go
yaoandrew: 👍

transformers/storage/repository.go
```diff
@@ -32,18 +32,19 @@ type Urn struct {
 var ErrNoFlips = errors.New("no flips exist in db")
 
 type IMakerStorageRepository interface {
+	GetCdpis() ([]string, error)
 	GetDaiKeys() ([]string, error)
 	GetFlapBidIds(string) ([]string, error)
+	GetFlipBidIds(contractAddress string) ([]string, error)
```
yaoandrew: 💯

PR #91, opened 2020-01-14
(VDB-1111) flap wards storage transformer

Gslaughl:
Includes wards mapping for flip, flap, flop + jug contracts.

Edit: also Vat


transformers/component_tests/jug/transformer_test.go
```diff
@@ -36,18 +39,18 @@ import (
...
+		storageKeysLookup = storage.NewKeysLookup(jug.NewKeysLoader(&mcdStorage.MakerStorageRepository{}, contractAddress))
+		repository        = jug.JugStorageRepository{ContractAddress: contractAddress}
 		transformer       = storage.Transformer{
 			HashedAddress:     vdbStorage.HexToKeccak256Hash(contractAddress),
 			StorageKeysLookup: storageKeysLookup,
 			Repository:        &repository,
 		}
-		headerID int64
 		header   = fakes.FakeHeader
+		headerID int64
```
rmulhol: Super minor and not introduced by this PR but I've become a fan of not having a `headerID` var and instead always accessing that val via `header.Id` in tests so that we don't get weird failures deriving from `headerID != header.Id`

transformers/component_tests/jug/transformer_test.go
```diff
@@ -36,18 +39,18 @@ import (
...
+		err      error
```
rmulhol: Similar thing for `err` vars - generally prefer to scope these as locally as possible so that we have unique errors in each place they can occur (so that the compiler will tell us if they're unchecked)

transformers/component_tests/jug/transformer_test.go
```diff
@@ -36,18 +39,18 @@ import (
 var _ = Describe("Executing the transformer", func() {
```
rmulhol: 🤔 I wonder if we want to name this `configured_transformer_test`. Still getting used to `converter => transformer`, which I think is a win - but the various vat "transformer tests" threw me off for a minute

transformers/storage/utilities/wards/repository.go
```diff
@@ -49,7 +49,7 @@ func InsertWards(diffID, headerID int64, metadata storage.ValueMetadata, contrac
 	return tx.Commit()
 }
 
-func getUser(keys map[storage.Key]string) (string, error) {
+func GetUser(keys map[storage.Key]string) (string, error) {
```
elizabethengelman: curious why we're exposing this method outside of this package now 🤔

db/migrations/20200108074134_create_vat_rely.sql
```diff
@@ -5,17 +5,14 @@ CREATE TABLE maker.vat_rely
     id         SERIAL PRIMARY KEY,
     header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
     log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
-    address_id INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
```
elizabethengelman: Are we removing the address_id from here (and from vat_deny) because there's only going to be one vat contract per release?

transformers/component_tests/flap/transformer_test.go
```diff
@@ -153,6 +156,45 @@ var _ = Describe("Executing the flap transformer", func() {
...
+
+			denyModel.ColumnValues[event.HeaderFK] = header.Id
+			denyModel.ColumnValues[event.LogFK] = denyLog.ID
+			denyModel.ColumnValues[event.AddressFK] = flapAddressID
+			denyModel.ColumnValues[constants.MsgSenderColumn] = msgSenderAddressID
+			denyModel.ColumnValues[constants.UsrColumn] = userAddressID
+			insertErr := event.PersistModels([]event.InsertionModel{denyModel}, db)
+			Expect(insertErr).NotTo(HaveOccurred())
+
+			key := common.HexToHash("614c9873ec2671d6eb30d7a22b531442a34fc10f8c24a6598ef401fe94d9cb7d")
```
elizabethengelman: I wonder if we could go through the steps to calculate the key here so it's more explicit about where it came from?

db/migrations/20200108074134_create_vat_rely.sql
Gslaughl: Yeah exactly

transformers/storage/utilities/wards/repository.go
Gslaughl: Nice catch! I was using it somewhere else, but ended up undoing the change, so this can totally be a private function.

transformers/component_tests/flap/transformer_test.go
Gslaughl: I like that idea, but in this instance I just grabbed a record from a CSV storage diff. Maybe it would be more thorough to actually calculate the key in the test though 🤔 

The code responsible for the calculation is pretty simple:
```
func GetKeyForMapping(indexOnContract, key string) common.Hash {
	keyBytes := common.FromHex(key + indexOnContract)
	return crypto.Keccak256Hash(keyBytes)
}
```

PR #92, opened 2020-01-16
(VDB-1064) Query historical_ilk_states table instead of using get_ilk

Gslaughl:
This PR replaces the `api.get_ilk` function with queries to the more performant `api.historical_ilk_state` table. We're still using `get_ilk` in some of our tests (specifically the `ilk_state` computed columns tests), but once we refactor the `all_ilks` query to use `historical_ilk_states`, we should be able to delete the `ilk_state` type altogether, along with its computed columns.


db/migrations/20190308111729_create_ilk_history_query.sql
```diff
@@ -1,30 +0,0 @@
--- +goose Up
--- +goose StatementBegin
-
--- Function returning the history of a given ilk as of the given block height
-CREATE FUNCTION api.all_ilk_states(ilk_identifier TEXT, block_height BIGINT DEFAULT api.max_block(),
```
Gslaughl: This function should be totally obsolete now, since its functionality can replicated by filtering the `allHistoricalIlkStates` function that postgraphile exposes

db/migrations/20190329104428_create_frob_event_computed_columns.sql
```diff
@@ -3,10 +3,14 @@
 
 -- Extend frob_event with ilk_state
 CREATE FUNCTION api.frob_event_ilk(event api.frob_event)
-    RETURNS api.ilk_state AS
+    RETURNS api.historical_ilk_state AS
```
rmulhol: Probably missing something but I can't see where in the migrations the `api.historical_ilk_state` is defined 🤔 

I see it in the schema but wanting to double check that the relevant migration (with the name change) was checked in

transformers/component_tests/queries/test_helpers/test_helpers.go
```diff
@@ -235,6 +235,26 @@ func CreateIlk(db *postgres.DB, header core.Header, valuesMap map[string]interfa
 	CreateCatRecords(db, header, valuesMap, catMetadatas, catRepo)
 	CreateJugRecords(db, header, valuesMap, jugMetadatas, jugRepo)
 	CreateSpotRecords(db, header, valuesMap, spotMetadatas, spotRepo)
+	if len(vatMetadatas) >= 1 {
```
rmulhol: What's the rationale behind this predicate? Are there places where we want to create metadata for non-vat contracts but not for the vat?

db/migrations/20190329104428_create_frob_event_computed_columns.sql
Gslaughl: Yeah, `historical_ilk_state` is the name of the trigger table. 

https://github.com/makerdao/vdb-mcd-transformers/blob/vdb-1064-replace-ilk-query/db/migrations/20190301141224_create_ilk_state_history_table_and_triggers.sql#L2

transformers/component_tests/queries/test_helpers/test_helpers.go
Gslaughl: Yeah, we're doing that a few times in the tests for `get_ilk`. E.g. https://github.com/makerdao/vdb-mcd-transformers/blob/vdb-1064-replace-ilk-query/transformers/component_tests/queries/get_ilk_query_test.go#L233

So I guess we could get rid of this conditional when we delete the `get_ilk` function, after the `ilk_state` type is deleted.

rmulhol: Awesome. My initial thought is to wonder whether those tests corrrectly simulating realistic conditions, but I suppose it probably makes sense that we'd want to be able to verify that the records are correct even if some important data is missing.

PR #98, opened 2020-01-27
(VDB-1133) Add auction reorg-safety triggers

Gslaughl:
Here's another PR where we're making very similar changes to every auction's triggers. Plus, each individual trigger is undergoing very similar changes, with the exception of the `created` triggers which are a little simpler, since all the records pertaining to an individual bid have the same `created` value.

### DIFF
--- a/db/migrations/20190726141224_create_flap_table_and_triggers.sql
+++ b/db/migrations/20190726141224_create_flap_table_and_triggers.sql
@@ -11,7 +11,7 @@ CREATE TABLE maker.flap
     bid          NUMERIC   DEFAULT NULL,
     created      TIMESTAMP DEFAULT NULL,
     updated      TIMESTAMP NOT NULL,
-    PRIMARY KEY (block_number, bid_id, address_id)
+    PRIMARY KEY (address_id, bid_id, block_number)
 );
 
 COMMENT ON TABLE maker.flap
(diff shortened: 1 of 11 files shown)


transformers/test_data/shared_behaviors/shared_flip_bid_trigger_suite.go
```diff
@@ -34,35 +34,43 @@ type BidTriggerTestInput struct {
 	PackedValueType types.ValueType
 }
 
-func SharedBidHistoryTriggerTests(input BidTriggerTestInput) {
+func BidHistoryTriggerTests(input BidTriggerTestInput) {
```
rmulhol: Should this file be renamed?

transformers/test_data/shared_behaviors/shared_flip_bid_trigger_suite.go
```diff
@@ -145,17 +157,93 @@ func SharedBidHistoryTriggerTests(input BidTriggerTestInput) {
...
+				_, deleteErr := db.Exec(deleteRowQuery, headerOne.Id)
+				Expect(deleteErr).NotTo(HaveOccurred())
+
+				var valueHistory []sql.NullString
+				queryErr := db.Select(&valueHistory, getFieldQuery)
+				Expect(queryErr).NotTo(HaveOccurred())
+				Expect(valueHistory[0].Valid).To(BeFalse())
+			})
+
+			It("deletes ilk state associated with diff if identical to previous state", func() {
```
rmulhol: :%s/ilk/bid

db/migrations/20190726141224_create_flap_table_and_triggers.sql
```diff
@@ -134,32 +174,31 @@ VALUES (new_diff.bid_id,
         flap_bid_bid_before_block(new_diff.bid_id, new_diff.address_id, new_diff.header_id),
         (SELECT api.epoch_to_datetime(block_timestamp) FROM diff_block),
         flap_bid_time_created(new_diff.address_id, new_diff.bid_id))
-ON CONFLICT (block_number, bid_id, address_id) DO UPDATE SET guy = new_diff.guy;
+ON CONFLICT (block_number, bid_id, address_id) DO UPDATE SET guy = new_diff.guy
 $$
     LANGUAGE sql;
--- +goose StatementEnd
```
rmulhol: Did you intend to remove this? Curious because I don't see a `StatementBegin` removed above, but I see that there is one immediately below

db/migrations/20190726141224_create_flap_table_and_triggers.sql
```diff
@@ -134,32 +174,31 @@ VALUES (new_diff.bid_id,
...
 
 -- +goose StatementBegin
-CREATE OR REPLACE FUNCTION maker.update_flap_guys_until_next_diff(new_diff maker.flap_bid_guy) RETURNS VOID
+CREATE OR REPLACE FUNCTION maker.update_flap_guys_until_next_diff(start_at_diff maker.flap_bid_guy, new_guy TEXT) RETURNS VOID
```
rmulhol: What's the rationale for breaking out the `new_guy` param rather than just reading it off of the diff? Wondering if that could make it trickier to change the type from text -> int referencing addresses (id)

db/migrations/20190726141224_create_flap_table_and_triggers.sql
Gslaughl: Yeah this was a `StatementEnd` without a matching `StatementBegin`; it was actually causing a weird situation where tests were failing in CI but for some reason I couldn't get the same failure locally. Took a minute to troubleshoot what was going on.

As far as I can tell, Goose requires the `StatementBegin`s and `-End`s for `plpgsql` queries, and not for vanilla `sql`. Although it doesn't complain if you include them when they aren't necessary aside from the aforementioned situation where you have a `Begin` w/o an `End` or vice-versa.

db/migrations/20190726141224_create_flap_table_and_triggers.sql
Gslaughl: With this new set of triggers, we had to make these `update_until_next_diff` functions more flexible. The `DELETE` triggers use the same `UPDATE` logic, but updates snapshots to the value of the next-most-recent diff, rather than using the value already contained within the diff being deleted. See line 222.

As a side-note, I've been deliberately avoiding including FKs like `address_id` in the trigger-tables where possible, in order to minimize the number of `JOIN`s necessary at request-time, since the trigger-tables were originally intended as performance-optimizations. Definitely open to reconsidering that approach though, especially where performance isn't likely to be as big of an issue.

rmulhol: 👍 to avoiding foreign keys in the trigger tables - would def prefer any necessary joins be handled in the trigger. Enabling re-use of the update logic in the `DELETE` triggers makes sense to me 👌

db/migrations/20190726141224_create_flap_table_and_triggers.sql
rmulhol: Ah that makes sense - nice catch! Maybe worth creating a story to see if we can isolate the consequences of them missing/verify they're there?

Gslaughl: I might be misunderstanding you, but in my experience the migrations totally fail if you don't have those statements where they're necessary. So it _shouldn't_ be necessary to manually go through and verify that they're there, although I can see a case for making sure they're _not_ there where they're unnecessary, for the sake of reducing clutter.

rmulhol: Ok cool - as long as things break when broken 👍

PR #99, opened 2020-01-27
Include Oasis Contract Watcher in docker-compose.yml

rmulhol:
Don't need to merge this, but opening it up for review since folks may check out this branch to run the API + Oasis watching - would appreciate any thoughts.

One quirk about the contract watcher that's relevant - new tables don't get created until we actually get a matching event from the ABI. This is actually ok in this setup, since it enables us to run all the migrations before the code starts modifying the DB itself, but could potentially lead to issues restarting or applying new migrations later - may want to think about routing all DB updates (including from the contract watcher) through migrations.


rmulhol:
Closing because we don't intend to include this in the standard `docker-compose.yml`

PR #100, opened 2020-01-29
(VDB-1066) Refactor all_ilks query to use historical_ilk_state

Gslaughl:
After updating `all_ilks` to return `historical_ilk_state` instead of `ilk_state`, `ilk_state` was no longer being used in any queries. Which means we were able to delete a bunch of code :D 

I also took the liberty of renaming `historical_ilk_state` to `ilk_snapshot`, which I think makes it easier to immediately understand what's in that table. Interested to hear what people think; the renaming is in its own commit, so we can easily forgo that change if folks don't like it.


db/migrations/20190411132245_create_all_ilks_query.sql
```diff
@@ -1,127 +1,17 @@
 -- +goose Up
 
 -- Function returning state for all ilks as of the given block height
-CREATE FUNCTION api.all_ilks(block_height BIGINT DEFAULT api.max_block(), max_results INTEGER DEFAULT NULL,
-                             result_offset INTEGER DEFAULT 0)
-    RETURNS SETOF api.ilk_state
+CREATE FUNCTION api.all_ilks(block_height BIGINT DEFAULT api.max_block())
```
Gslaughl: I removed the pagination stuff from this query. It can only return as many results as there are collateral-types (currently 2), so pagination felt like overkill haha

db/migrations/20190301141224_create_ilk_state_history_table_and_triggers.sql
```diff
@@ -1,5 +1,5 @@
 -- +goose Up
-CREATE TABLE api.historical_ilk_state
+CREATE TABLE api.ilk_snapshot
```
rmulhol: I dig the renaming, and am wondering whether we might want to go ahead and apply it to all of the other `historical_state` tables for consistency

db/migrations/20190411132245_create_all_ilks_query.sql
rmulhol: Dig it. Has me wondering whether there might be a better name for this query 🤔

db/migrations/20190411132245_create_all_ilks_query.sql
```diff
@@ -1,127 +1,17 @@
...
-              mats.mat is not null
-          )
-ORDER BY updated DESC
-LIMIT all_ilks.max_results
-OFFSET
-all_ilks.result_offset
+SELECT DISTINCT ON (ilk_identifier) *
+FROM api.ilk_snapshot
+WHERE block_number <= block_height
+ORDER BY ilk_identifier, block_number DESC
```
rmulhol: 😍 ✂️

db/migrations/20190901110911_create_total_ink_query.sql
```diff
@@ -1,7 +1,6 @@
 -- +goose Up
 -- SQL in this section is executed when the migration is applied.
 
--- Extend managed_cdp with ilk_state
```
rmulhol: nice catch!

transformers/component_tests/queries/historical_ilk_state_computed_columns_test.go
```diff
@@ -32,7 +32,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("current ilk state computed columns", func() {
+var _ = Describe("ilk_snapshot computed columns", func() {
```
rmulhol: maybe worth also renaming the file?

transformers/storage/vat/repository_test.go
```diff
@@ -544,7 +544,7 @@ var _ = Describe("Vat storage repository", func() {
 			Expect(err).To(MatchError(types.ErrMetadataMalformed{MissingData: constants.Guy}))
 		})
 
-		Describe("updating historical_ilk_state trigger table", func() {
+		Describe("updating historical_urn_state trigger table", func() {
```
rmulhol: 👍

transformers/component_tests/queries/historical_ilk_state_computed_columns_test.go
Gslaughl: 🤦‍♂ good call! I had a feeling I was forgetting something

PR #101, opened 2020-01-31
Vdb 1067 fk addresses

yaoandrew:
Going forward I'll do one of these per PR. This PR contains:
spot_file_pip
newCDP
vat_suck

### DIFF
--- a/transformers/events/new_cdp/transformer.go
+++ b/transformers/events/new_cdp/transformer.go
@@ -55,13 +55,23 @@ func (Transformer) toEntities(contractAbi string, logs []core.EventLog) ([]NewCd
 	return entities, nil
 }
 
-func (t Transformer) ToModels(abi string, logs []core.EventLog, _ *postgres.DB) ([]event.InsertionModel, error) {
+func (t Transformer) ToModels(abi string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
 	entities, entityErr := t.toEntities(abi, logs)
 	if entityErr != nil {
 		return nil, fmt.Errorf("NewCDP transformer couldn't convert logs to entities: %v", entityErr)
 	}
 	var models []event.InsertionModel
 	for _, newCdpEntity := range entities {
+
+		usrAddressID, usrAddressErr := shared.GetOrCreateAddress(newCdpEntity.Usr.Hex(), db)
+		if usrAddressErr != nil {
+			return nil, usrAddressErr
+		}
+		ownAddressID, ownAddressErr := shared.GetOrCreateAddress(newCdpEntity.Own.Hex(), db)
+		if ownAddressErr != nil {
+			return nil, ownAddressErr
+		}
+
 		model := event.InsertionModel{
 			SchemaName: constants.MakerSchema,
 			TableName:  constants.NewCdpTable,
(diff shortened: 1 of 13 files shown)


transformers/events/new_cdp/transformer_test.go
```diff
@@ -26,13 +28,25 @@ import (
 )
 
 var _ = Describe("NewCdp Transformer", func() {
-	var transformer = new_cdp.Transformer{}
+	var (
+		transformer = new_cdp.Transformer{}
+		db          = test_config.NewTestDB(test_config.NewTestNode())
```
rmulhol: I think we probably want to clean the test db in a `BeforeEach` for these tests where we're now initializing a `db`

transformers/integration_tests/spot_file_test.go
```diff
@@ -151,13 +151,17 @@ var _ = Describe("SpotFile EventTransformers", func() {
 
 		It("fetches and transforms a Spot.file pip event", func() {
 			var dbResult spotFilePipModel
-			getSpotErr := db.Get(&dbResult, `SELECT ilk_id, pip from maker.spot_file_pip`)
+			getSpotErr := db.Get(&dbResult, `SELECT ilk_id, what, pip from maker.spot_file_pip`)
```
rmulhol: 👍

transformers/test_data/vat_suck.go
```diff
@@ -54,7 +54,9 @@ var VatSuckEventLog = core.EventLog{
 	Transformed: false,
 }
 
-var VatSuckModel = event.InsertionModel{
+func VatSuckModel() event.InsertionModel { return CopyModel(vatSuckModel) }
```
rmulhol: 🎉

rmulhol (approved):
LGTM!

Probably worth adding 
```
	BeforeEach(func() {
		test_config.CleanTestDB(db)
	})
```
in the test files where we're now initializing a db, but otherwise no blockers! 👍

transformers/events/yank/transformer_test.go
```diff
@@ -35,6 +35,10 @@ var _ = Describe("Yank Transformer", func() {
 		db          = test_config.NewTestDB(test_config.NewTestNode())
 	)
 
+	BeforeEach(func() {
+		test_config.CleanTestDB(db)
```
Gslaughl: 🎉

transformers/events/new_cdp/transformer.go
```diff
@@ -71,8 +81,8 @@ func (t Transformer) ToModels(abi string, logs []core.EventLog, _ *postgres.DB)
 			ColumnValues: event.ColumnValues{
 				event.HeaderFK:      newCdpEntity.HeaderID,
 				event.LogFK:         newCdpEntity.LogID,
-				constants.UsrColumn: newCdpEntity.Usr.Hex(),
-				constants.OwnColumn: newCdpEntity.Own.Hex(),
+				constants.UsrColumn: usrAddressID,
+				constants.OwnColumn: ownAddressID,
```
elizabethengelman: Wondering if we should change the `usr` and `own` fields in the database to be numerics and foreign keys to the address table.

elizabethengelman (commented):
Looking good. The only question I have is if we'd want to change the fields that reference address_ids to be numerics and foreign keys in the database.

gslaughl (commented):
Looking good! My only question is whether there should be some kind of corresponding changes in the migrations, where we tell Postgres that these fields are FKs, and probably change the `TEXT` type to an `INTEGER`.

transformers/events/new_cdp/transformer.go
rmulhol: ➕

gslaughl (approved):

elizabethengelman (approved):

PR #102, opened 2020-01-31
Disable storage transformers

rmulhol:
- enable running against non-patched client


rmulhol:
Closing because we probably don't actually want to merge this

PR #105, opened 2020-02-02
Add urn_bid_events query

Gslaughl:


transformers/component_tests/queries/urn_bid_events_test.go
```diff
@@ -0,0 +1,85 @@
...
+	BeforeEach(func() {
+		test_config.CleanTestDB(db)
+		timestamp := int64(rand.Int31())
+		blockNumber := rand.Int()
+		headerOneID := storageHelpers.CreateHeader(timestamp, blockNumber, db).Id
+		headerTwoID := storageHelpers.CreateHeader(timestamp+1, blockNumber+1, db).Id
+
+		bidOneID = rand.Int()
+		bidTwoID = rand.Int()
+		flipIlkMetadata := types.GetValueMetadata(storage.Ilk, nil, types.Bytes32)
```
rmulhol: Any reason not to use `flip.IlkMetadata` here?

transformers/component_tests/queries/urn_bid_events_test.go
```diff
@@ -0,0 +1,85 @@
...
+		repo := flip.FlipStorageRepository{ContractAddress: flipAddress}
+		repo.SetDB(db)
+
+		// insert records used in join
+		ethIlkErr := repo.Create(diffID, headerOneID, flipIlkMetadata, test_helpers.FakeIlk.Hex)
+		Expect(ethIlkErr).NotTo(HaveOccurred())
+		ethBidUsrErr := repo.Create(diffID, headerOneID, flipBidUsrMetadata, usrOne)
+		Expect(ethBidUsrErr).NotTo(HaveOccurred())
+
+		flipBidUsrMetadata.Keys[constants.BidId] = strconv.Itoa(bidTwoID)
```
rmulhol: nbd but I'd be tempted to create a new metadata object for `bidTwoID` so that there's less risk of these tests failing if the order of operations are modified (e.g. if this assignment happened before the insert on line 49)

transformers/component_tests/queries/urn_bid_events_test.go
```diff
@@ -0,0 +1,85 @@
...
+	It("gets another urn's bid events", func() {
+		expectedFlip := []test_helpers.BidEvent{{ContractAddress: flipAddress, BidId: strconv.Itoa(bidTwoID)}}
+
+		var actualEvents []test_helpers.BidEvent
+		err := db.Select(&actualEvents, `SELECT contract_address, bid_id  FROM api.urn_bid_events($1, $2)`,
+			usrTwo, test_helpers.FakeIlk.Identifier)
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(actualEvents).To(ConsistOf(expectedFlip))
+	})
```
rmulhol: Reluctant to suggest this because these tests already require some weighty setup, but I think it'd be good to add a row for a different ilk and verify that we're correctly filtering on that field

rmulhol: One other test that might provide helpful documentation could mention that it provides results from multiple auctions for an urn (if that urn has been flipped several times)

transformers/component_tests/queries/urn_bid_events_test.go
Gslaughl: Nope! :D

PR #107, opened 2020-02-03
Updates vat_move to use addressFK

yaoandrew:


rmulhol (commented):
I believe this still requires migrations updates

PR #115, opened 2020-02-11
Creates and adds script to travis

yaoandrew:
This PR should check the docker.toml file vs the transformerExporter.go file. Should we potentially pass the config file in via env var? Its a pretty basic check. Other enhancements could be checking the toml file for each transformerName listed that there is a corresponding exporter block.

### DIFF
--- a/scripts/check_config.sh
+++ b/scripts/check_config.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+REPOPATH=$PWD
+exporterFile=$REPOPATH'/plugins/transformerExporter.go'
+exportedTransformers=( $( sed -n '/vdb-mcd/p' $exporterFile | cut -d' ' -f 1 | sort ) )
+
+configFile=$REPOPATH'/environments/docker.toml'
+configTransformers=( $(sed -n "/exporter\..*/p" $configFile | awk -F "." '{print $2}' | tr -d "]" | sort ) )
+
+if [ ${#exportedTransformers[@]} != ${#configTransformers[@]} ]; then
+  echo transformerExporter and config contain different transformer totals
+  exit 1
+fi
+
+for ((i=0; i<${#exportedTransformers[@]}; i++)); do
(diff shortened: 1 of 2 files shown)


environments/docker.toml
```diff
@@ -165,6 +159,12 @@
         migrations = "db/migrations"
         contracts = ["MCD_CAT"]
         rank = "0"
+    [exporter.cdp_manager]
```
rmulhol: Why move this? I kinda like having the storage transformers alphabetical and then the event transformers alphabetical, though would also be open to strict alphabetical

scripts/check_config.sh
```diff
@@ -0,0 +1,20 @@
...
+configFile=$REPOPATH'/environments/docker.toml'
+configTransformers=( $(sed -n "/exporter\..*/p" $configFile | awk -F "." '{print $2}' | tr -d "]" | sort ) )
+
+if [ ${#exportedTransformers[@]} != ${#configTransformers[@]} ]; then
+  echo transformerExporter and config contain different transformer totals
+  exit 1
+fi
+
+for ((i=0; i<${#exportedTransformers[@]}; i++)); do
+  if [ ${exportedTransformers[$i]} != ${configTransformers[$i]} ]; then
```
rmulhol: Does this depend on the order of the transformers being the same between `environments/docker.toml` and `plugins/transformerExporter.go`?

environments/docker.toml
yaoandrew: Ah.. I don't think I noticed that order. I had deleted that to test a failing case in travis and put it back out of order.

scripts/check_config.sh
yaoandrew: It does. I sort both collections when I create them. I guess the O(n**2) solution wouldn't be terrible since we don't expect either collection to be large and then we wouldn't have to worry about order.

rmulhol: 👍 🎉

PR #124, opened 2020-02-12
1189 add addresses to storage initializers

elizabethengelman:
companion vdb PR: https://github.com/makerdao/vulcanizedb/pull/38

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/howeyc/fsnotify v0.9.0 // indirect
 	github.com/jmoiron/sqlx v0.0.0-20181024163419-82935fac6c1a
 	github.com/magiconair/properties v1.8.0
-	github.com/makerdao/vulcanizedb v0.0.12-rc.1.0.20200212223442-9582deb86491
+	github.com/makerdao/vulcanizedb v0.0.12-rc.1.0.20200214164838-1c2e73816570
 	github.com/mattn/go-runewidth v0.0.6
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -4,12 +4,15 @@ go 1.12
 
 require (
 	github.com/ethereum/go-ethereum v1.9.9
+	github.com/go-stack/stack v1.8.0 // indirect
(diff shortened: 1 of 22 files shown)


transformers/storage/cat/initializer/initializer.go
```diff
@@ -27,6 +28,7 @@ import (
 
 var catAddress = constants.GetContractAddress("MCD_CAT")
 var StorageTransformerInitializer transformer.StorageTransformerInitializer = storage.Transformer{
+	Address:           common.HexToAddress(constants.GetContractAddress("MCD_CAT")),
 	HashedAddress:     types.HexToKeccak256Hash(constants.GetContractAddress("MCD_CAT")),
```
rmulhol: Not really a big deal but I'm wondering if we might want to only inject the `Address` and then make `HashedAddress` private/derive it from the `Address`. I think it'd be hard to make this mistake but it'd be a bummer if we accidentally passed in different constants to `HexToAddress` vs `HexToKeccak256Hash`

go.mod
```diff
@@ -4,12 +4,15 @@ go 1.12
...
+	github.com/go-stack/stack v1.8.0 // indirect
 	github.com/gogo/protobuf v1.1.1
 	github.com/howeyc/fsnotify v0.9.0 // indirect
 	github.com/jmoiron/sqlx v0.0.0-20181024163419-82935fac6c1a
 	github.com/magiconair/properties v1.8.0
-	github.com/makerdao/vulcanizedb v0.0.12-rc.1
+	github.com/makerdao/vulcanizedb v0.0.12-rc.1.0.20200212223442-9582deb86491
 	github.com/mattn/go-runewidth v0.0.6
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
```
rmulhol: I've noticed these two deps seeming to move in and out of the `go.mod` 🤔

transformers/storage/cat/initializer/initializer.go
elizabethengelman: Yep, I really like that idea! I'll make that change in this PR and in makerdao/vulcanizedb#38

go.mod
elizabethengelman: good call. i may leave them out of this PR the until we can get to the bottom of if this change is necessary/what we want

PR #125, opened 2020-02-12
Makes updates to dockerfile/README.md

yaoandrew:


dockerfiles/README.md
```diff
@@ -59,7 +59,7 @@ For execution on linux, replace instances of `host.docker.internal` with `localh
         - `--statediff.watchedaddresses <contract address>`
     - e.g.
     ```shell script
-        docker run -v /Users/elizabethengelman/Library/Ethereum:/root/.ethereum -p 8545:8545 -p 8546:8546 -p 30303:30303 geth-statediffing
```
elizabethengelman: I think we still need to include the `:/root/.ethereum` -> that is where the volume is being mounted on the geth docker container, so we should be able to rely on the fact that that directory will be consistent regardless of one's OS. So something like this: `</path/to/user/geth/directory>:/root/.ethereum`. Not sure how clear that is though 🤔

yaoandrew: Cool... I tried to figure out where to cutoff the path in the geth code but didn't see anything immediate.

PR #128, opened 2020-02-18
(VDB-1164) Copy over migrations from vDB

Gslaughl:
Companion to https://github.com/makerdao/vulcanizedb/pull/46

### DIFF
--- a/db/migrations/00001_create_nodes_table.sql
+++ b/db/migrations/00001_create_nodes_table.sql
@@ -6,7 +6,7 @@ CREATE TABLE eth_nodes
     genesis_block VARCHAR(66),
     network_id    NUMERIC,
     eth_node_id   VARCHAR(128),
-    UNIQUE (genesis_block, network_id, eth_node_id)
+    UNIQUE (genesis_block, network_id, eth_node_id, client_name)
 );
 
 COMMENT ON TABLE public.eth_nodes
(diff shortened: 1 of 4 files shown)


db/migrations/00001_create_nodes_table.sql
```diff
@@ -6,7 +6,7 @@ CREATE TABLE eth_nodes
     genesis_block VARCHAR(66),
     network_id    NUMERIC,
     eth_node_id   VARCHAR(128),
-    UNIQUE (genesis_block, network_id, eth_node_id)
+    UNIQUE (genesis_block, network_id, eth_node_id, client_name)
```
rmulhol: Should have mentioned this on the other PR but heads up that we need to avoid modifying earlier migrations starting asap if we're working against a deployed db - otherwise the changes won't take hold.

db/migrations/00001_create_nodes_table.sql
```diff
@@ -6,7 +6,7 @@ CREATE TABLE eth_nodes
...
 );
```
rmulhol: Perhaps worth modifying the go.mod to point at the branch with these changes (unless they're already on staging?) so that CI verifies any other changes on the vdb side avoid causing any test failures

db/migrations/00001_create_nodes_table.sql
Gslaughl: 🤦‍♂ oh boy, that's gonna take some getting used to haha

PR #129, opened 2020-02-18
Adds gomoderator docker to travis build

yaoandrew:
This PR contains a `Dockerfile` that is used in the `.travis.yml.` It should run `gomoderator` across both repos. Exit codes have been updated to be more explicit in this PR:
https://github.com/makerdao/vulcanizedb/pull/49

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_install:
 - echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
 - sudo apt-get update && sudo apt-get install yarn python3-pip -y
 - sudo pip3 install awscli
+- docker build dockerfiles/gomoderator
 script:
 - scripts/check_config.sh
 - env GO111MODULE=on make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
 - echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
 - sudo apt-get update && sudo apt-get install yarn python3-pip -y
 - sudo pip3 install awscli
-- docker build dockerfiles/gomoderator
+- docker build -f dockerfiles/gomoderator/Dockerfile .
 script:
 - scripts/check_config.sh
 - env GO111MODULE=on make test
--- a/dockerfiles/gomoderator/Dockerfile
+++ b/dockerfiles/gomoderator/Dockerfile
@@ -12,7 +12,6 @@ RUN git checkout $VDB_VERSION
 RUN go build
 WORKDIR /go/src/github.com/makerdao/vdb-mcd-transformers
 COPY . .
-RUN go build
 
 # app container
 FROM golang:alpine
--- a/dockerfiles/gomoderator/Dockerfile
+++ b/dockerfiles/gomoderator/Dockerfile
@@ -1,0 +1,38 @@
+FROM golang:alpine as builder
+
+RUN apk --update --no-cache add make git g++ linux-headers python3
+
+ARG VDB_VERSION=staging
+ENV GO111MODULE on
+
+WORKDIR /go/src/github.com/makerdao
+RUN git clone https://github.com/makerdao/vulcanizedb.git
+RUN git clone https://github.com/makerdao/vdb-mcd-transformers.git
+WORKDIR /go/src/github.com/makerdao/vulcanizedb
+RUN git checkout $VDB_VERSION
+RUN go build
+WORKDIR /go/src/github.com/makerdao/vdb-mcd-transformers
+RUN git checkout $VDB_VERSION
+RUN go build
+
+# app container
+FROM golang:alpine
+WORKDIR /go/src/github.com/makerdao/vulcanizedb
+
+# add certificates for node requests via https
+# bash for wait-for-it
+RUN apk update \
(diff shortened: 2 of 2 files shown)


dockerfiles/gomoderator/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ENV GO111MODULE on
+
+WORKDIR /go/src/github.com/makerdao
+RUN git clone https://github.com/makerdao/vulcanizedb.git
+RUN git clone https://github.com/makerdao/vdb-mcd-transformers.git
+WORKDIR /go/src/github.com/makerdao/vulcanizedb
+RUN git checkout $VDB_VERSION
+RUN go build
+WORKDIR /go/src/github.com/makerdao/vdb-mcd-transformers
+RUN git checkout $VDB_VERSION
```
yaoandrew: Does this seem right or would I want to run `vulcanizedb[staging] `vs. the `mcd` branch currently being built?

rmulhol: I think you probably want to stick with the mcd branch that's currently being built so that the status indicates whether a given feature branch/PR introduces new dep conflicts

.travis.yml
```diff
@@ -14,6 +14,7 @@ before_install:
 - curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
 - echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
 - sudo apt-get update && sudo apt-get install yarn
+- docker build dockerfiles/gomoderator
```
rmulhol: I think we probably need to run this in addition to building it if we want it to throw an error when the `gomoderator` detects conflicts

dockerfiles/gomoderator/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+FROM golang:alpine
+WORKDIR /go/src/github.com/makerdao/vulcanizedb
+
+# add certificates for node requests via https
+# bash for wait-for-it
+RUN apk update \
+        && apk upgrade \
+        && apk add --no-cache \
+        ca-certificates \
+        bash \
```
elizabethengelman: Super small, but i'm not sure if we need these two lines. I don't think we're using bash in this dockerfile (we're using it for the wait-for-it script for some other commands). And then based on the comment about the certificates, I _think_ that's just for if we're making node requests. Not 100% sure on that though.

dockerfiles/gomoderator/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+        && update-ca-certificates 2>/dev/null || true
+
+# add go so we can build the plugin
+RUN apk add --update --no-cache git g++ linux-headers python3
+
+# get access to go.mod files in directories
+COPY --from=builder /go/src/github.com/makerdao/vulcanizedb .
+COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/go.mod ./vdb/go.mod
```
elizabethengelman: Is `./vdb/go.mod` making it so that you're putting the vdb-mcd-transformers' go.mod in the same dir, so it's easier to call the gomoderator below?

Super nitpicky, but I wonder if we could change `./vdb/go.mod` to `./mcd-transformers/go.mod`? Don't feel super strongly about it, but it just confused me at first that this is the transformers' go.mod file, not the vulcanizedb one.

yaoandrew: yes on `./vdb/go.mod`. Is that also confusing? I couldn't figure out how to get the PATH working correctly but could try again if this is a confusing step.

Not nitpicky at all on the dir naming. That's why we have another set of 👀

.travis.yml
yaoandrew: Am I interpreting [this](https://travis-ci.com/makerdao/vdb-mcd-transformers/builds/150565668) wrong? 
On line 1688 it says `Step 20/20 : RUN python3 scripts/gomoderator.py ./ vdb/` 
followed up by line 1867 `no conflicts to resolve`. 

Is this just running in some intermediate container?

rmulhol: Nope, you're right 👍 

I didn't realize we were calling it with `RUN` in the build step

dockerfiles/gomoderator/Dockerfile
yaoandrew: Cool yeah, I think we can get rid of all of it.

dockerfiles/gomoderator/Dockerfile
elizabethengelman: Copying the vdb-mcd-transformers go.mod over to the current directory makes sense to me!

dockerfiles/gomoderator/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM golang:alpine as builder
+
+RUN apk --update --no-cache add make git g++ linux-headers python3
+
+ARG VDB_VERSION=staging
+ENV GO111MODULE on
+
+WORKDIR /go/src/github.com/makerdao
+RUN git clone https://github.com/makerdao/vulcanizedb.git
+RUN git clone https://github.com/makerdao/vdb-mcd-transformers.git
```
rmulhol: I don't think we need to clone this repo (and probably don't want to so we avoid being on the wrong branch).

Instead maybe we can just `COPY . .` (may want to enter the appropriate workdir first)

PR #131, opened 2020-02-20
(VDB-1207) Remove non-essential smart comments

rmulhol:
This PR has morphed into removing smart tags for anything that's not a mutation and adding smart tags to omit mutations where we weren't doing so already. 

Work on using tags for documentation/omitting specific read operations (e.g. attributes) has moved to https://github.com/makerdao/vdb-postgraphile/pull/4 where we'll have the ability to make continuous changes without modifying the database.

### DIFF
--- a/db/migrations/20180809214843_create_ilks.sql
+++ b/db/migrations/20180809214843_create_ilks.sql
@@ -7,7 +7,7 @@ CREATE TABLE maker.ilks
 );
 
 COMMENT ON TABLE maker.ilks
-    IS E'@name raw_ilks';
+    IS E'@name raw_ilks\nCollateral type.';
 
 -- +goose Down
 DROP TABLE maker.ilks;
--- a/db/migrations/20180809214843_create_ilks.sql
+++ b/db/migrations/20180809214843_create_ilks.sql
@@ -6,8 +6,5 @@ CREATE TABLE maker.ilks
     identifier TEXT UNIQUE NOT NULL
 );
 
-COMMENT ON TABLE maker.ilks
-    IS E'@name raw_ilks\nCollateral type.';
-
 -- +goose Down
 DROP TABLE maker.ilks;
--- a/db/migrations/20180831002535_create_vat_file.sql
+++ b/db/migrations/20180831002535_create_vat_file.sql
@@ -10,6 +10,9 @@ CREATE TABLE maker.vat_file_ilk
     UNIQUE (header_id, log_id)
 );
 
+COMMENT ON TABLE maker.vat_file_ilk
+    IS E'Note event emitted when file(bytes32,bytes32,uint256) is called on Vat contract.';
+
 CREATE INDEX vat_file_ilk_header_index
     ON maker.vat_file_ilk (header_id);
 CREATE INDEX vat_file_ilk_log_index
@@ -26,6 +29,9 @@ CREATE TABLE maker.vat_file_debt_ceiling
     data      NUMERIC,
     UNIQUE (header_id, log_id)
 );
+
+COMMENT ON TABLE maker.vat_file_debt_ceiling
+    IS E'Note event emitted when file(bytes32,uint256) is called on Vat contract.';
 
 CREATE INDEX vat_file_debt_ceiling_header_index
     ON maker.vat_file_debt_ceiling (header_id);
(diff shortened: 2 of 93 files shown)


db/migrations/20180809214843_create_ilks.sql
```diff
@@ -7,7 +7,7 @@ CREATE TABLE maker.ilks
 );
 
 COMMENT ON TABLE maker.ilks
-    IS E'@name raw_ilks';
+    IS E'@name raw_ilks\nCDP type';
```
Gslaughl: Should this say "Collateral type"? Although idk maybe "CDP type" will make more sense to people using the API.

db/migrations/20180831002535_create_vat_file.sql
```diff
@@ -10,6 +10,9 @@ CREATE TABLE maker.vat_file_ilk
     UNIQUE (header_id, log_id)
 );
 
+COMMENT ON TABLE maker.vat_file_ilk
+    IS E'Note event emitted when file(bytes32,bytes32,uint256) is called on Vat contract';
```
Gslaughl: 👍

db/migrations/20190309165205_create_urn_history_query.sql
```diff
@@ -42,5 +42,8 @@ $$
     STABLE;
 -- +goose StatementEnd
 
+COMMENT ON FUNCTION api.all_urn_states(ilk_identifier TEXT, urn_identifier TEXT, block_height BIGINT, max_results INTEGER, result_offset INTEGER)
+    IS E'Get all historical states for an Urn prior to a given block. Ilk identifier (e.g. "ETH-A") and Urn identifier (e.g."") are required. Block height, limit, and offset are optional. Block height defaults to most recent block. Limit defaults to null (no limit). Offset defaults to 0.';
```
Gslaughl: Empty example Urn identifier ;)

db/migrations/20190715152452_create_single_flip_query.sql
```diff
@@ -19,6 +19,9 @@ CREATE TYPE api.flip_bid_snapshot AS
     updated      TIMESTAMP
 );
 
+COMMENT ON TYPE api.flip_bid_snapshot
+    IS E'Historical snapshots of the state of auctions on the Flip contracts, with data about associated Urn.';
```
rmulhol: I found it pretty difficult to distinguish between `maker.flip` and `api.flip_bid_snaphot` in the docs, and there are similar ambiguities (for me) on the other auctions contracts as well as urns. Wondering whether part of it might be that we were originally only going to expose the `api` schema and are now planning to expose all. If so, maybe it's worthwhile to consolidate? Alternatively, if anyone with a better understanding can chime in I'd be happy to update the docs.

Gslaughl: Yeah you're exactly right; the `maker.flip/flap/flop` tables were meant to be sort of intermediary caches, not exposed to the client. The client-facing functions like `api.get_flip`, `api.all_flips` etc. (which return `snapshot`s) make use of those tables, and also do a couple more things at request-time, like determine whether the bid is `dealt`, and (in `flip`'s case) figure out which ilk and urn are associated with the bid. Also, since only the `api.-bid_snapshot` types were meant to be client-facing, they have "computed columns", while the ones in the `maker` schema don't.

I agree that it would probably make sense to do some consolidation. To that end, it seems feasible to remove all the request-time logic that `get_flip/flap/flop` are doing, and instead add the bids' `dealt`, `ilk`, and `urn` data to our trigger tables. Then we could probably trim the API down to just using a single type for each kind of bid.

rmulhol:
Currently working on removing all smart tags except for `@omit` tags on mutations (which I think we should never expose).

Instead I'm planning to move all smart tags over to the vdb-postgraphile repo so that we can modify them without touching migrations: https://github.com/makerdao/vdb-postgraphile/pull/4

db/migrations/20181114200131_create_vat_storage_tables.sql
```diff
@@ -12,7 +12,7 @@ CREATE INDEX vat_debt_header_id_index
     ON maker.vat_debt (header_id);
 
 COMMENT ON TABLE maker.vat_debt
-    IS E'@omit';
+    IS E'';
```
elizabethengelman: i think we can probably remove this line altogether now, right?

PR #132, opened 2020-02-21
Update postgres version in travis build

elizabethengelman:


rmulhol:
Looks like the build failure also included unstaged changes to the `go.mod` - might need to update that file as well? Wondering if it's adding/removing those two deps that keep coming and going

PR #139, opened 2020-03-02
Vdb 1212 val storage

yaoandrew:
This still needs DB migrations written but I wanted to get some feedback on the datatypes before writing those. Noted in comments below.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 11.6
--- Dumped by pg_dump version 11.6
+-- Dumped from database version 12.1
+-- Dumped by pg_dump version 12.1
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
@@ -825,7 +825,7 @@ $$;
 
 SET default_tablespace = '';
 
-SET default_with_oids = false;
+SET default_table_access_method = heap;
 
 --
 -- Name: ilk_snapshot; Type: TABLE; Schema: api; Owner: -
@@ -10030,13 +10030,6 @@ CREATE TABLE maker.osm_change (
 
 
 --
--- Name: TABLE osm_change; Type: COMMENT; Schema: maker; Owner: -
---
-
-COMMENT ON TABLE maker.osm_change IS 'Note event emitted when change is called on OSM/PIP contract(s)';
-
-
---
 -- Name: osm_change_id_seq; Type: SEQUENCE; Schema: maker; Owner: -
 --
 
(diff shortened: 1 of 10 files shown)


transformers/storage/val/initializer/initializer.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	mcdStorage "github.com/makerdao/vdb-mcd-transformers/transformers/storage"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/storage/val"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/storage"
+)
+
+var valAddress = constants.GetContractAddress("MCD_VAL")
+var StorageTransformerInitializer storage.TransformerInitializer = storage.Transformer{
+	Address:           common.HexToAddress(constants.GetContractAddress("MCD_VAL")),
+	StorageKeysLookup: storage.NewKeysLookup(val.NewKeysLoader(&mcdStorage.MakerStorageRepository{}, valAddress)),
```
yaoandrew: Not sure what address to use here. This `MCD_VAL` is just a placeholder.

transformers/storage/val/keys_loader.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	Has = "has"
+	Val = "val"
+)
+
+var (
+	HasKey      = common.HexToHash(vdbStorage.IndexZero)
+	HasMetadata = types.GetValueMetadata(Has, nil, types.Uint256)
```
yaoandrew: I suppose I should add types.bool here? Just felt strange that it was the first one.

transformers/storage/val/initializer/initializer.go
Gslaughl: Imo it would make sense to just use the actual addresses of the existing Medianizers for now, although we'll probably need to use a different method when we wire these transformers up to track new contracts on the fly.

i.e. 0x64DE91F5A373Cd4c28de3600cB34C7C6cE410C85 and 0x18B4633D6E39870f398597f3c1bA8c4A41294966

transformers/test_data/config_values.go
```diff
@@ -36,6 +36,7 @@ func OsmAddresses() []string {
 }
 func PotAddress() string        { return checksum(constants.GetContractAddress("MCD_POT")) }
 func SpotAddress() string       { return checksum(constants.GetContractAddress("MCD_SPOT")) }
+func ValAddress() string        { return checksum(constants.GetContractAddress("MCD_VAL")) }
```
yaoandrew: Will update this to the medianizer addresses

db/schema.sql
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 11.6
--- Dumped by pg_dump version 11.6
+-- Dumped from database version 12.1
+-- Dumped by pg_dump version 12.1
```
yaoandrew: 😞 Forgot to rollback postgres on my machines.

PR #143, opened 2020-03-11
(VDB-1228) Median kiss + diss transformers

gslaughl:


test_config/test_config.go
```diff
@@ -91,7 +91,11 @@ func CleanTestDB(db *postgres.DB) {
 	}
 
 	for _, query := range wipeTableQueries {
-		db.MustExec(query)
+		_, wipeErr := db.Exec(query)
+		if wipeErr != nil {
+			fmt.Println("Unable to execute query: ", query)
+			fmt.Println(wipeErr)
+		}
```
Gslaughl: I had to make `CleanTestDB` stop panicking on error in order to get the tests consistently passing in Travis. I can't figure out exactly why it keeps erroring out, nor can I reproduce the error locally. So, I'll need to dig into it a little more before merging, but planning to hold off on that since these transformers aren't top-priority right now.

transformers/events/median_diss/single/single_suite_test.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"io/ioutil"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
+)
+
+func TestVow(t *testing.T) {
```
yaoandrew: Updated function name needed here.

db/migrations/20200203102132_create_bid_event_table.sql
```diff
@@ -1,7 +1,7 @@
 -- +goose Up
 CREATE TABLE maker.bid_event
 (
-    log_id           BIGINT PRIMARY KEY REFERENCES event_logs (id) ON DELETE CASCADE,
+    log_id           BIGINT PRIMARY KEY REFERENCES event_logs (id) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED,
```
Gslaughl: This line fixes the flaky test we've been seeing, by specifying that the `CASCADE DELETE` should happen at the end of the transaction. This way we won't run into the issue where a trigger is trying to update a row that was already deleted, because the row won't have been deleted yet.

[Here](https://www.postgresql.org/docs/9.1/sql-set-constraints.html) are some docs about deferrable constraints if you're interested

rmulhol: Don't think we want to modify the earlier migration here. Is it possible to update this column in a new migration?

db/migrations/20200310105940_create_median_kiss_single.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE INDEX median_kiss_single_address_index
+    ON maker.median_kiss_single (address_id);
+CREATE INDEX median_kiss_single_msg_sender_index
+    ON maker.median_kiss_single (msg_sender);
+CREATE INDEX median_kiss_single_a_index
+    ON maker.median_kiss_single (a);
+
+
+-- +goose Down
+DROP TABLE maker.median_kiss_single;
```
rmulhol, line 26: Not an issue right now but _if_ this is merged after the backfill stuff then we'll need updated timestamps on here to make sure they come after that. Mostly just posting as a general reminder to make sure all new migrations are the latest against staging for every PR

db/migrations/20200203102132_create_bid_event_table.sql
Gslaughl: Haha oh yeah 🤦‍♂ 

Can def put this into a separate migration

transformers/integration_tests/median_diss_single_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/libraries/shared/fetcher"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+// TODO: Update once a diss event has happened
```
rmulhol, line 18: obsolete?

rmulhol (approved):
LGTM! :shipit:

transformers/integration_tests/median_diss_single_test.go
Gslaughl: Looking in etherscan I still can't find any diss events, unfortunately

PR #144, opened 2020-03-11
(VDB-1218) Add function for inserting headers

rmulhol:
- Compatibility with upstream vdb changes

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -7194,6 +7194,54 @@ COMMENT ON FUNCTION public.flop_bid_time_created(address_id integer, bid_id nume
 
 
 --
+-- Name: get_or_create_header(bigint, character varying, jsonb, numeric, integer); Type: FUNCTION; Schema: public; Owner: -
+--
+
+CREATE FUNCTION public.get_or_create_header(block_number bigint, hash character varying, raw jsonb, block_timestamp numeric, eth_node_id integer) RETURNS integer
+    LANGUAGE plpgsql
+    AS $$
+DECLARE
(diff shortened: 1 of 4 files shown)


db/schema.sql
```diff
@@ -10029,13 +10077,6 @@ CREATE TABLE maker.osm_change (
 );
 
 
---
--- Name: TABLE osm_change; Type: COMMENT; Schema: maker; Owner: -
---
-
-COMMENT ON TABLE maker.osm_change IS 'Note event emitted when change is called on OSM/PIP contract(s)';
-
-
```
rmulhol: Idk why this is showing up as part of the PR but seems like maybe a migration was updated without the schema getting a corresponding update as we removed comments?

go.mod
```diff
@@ -4,15 +4,12 @@ go 1.12
 
 require (
 	github.com/ethereum/go-ethereum v1.9.9
-	github.com/go-stack/stack v1.8.0 // indirect
 	github.com/gogo/protobuf v1.1.1
 	github.com/howeyc/fsnotify v0.9.0 // indirect
 	github.com/jmoiron/sqlx v0.0.0-20181024163419-82935fac6c1a
 	github.com/magiconair/properties v1.8.0
-	github.com/makerdao/vulcanizedb v0.0.13-rc.2
+	github.com/makerdao/vulcanizedb v0.0.13-rc.2.0.20200309155515-0dcc44e856be
```
rmulhol: will switch back to staging and/or a release if we get it merged on the vdb side

PR #145, opened 2020-03-17
VDB-1227 Median drop event transformer

yaoandrew:
I'll rebase this after Gabe merges #143. Sorry for the clutter for now but it made sense to piggyback off that for some his fine work on medianizer.


transformers/events/median_drop/transformer_test.go
```diff
@@ -0,0 +1,112 @@
...
+		Expect(aErr).NotTo(HaveOccurred())
+
+		expectedModel := test_data.MedianDropModelWithOneAccount()
+		contractAddressID, contractAddressErr := shared.GetOrCreateAddress(test_data.MedianDropLogWithOneAccount.Log.Address.String(), db)
+		Expect(contractAddressErr).NotTo(HaveOccurred())
+		msgSenderAddressID, msgSenderAddressErr := shared.GetOrCreateAddress(test_data.MedianDropLogWithOneAccount.Log.Topics[1].Hex(), db)
+		Expect(msgSenderAddressErr).NotTo(HaveOccurred())
+		expectedModel.ColumnValues[event.AddressFK] = contractAddressID
+		expectedModel.ColumnValues[constants.MsgSenderColumn] = msgSenderAddressID
+		expectedModel.ColumnValues[constants.AColumn] = addressID1
```
Gslaughl: I think that we might need to either use a Postgres array to hold these addresses (rather than storing them in a fixed number of columns), or use an additional table to store the relation between addresses and `drop` events. The reason being that I'm not sure we can rely on the address array being passed to `drop` only having a length of 4.

transformers/shared/transformer_utils.go
```diff
@@ -52,3 +56,22 @@ func VerifyLog(log types.Log, expectedNumTopics int, isDataRequired bool) error
 	}
 	return nil
 }
+
+func GetLogNoteData(startingIndex int, eventLogData []byte, db *postgres.DB) ([]int64, error) {
+	var addressIDs []int64
+	for j := startingIndex; j <= lastDataIndex; j++ {
+		logDataBytes, _ := GetLogNoteArgumentAtIndex(j, eventLogData)
```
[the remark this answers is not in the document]
yaoandrew: Cool, I'll check that out. I tried passing a long array of accounts to `drop` which basically looked like it truncated everything after the 4th account so I based a lot of assumptions on that. I'll check out truffle to see what I can find.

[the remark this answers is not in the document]
Gslaughl: Yeah, I could be wrong but to me it seems important that we transform every address in the array; otherwise our storage records will be incomplete

db/migrations/20200316235856_median_drop.sql
```diff
@@ -0,0 +1,26 @@
+-- +goose Up
+CREATE TABLE maker.median_drop
+(
+    id         SERIAL PRIMARY KEY,
+    log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    address_id INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    msg_sender INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    a        INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
```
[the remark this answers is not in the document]
yaoandrew: added a0 - a3

db/migrations/20200310105940_create_median_kiss_single.sql
```diff
@@ -0,0 +1,26 @@
+-- +goose Up
+CREATE TABLE maker.median_kiss_single
```
rmulhol, line 2: Sorry to be that guy - stopping in to block merging until timestamps are updated 😬

test_config/test_config.go
```diff
@@ -91,7 +91,11 @@ func CleanTestDB(db *postgres.DB) {
 	}
 
 	for _, query := range wipeTableQueries {
-		db.MustExec(query)
+		_, wipeErr := db.Exec(query)
+		if wipeErr != nil {
+			fmt.Println("Unable to execute query: ", query)
+			fmt.Println(wipeErr)
+		}
```
Gslaughl, line 98: I think this is a remnant from when I was troubleshooting a flaky test-- should be fine to omit these changes now

transformers/events/median_drop/transformer_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+		a3Bytes, a3Err := shared.GetLogNoteArgumentAtIndex(4, test_data.MedianDropLogWithFiveAccounts.Log.Data)
+		Expect(a3Err).NotTo(HaveOccurred())
+		address2 := common.BytesToAddress(a3Bytes).Hex()
+
+		a4Bytes, a4Err := shared.GetLogNoteArgumentAtIndex(5, test_data.MedianDropLogWithFiveAccounts.Log.Data)
+		Expect(a4Err).NotTo(HaveOccurred())
+		address3 := common.BytesToAddress(a4Bytes).Hex()
+
+		expectedModel := test_data.MedianDropModelWithFiveAccounts()
```
rmulhol, line 64: Is this model named correctly? Seems like 4 rather than 5 accounts, right?

transformers/events/median_drop/transformer.go
```diff
@@ -0,0 +1,65 @@
...
+			return nil, shared.ErrCouldNotCreateFK(contractAddressErr)
+		}
+
+		msgSender := log.Log.Topics[1].Hex()
+		msgSenderAddressID, msgSenderAddressErr := shared.GetOrCreateAddress(msgSender, db)
+		if msgSenderAddressErr != nil {
+			return nil, shared.ErrCouldNotCreateFK(msgSenderAddressErr)
+		}
+
+		aLength, aLengthErr := strconv.ParseUint(log.Log.Topics[3].Hex(), 0, 64)
```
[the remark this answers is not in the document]
Gslaughl: Not sure, but I think that when a `LogNote` event gets an array, `Topics[2]` is the size of 1 element of the array in bytes. For all the events I've seen, it's always just the number 20 (padded out to fit the topic ofc), which makes sense since the only arrays I've seen as arguments are arrays of addresses, which are 20 bytes long.

transformers/events/median_drop/transformer_test.go
Gslaughl: I believe the idea with this test is to demonstrate that if `drop` was passed an array of 5+ addresses, then the event truncates the array partway through the 5th address-- looking at the log's `Data`, this appears to be an example of that kind of event log. Might be worth updating the test's docstring to clarify what's happening.

transformers/events/median_drop/transformer_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+		_, err := transformer.ToModels(constants.MedianABI(), []core.EventLog{badLog}, db)
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("convert a log with 4 accounts to an insertion model", func() {
+		models, err := transformer.ToModels(constants.MedianABI(), []core.EventLog{test_data.MedianDropLogWithFiveAccounts}, db)
+		Expect(err).NotTo(HaveOccurred())
+
+		a1Bytes, aErr := shared.GetLogNoteArgumentAtIndex(2, test_data.MedianDropLogWithFiveAccounts.Log.Data)
```
Gslaughl, line 48: If we're zero-indexing the address name on line 50, then I'd be a fan of zero-indexing `a1Bytes` here as well

environments/docker.toml
```diff
@@ -283,6 +284,13 @@
         migrations = "db/migrations"
         contracts = ["MEDIAN_BAT", "MEDIAN_ETH"]
         rank = "0"
+    [exporter.median_drop]
+    	path = "transformers/events/media_drop/initializer"
```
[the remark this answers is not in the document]
yaoandrew: Nice catch!

transformers/events/median_drop/transformer_test.go
rmulhol: Ah makes sense - I guess we probably don't need that data (and could potentially grab it off the log if we wanted it in the future?). Cool that the log contains array length + item length - I imagine that could prove useful for general purpose helpers built around parsing arrays in events

transformers/events/median_drop/transformer.go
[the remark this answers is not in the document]
rmulhol: Ah makes sense - I guess we probably don't need that data (and could potentially grab it off the log if we wanted it in the future?). Cool that the log contains array length + item length - I imagine that could prove useful for general purpose helpers built around parsing arrays in events

transformers/events/median_drop/transformer_test.go
rmulhol: word yeah sounds like a docstring update is in order - was just confused if we were dealing with four or five entries in the array

transformers/events/median_drop/transformer.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
+)
+
+type Transformer struct{}
+
+func (Transformer) ToModels(_ string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
+	var (
+		models []event.InsertionModel
+	)
```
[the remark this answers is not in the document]
yaoandrew: Makes sense. That was leftover from a refactor.

transformers/events/median_drop/transformer_test.go
yaoandrew: :point_up: exactly this. I forgot to update the docstring

db/migrations/20200407131156_create_median_drop.sql
```diff
@@ -0,0 +1,25 @@
...
+    ON maker.median_drop (log_id);
+CREATE INDEX median_drop_header_index
+    ON maker.median_drop (header_id);
+CREATE INDEX median_drop_address_index
+    ON maker.median_drop (address_id);
+CREATE INDEX median_drop_msg_sender_index
+    ON maker.median_drop (msg_sender);
+-- +goose Down
+
+DROP TABLE maker.median_drop;
```
rmulhol, line 25: sorry to be a pain, but I think it'd be good to include the changes to the db/schema in this PR if possible. Happy to pull down the branch and add it myself if you're running a version other than 11.6 and that'd be helpful

PR #149, opened 2020-03-31
Vdb 1226 median lift

yaoandrew:


db/migrations/20200409194015_create_median_lift.sql
```diff
@@ -0,0 +1,24 @@
+-- +goose Up
```
rmulhol, line 1: Dropping in to request a timestamp update 😬

transformers/events/median_lift/transformer.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
+)
+
+type Transformer struct{}
+
+func (Transformer) ToModels(_ string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
+	var (
```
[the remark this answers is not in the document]
yaoandrew: Sorry these keep popping up. I feel like something is happening in my rebase.

transformers/events/median_lift/median_lift_suite_test.go
```diff
@@ -0,0 +1,19 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
+)
+
+func TestMedianLift(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "MedianLift Suite")
+}
+
+var _ = BeforeSuite(func() {
```
rmulhol, line 17: nbd but I don't _think_ we need this anymore if we're not logging in the transformer

transformers/integration_tests/median_lift_test.go
```diff
@@ -0,0 +1,87 @@
...
+		blockNumber := int64(8936530)
+		medianLiftConfig.StartingBlockNumber = blockNumber
+		medianLiftConfig.EndingBlockNumber = blockNumber
+
+		test_config.CleanTestDB(db)
+
+		header, headerErr := persistHeader(db, blockNumber, blockChain)
+		Expect(headerErr).NotTo(HaveOccurred())
+
+		// TODO: fetch event from blockchain once one exists
```
rmulhol, line 36: I'd be a fan of marking this test as pending if we don't have an actual event to lookup from the chain, otherwise this is basically just reproducing the transformer test, right?

transformers/integration_tests/median_lift_test.go
```diff
@@ -0,0 +1,87 @@
...
+		transformer := event.ConfiguredTransformer{
+			Config:      medianLiftConfig,
+			Transformer: median_lift.Transformer{},
+		}.NewTransformer(db)
+
+		transformErr := transformer.Execute(eventLogs)
+		Expect(transformErr).NotTo(HaveOccurred())
+
+		var dbResults []MedianLiftModel
+		err := db.Select(&dbResults, `SELECT address_id, msg_sender, a_length FROM maker.median_lift ORDER BY id`)
```
rmulhol, line 55: What is the `ORDER BY` accomplishing here?

transformers/integration_tests/median_lift_test.go
```diff
@@ -0,0 +1,87 @@
...
+		Expect(dbResult.ALength).To(Equal("5"))
+
+		var addresses []string
+		addressesError := db.Get(pq.Array(&addresses), `SELECT a FROM maker.median_lift ORDER BY id`)
+		Expect(addressesError).NotTo(HaveOccurred())
+		Expect(addresses).To(ConsistOf(
+			"0x6bDbc0ccC17d72a33Bf72a4657781a37DC2aa94E",
+			"0x26c45f7B0E456E36fC85781488A3CD42A57CcbD2",
+			"0x20c576F989EE94E571F027b30314aCF709267F7C",
+			"0xFCb1fB52E114b364B3Aab63d8a6f65Fe8dcbeF9D",
```
rmulhol, line 78: Apologies for being out of the loop here but have we discussed whether we might want this array to contain foreign keys rather than text? Wondering if we'll want to use these addresses for spinning up/dropping medianizer transformers and, if so, whether we might need the hashed address from that table

rmulhol: Can totally handle this later though, especially if there are other merged batch transformers already following this pattern

Gslaughl: I could be mistaken, but I don't _think_ these addresses correspond to contracts we'll want to track. I initially thought it would be a better idea to store the actual addresses rather than a FK, since Postgres can't enforce referential integrity on arrays of FKs. So it just seemed like going down that road would be asking for trouble (out-of-date references etc). Definitely down to rethink that decision though.

transformers/integration_tests/median_lift_test.go
Gslaughl: I think the main difference is that the transformer test doesn't interact with the DB at all-- it still passes even if there's no migration adding these events' table. Personally I'm indifferent whether we mark the whole test pending or have a 'partial' test like this one, just thought it was worth pointing out.

rmulhol: ah good point 👍 

for that matter this probably means we don't have tests that verify inserts work for all the integration tests that are currently pending 😱

transformers/integration_tests/median_lift_test.go
rmulhol: Cool, the referential integrity thing is valid in it's own right, and 💯 if we don't need to track the addresses for secondary reasons

transformers/integration_tests/median_lift_test.go
yaoandrew: Err.. nothing.

transformers/integration_tests/median_lift_test.go
yaoandrew: ha ha.. yeah, I think I'm going to make this pending since we may forget to update it when its not pending.

PR #150, opened 2020-03-31
(VDB-1230) Create kiss + diss batch event transformers

Gslaughl:
`kiss` and `diss` are basically carbon copies, aside from the words "kiss" and "diss" being different. I wonder if it would be worth trying to reduce duplication between events with very similar signatures; as it stands, that would take some pretty serious refactoring.


transformers/integration_tests/median_diss_batch_test.go
```diff
@@ -0,0 +1,87 @@
...
+		blockNumber := int64(8936530)
+		medianDissConfig.StartingBlockNumber = blockNumber
+		medianDissConfig.EndingBlockNumber = blockNumber
+
+		test_config.CleanTestDB(db)
+
+		header, headerErr := persistHeader(db, blockNumber, blockChain)
+		Expect(headerErr).NotTo(HaveOccurred())
+
+		// TODO: fetch event from blockchain once one exists
```
Gslaughl, line 36: Thought it would be a good idea to have at least a partial integration test in this instance, since the transformer test never actually interacts with the DB, and I wanted to make sure the new Postgres array stuff worked as expected.

transformers/shared/transformer_utils.go
```diff
@@ -52,3 +54,19 @@ func VerifyLog(log types.Log, expectedNumTopics int, isDataRequired bool) error
...
+	maxVisibleAddresses := 4
+	endIndex := startingIndex + int(math.Min(float64(maxVisibleAddresses), float64(arrayLength))-1)
+	var addresses []string
+	for i := startingIndex; i <= endIndex; i++ {
+		logDataBytes, logNoteErr := GetLogNoteArgumentAtIndex(i, eventLogData)
+		if logNoteErr != nil {
+			return nil, logNoteErr
+		}
+		addressHex := common.BytesToAddress(logDataBytes).Hex()
+		addresses = append(addresses, addressHex)
```
Gslaughl, line 69: Worth noting that we're never actually adding the addresses in this list to `public.addresses`. We could hypothetically return an array of address FKs rather than the addresses themselves, but Postgres can't enforce referential integrity on an array of FKs, so I thought it would be safer to just use the raw addresses.

Maybe it would make sense to add a `GetOrCreateAdress` line in here, just so we have it in the `addresses` table, even if we aren't referencing it? But I figured that would be happening in the storage transformer anyway, so there's not much need to do it here. Interested to hear people's thoughts.

db/migrations/20200329134301_create_median_diss_batch.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose Up
+CREATE TABLE maker.median_diss_batch
+(
+    id         SERIAL PRIMARY KEY,
+    log_id     BIGINT     NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    header_id  INTEGER    NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    address_id INTEGER    NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    msg_sender INTEGER    NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    a_length   INTEGER    NOT NULL,
```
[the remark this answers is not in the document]
yaoandrew: Sounds good. I'll add to lift and drop as well.

db/migrations/20200331102914_create_median_kiss_batch.sql
```diff
@@ -0,0 +1,25 @@
...
+CREATE INDEX median_kiss_batch_header_index
+    ON maker.median_kiss_batch (header_id);
+CREATE INDEX median_kiss_batch_address_index
+    ON maker.median_kiss_batch (address_id);
+CREATE INDEX median_kiss_batch_msg_sender_index
+    ON maker.median_kiss_batch (msg_sender);
+
+
+-- +goose Down
+DROP TABLE maker.median_kiss_batch;
```
yaoandrew, line 25: Its probably redundant when we `DROP` the table, but in other migrations we seem to be adding a `DROP` for each index as well in the rollback commands.

transformers/events/median_diss/batch/transformer_test.go
```diff
@@ -0,0 +1,94 @@
...
+		a2Bytes, a2Err := shared.GetLogNoteArgumentAtIndex(4, test_data.MedianDissBatchLogFiveAddresses.Log.Data)
+		Expect(a2Err).NotTo(HaveOccurred())
+		address2 := common.BytesToAddress(a2Bytes).Hex()
+		a3Bytes, a3Err := shared.GetLogNoteArgumentAtIndex(5, test_data.MedianDissBatchLogFiveAddresses.Log.Data)
+		Expect(a3Err).NotTo(HaveOccurred())
+		address3 := common.BytesToAddress(a3Bytes).Hex()
+
+		expectedModel.ColumnValues[event.AddressFK] = contractAddressID
+		expectedModel.ColumnValues[constants.MsgSenderColumn] = msgSenderAddressID
+		expectedModel.ColumnValues[constants.AColumn] = pq.Array([]string{address0, address1, address2, address3})
```
yaoandrew, line 90: :100:

transformers/events/median_kiss/batch/transformer.go
```diff
@@ -0,0 +1,68 @@
...
+			return nil, shared.ErrCouldNotCreateFK(contractAddressErr)
+		}
+
+		msgSender := log.Log.Topics[1].Hex()
+		msgSenderAddressID, msgSenderAddressErr := shared.GetOrCreateAddress(msgSender, db)
+		if msgSenderAddressErr != nil {
+			return nil, shared.ErrCouldNotCreateFK(msgSenderAddressErr)
+		}
+
+		aLength, aLengthErr := strconv.ParseUint(log.Log.Topics[3].Hex(), 0, 64)
```
yaoandrew, line 35: 🥇

transformers/events/median_kiss/batch/transformer.go
```diff
@@ -0,0 +1,68 @@
...
+				constants.ALengthColumn,
+				constants.AColumn,
+			},
+			ColumnValues: event.ColumnValues{
+				event.HeaderFK:            log.HeaderID,
+				event.LogFK:               log.ID,
+				event.AddressFK:           contractAddressID,
+				constants.MsgSenderColumn: msgSenderAddressID,
+				constants.ALengthColumn:   strconv.FormatUint(aLength, 10),
+				constants.AColumn:         pq.Array(aAddresses),
```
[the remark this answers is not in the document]
yaoandrew: 👍 if this is a problem, maybe there's a way in the `sqlx` library to support array input.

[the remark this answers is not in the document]
Gslaughl: Good call, I forgot about that library. I'll see if that would be a better alternative

db/migrations/20200331102914_create_median_kiss_batch.sql
Gslaughl: Yeah, I've moved away from adding those statements in newer code. Like you said, they seem to be totally unnecessary.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/gogo/protobuf v1.1.1
 	github.com/howeyc/fsnotify v0.9.0 // indirect
 	github.com/jmoiron/sqlx v0.0.0-20181024163419-82935fac6c1a
+	github.com/lib/pq v1.0.0
```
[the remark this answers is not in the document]
yaoandrew: Is there a reason we are using v1.0.0 vs v.1.3.0?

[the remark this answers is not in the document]
Gslaughl: Nope, GoLand did that automatically. I'd be fine with bumping the version

transformers/shared/transformer_utils.go
Gslaughl: Good call-- definitely worth thinking about, but so far I'm not really finding that it's any tougher when the addresses aren't in `public.addresses`, since we pretty much always interact with that table using the `GetOrCreateAddress` function.

And on the `storageRepository` side of things, we ultimately need a list of addresses (not address IDs) for the key loader, so it's pretty easy to just query them directly from the array in `median_kiss_batch`/`median_diss_batch` without doing any joins or anything.

db/migrations/20200329134301_create_median_diss_batch.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose Up
+CREATE TABLE maker.median_diss_batch
```
rmulhol, line 2: Looks like this needs an updated timestamp

Gslaughl: Yup! I was just about to do that :) thanks for looking out though

db/migrations/20200331102914_create_median_kiss_batch.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose Up
+CREATE TABLE maker.median_kiss_batch
```
rmulhol, line 2: also true of this file

PR #151, opened 2020-03-31
(VDB-1249) Add Dockerfile for backfill_storage

rmulhol:
Pointed at the corresponding vDB branch but can switch to staging/release once merged/cut

### DIFF
--- a/dockerfiles/README.md
+++ b/dockerfiles/README.md
@@ -64,21 +64,21 @@ For execution on linux, replace instances of `host.docker.internal` with `localh
           --statediff.watchedaddresses <contract address>
     ```
 
-## getStorageAt
-Dockerfile for getting storage for all configured transformers at the given block, and persisting them into the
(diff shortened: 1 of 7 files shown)


dockerfiles/README.md
```diff
@@ -64,21 +64,21 @@ For execution on linux, replace instances of `host.docker.internal` with `localh
...
     ```
 
-## getStorageAt
-Dockerfile for getting storage for all configured transformers at the given block, and persisting them into the
-`public.storage_diff` table. This is useful in case it is suspected that a storage diff was missed. Please note that the
-storage value of every storage key for every transformer will be fetched and persisted, regardless if the value actually
-changed at the given block.
+## backfillStorage
+Dockerfile for getting storage for all configured transformers in a range of blocks, and persisting them into the
+`public.storage_diff` table. This is useful in case it storage diffs were missed. Please note that the storage value of
```
elizabethengelman, line 69: small typo: "This is useful in case ~it~ storage diffs were missed"

PR #154, opened 2020-04-08
VDB-1272: log item update

elizabethengelman:


environments/docker.toml
```diff
@@ -584,6 +584,14 @@
...
         deployed = 8925057
+    [contract.OASIS_ONE]
```
elizabethengelman, line 587: There are two Oasis contracts to track, not sure how to differentiate between the two besides "ONE" and "TWO" - totally up for some other ideas!

transformers/events/log_item_update/entity.go
```diff
@@ -0,0 +1,9 @@
+package log_item_update
```
elizabethengelman, line 1: Was _thinking_ of putting these Oasis transformers in their own `oasis` directory - curious what folks think about that.

environments/docker.toml
rmulhol: I think the numeric differentiator makes sense 👍 

My only question would be if we potentially want to be more specific - something like `OASIS_MATCHING_MARKET_ONE`. Mostly just not sure if there are other contracts in the Oasis ecosystem we'd want to add later

db/migrations/20200407102623_create_log_item_update.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+CREATE TABLE maker.log_item_update
+(
+    id         SERIAL PRIMARY KEY,
+    log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    offer_id   INT,
+    UNIQUE (header_id, log_id)
```
rmulhol, line 8: Wondering if we might want to add the contract's address_id to this table if it's populated from multiple contracts

transformers/events/log_item_update/entity.go
rmulhol: I like the idea! One thing I'd consider is if we might want to have the Oasis stuff run in a whole separate repo (`vdb-oasis-transformers`) and a different db schema. I think that'd be a cool way to start building off of the plugin architecture, but also may be too much complexity for the time being

transformers/events/log_item_update/transformer.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
+	"github.com/makerdao/vulcanizedb/pkg/eth"
+)
+
+const OfferId event.ColumnName = "offer_id"
```
rmulhol, line 15: Maybe this belongs in `transformers/shared/constants/keys.go`?

transformers/events/log_item_update/transformer.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		entity.HeaderID = log.HeaderID
+		entity.LogID = log.ID
+		entities = append(entities, entity)
+	}
+
+	return entities, nil
+}
+func (t Transformer) ToModels(abi string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
```
rmulhol, line 42: maybe worth replacing `db` with `_`

db/migrations/20200407102623_create_log_item_update.sql
elizabethengelman: oh yeah, good call!

transformers/events/log_item_update/transformer.go
elizabethengelman: with adding the address column, looks like we'll need the db in the transformer now. but good call!

transformers/events/log_item_update/transformer.go
elizabethengelman: That makes sense. I left it in the this transformer for now, since there are no other transformers with a field called `offer_id` but that's likely to change as we keep working through these Oasis events.

transformers/events/log_item_update/entity.go
elizabethengelman: 🤔 that's a really interesting idea. I wonder if its possible to start the transformers here, and then move them over to their own repo at some point. Or if that would be more trouble than it's worth.

PR #141, opened 2020-03-11
Vdb 1222 medianizer storage

yaoandrew:

PR #155, opened 2020-04-10
Medianizer storage

yaoandrew:
Ended up cherry picking a few commits to make this new branch instead of a crazy rebase on #141


rmulhol (changes requested):
Migration timestamp needs updating after #145 was merged

db/migrations/20200310230012_create_medianizer_storage_table.sql
```diff
@@ -0,0 +1,46 @@
+-- +goose Up
+-- SQL in this section is executed when the migration is applied.
+CREATE TABLE maker.medianizer_val
```
gslaughl: I think we probably want an `address_id` in these tables, since there are multiple `Median` contracts.

go.sum
```diff
@@ -1,330 +0,0 @@
-cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
gslaughl: Probably don't want to check this in ;)

gslaughl: Oh and also some component tests wouldn't hurt 😃

environments/docker.toml
```diff
@@ -13,6 +13,8 @@
         "usdc_flip",
         "wbtc_flip",
         "jug",
+        "median_bat",
+        "median_eth",
```
rmulhol, line 19: Is there potentially a median_wbtc contract we'd want to track?

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/ethereum/go-ethereum v1.9.11
 	github.com/howeyc/fsnotify v0.9.0 // indirect
 	github.com/jmoiron/sqlx v0.0.0-20181024163419-82935fac6c1a
-	github.com/lib/pq v1.0.0
+	github.com/lib/pq v1.3.0
```
rmulhol: I've become super paranoid about changes to `go.mod` 😅  - what's the rationale for bumping this? Do we potentially need a corresponding PR in vulcanizedb to use the same version?

transformers/component_tests/median/configured_transformer_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+			liftModel.ColumnValues[event.HeaderFK] = header.Id
+			liftModel.ColumnValues[event.LogFK] = liftLog.ID
+			liftModel.ColumnValues[event.AddressFK] = medianAddressID
+			liftModel.ColumnValues[constants.MsgSenderColumn] = msgSenderAddressID
+			liftModel.ColumnValues[constants.AColumn] = pq.Array([]string{aAddress})
+			insertErr := event.PersistModels([]event.InsertionModel{liftModel}, db)
+			Expect(insertErr).NotTo(HaveOccurred())
+
+			key := common.HexToHash("6666d42004e50ff3badadb0de5fc3899f37d3399b72f940c3f476b66e30ff7ba")
```
rmulhol: would be cool to use actual addresses/keys here so we could independently verify that the keys loader is working as expected

go.mod
yaoandrew: There's no real rationale. I think when Gabe first used this dependency to handle PG arrays, he didn't use the latest version which is 1.3.0. We can keep it at 1.0.0. If there's a risk to a version bump.

transformers/storage/median/repository.go
```diff
@@ -0,0 +1,286 @@
...
+	if budErr != nil {
+		return budErr
+	}
+
+	tx, txErr := repository.db.Beginx()
+	if txErr != nil {
+		return txErr
+	}
+
+	addressID, addressErr := shared.GetOrCreateAddress(contractAddress, repository.db)
```
rmulhol: nbd but it could be cool to extract a function that does the `GetOrCreateAddress` the first time it's called and returns the id without a DB query in subsequent calls - just so we don't need to go to the DB for the contract address ID on every insert

transformers/storage/median/repository.go
```diff
@@ -0,0 +1,286 @@
...
+		return addressErr
+	}
+
+	_, insertErr := tx.Exec(insertMedianBudQuery, diffID, headerID, addressID, budAddressID, bud)
+	if insertErr != nil {
+		rollbackErr := tx.Rollback()
+		if rollbackErr != nil {
+			return shared.FormatRollbackError("bud record with address", insertErr)
+		}
+		return insertErr
```
rmulhol: similarly nbd but could be cool to use `fmt.Errorf` throughout - so like here we could do:
```
return fmt.Errorf("error inserting median bud for address %s: %w", contractAddress, insertErr)
```
mostly just helps out with debugging so you don't get stack traces with errors like `error executing transformer: sql no rows` without intermediate context

transformers/storage/median/repository.go
```diff
@@ -0,0 +1,286 @@
...
+	case wards.Wards:
+		return wards.InsertWards(diffID, headerID, metadata, repository.ContractAddress, value.(string), repository.db)
+	case Bar:
+		return repository.insertMedianBar(diffID, headerID, repository.ContractAddress, value.(string))
+	case Bud:
+		return repository.insertBud(diffID, headerID, metadata, repository.ContractAddress, value.(string))
+	case Orcl:
+		return repository.insertOrcl(diffID, headerID, metadata, repository.ContractAddress, value.(string))
+	case Slot:
+		return repository.insertSlot(diffID, headerID, metadata, repository.ContractAddress, value.(string))
```
rmulhol: more nitpicking 😅 but why are we passing `repository.ContractAddress` in the args? Seems like these functions could just access that value directly

transformers/storage/median/repository.go
```diff
@@ -0,0 +1,286 @@
...
+func (repository *MedianStorageRepository) insertPackedValueRecord(diffID, headerID int64, contractAddress string, metadata types.ValueMetadata, packedValues map[int]string) error {
+	for order, value := range packedValues {
+		var insertErr error
+		switch metadata.PackedNames[order] {
+		case Val:
+			insertErr = repository.insertMedianVal(diffID, headerID, contractAddress, value)
+		case Age:
+			insertErr = repository.insertMedianAge(diffID, headerID, contractAddress, value)
+		default:
+			panic(fmt.Sprintf("unrecognized median contract storage name in packed values: %s", metadata.Name))
```
rmulhol: I know we've been using `panic`s in a few places like this, but maybe we could return the error here?

transformers/storage/median/repository.go
```diff
@@ -0,0 +1,286 @@
...
+		if rollbackErr != nil {
+			return shared.FormatRollbackError("slot record with address", insertErr)
+		}
+		return insertErr
+	}
+	return tx.Commit()
+}
+
+func getSlotID(keys map[types.Key]string) (int, error) {
+	slotID, err := strconv.Atoi(keys[constants.SlotId])
```
rmulhol: might be good to derive the string representation of the slot id with
```
slotIDstr, ok := keys[constants.SlotId]
```
so that we can also return an error if the key is missing

transformers/storage/median/repository_test.go
```diff
@@ -0,0 +1,272 @@
...
+			query := fmt.Sprintf(`SELECT diff_id, header_id, address_id, a AS key, orcl AS value FROM %s`, shared.GetFullTableName(constants.MakerSchema, constants.MedianOrclTable))
+			readErr := db.Get(&result, query)
+			Expect(readErr).NotTo(HaveOccurred())
+			contractAddressID, contractAddressErr := shared.GetOrCreateAddress(repo.ContractAddress, db)
+			Expect(contractAddressErr).NotTo(HaveOccurred())
+			orclAddressID, orclAddressErr := shared.GetOrCreateAddress(fakeOrclAddress, db)
+			Expect(orclAddressErr).NotTo(HaveOccurred())
+			Expect(result.AddressID).To(Equal(strconv.FormatInt(contractAddressID, 10)))
+			AssertMapping(result.MappingRes, diffID, fakeHeaderID, strconv.FormatInt(orclAddressID, 10), fakeUint256)
+		})
```
rmulhol, line 235: worth adding a `does not duplicate row`?

db/migrations/20200615231106_create_median_storage_tables.sql
```diff
@@ -0,0 +1,116 @@
...
+
+CREATE TABLE maker.median_orcl
+(
+    id         SERIAL PRIMARY KEY,
+    diff_id    BIGINT  NOT NULL REFERENCES storage_diff (id) ON DELETE CASCADE,
+    header_id  INTEGER NOT NULL REFERENCES headers (id) ON DELETE CASCADE,
+    address_id INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    a          INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    orcl        INTEGER NOT NULL,
+    UNIQUE (diff_id, header_id, address_id, a, orcl)
```
rmulhol: This would be different from what we did previously, _but_ I've been thinking maybe we should omit the `diff_id` from the uniqueness constraint on these tables. Unique on `header_id` + `address_id` + `a` + `orcl` would still prevent those duplicates, but _also_ catch cases where (for whatever reason) we're getting multiple diffs for the same state change

rmulhol (approved):
LGTM - all comments non-blocking 👍

transformers/storage/median/repository.go
yaoandrew: Whoops. Overlooked that one. 😓

go.mod
```diff
@@ -3,24 +3,16 @@ module github.com/makerdao/vdb-mcd-transformers
...
 	github.com/onsi/ginkgo v1.10.1
 	github.com/onsi/gomega v1.10.0
 	github.com/sirupsen/logrus v1.2.0
-	github.com/spaolacci/murmur3 v1.0.1-0.20190317074736-539464a789e9
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/viper v1.3.2
-	github.com/stretchr/testify v1.4.0
 	golang.org/x/crypto v0.0.0-20190404164418-38d8ce5564a5
-	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
-	gopkg.in/yaml.v2 v2.2.4
```
rmulhol, line 23: Have you had a chance to try building the `execute` docker image and verifying that you can run the container? Always paranoid when we change the deps 😅

transformers/storage/median/repository_test.go
```diff
@@ -0,0 +1,305 @@
...
+			Expect(err).To(MatchError(types.ErrMetadataMalformed{MissingData: constants.A}))
+		})
+	})
+
+	Describe("slot mapping", func() {
+		var fakeUint8 = strconv.Itoa(rand.Intn(255))
+		var fakeSlotAddress = "0x" + fakes.RandomString(40)
+
+		It("writes a row", func() {
+			//fakeSlotAddress := "0x" + fakes.RandomString(40)
```
rmulhol: ✂️

db/migrations/20200615231106_create_median_storage_tables.sql
```diff
@@ -0,0 +1,116 @@
...
+
+CREATE TABLE maker.median_orcl
+(
+    id         SERIAL PRIMARY KEY,
+    diff_id    BIGINT  NOT NULL REFERENCES storage_diff (id) ON DELETE CASCADE,
+    header_id  INTEGER NOT NULL REFERENCES headers (id) ON DELETE CASCADE,
+    address_id INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    a          INTEGER NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    orcl        INTEGER NOT NULL,
+    UNIQUE (header_id, address_id, a, orcl)
```
rmulhol, line 61: If we're omitting the `diff_id` here (👍), we should probably also do so for the other tables we're creating in this migration

PR #161, opened 2020-04-14
(VDB-1286) Add LogMake event transformer

rmulhol:


db/migrations/20200413164851_create_log_make.sql
```diff
@@ -0,0 +1,33 @@
+-- +goose Up
+CREATE TABLE maker.log_make
+(
+    id         serial primary key,
+    log_id     bigint  not null references public.event_logs (id) on delete cascade,
+    header_id  integer not null references public.headers (id) on delete cascade,
+    address_id integer not null references public.addresses (id) on delete cascade,
+    offer_id   numeric,
+    pair       character varying(66),
```
elizabethengelman, line 9: `pair` on LogKill is the same as it is here on LogMake where it's `pay_gem` and `buy_gem` packed together. Storing it as is as a varchar makes sense to me, but was just curious if you were knew why pair was stored like that, since we have `pay_gem` and `buy_gem` already on this event. 🤔

transformers/events/log_item_update/transformer.go
```diff
@@ -12,8 +12,6 @@ import (
 	"github.com/makerdao/vulcanizedb/pkg/eth"
 )
 
-const OfferId event.ColumnName = "offer_id"
-
```
elizabethengelman, line 16: 👍

transformers/test_data/log_make.go
```diff
@@ -0,0 +1,64 @@
...
+		Topics: []common.Hash{
+			common.HexToHash(constants.LogMakeSignature()),
+			common.HexToHash("0x00000000000000000000000000000000000000000000000000000000000c6280"),
+			common.HexToHash("0xcdd6659bca20e2b28ea10ead902280762ac8977c84459a152f90e561d50edf8c"),
+			common.HexToHash("0x0000000000000000000000006ff7d252627d35b8eb02607c8f27acdb18032718"),
+		},
+		Data:        hexutil.MustDecode("0x000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006b175474e89094c44da98b954eedeac495271d0f0000000000000000000000000000000000000000000000037d73b4e66fa60000000000000000000000000000000000000000000000000304f0cf4f5072f08000000000000000000000000000000000000000000000000000000000005e3e55fc"),
+		BlockNumber: 9440502,
+		TxHash:      common.HexToHash("0xb2f2f13a6ef0d1d6dd153d1a019929a31799498dcb1c100e04695fc0e95b9e58"),
+		TxIndex:     1772,
```
elizabethengelman, line 25: Silly question, but where did you find the `TxIndex` and the log `Index`? I think I used to find them on etherscan but didn't see them recently.

db/migrations/20200413164851_create_log_make.sql
rmulhol: My only thought is maybe a user would want to filter logs by the pair and hashing provides a consistent data structure that's easy to reproduce in various clients/ecosystems but I'm not sure 🤔

transformers/test_data/log_make.go
rmulhol: Yeah same found on Etherscan but think I'm doing it wrong 😅 

The transaction has a Nonce | Position field and I think we want to use Position for tx index (but I used Nonce, whoops!). 

And then the events on the transaction have a green circle to the left that I _think_ is the log index:
Screen Shot 2020-04-14 at 2 15 13 PM

PR #165, opened 2020-04-15
Remove yarn install from .travis.yml

rmulhol:
- Install was failing with "Bad header line" and it's no longer
  required for the build


paytonrules:
:shipit:  - and drinking a 🍺 in your honor.

PR #168, opened 2020-04-16
(VDB-1287) Add LogTrade event transformer

rmulhol:


transformers/integration_tests/vat_slip_test.go
```diff
@@ -87,3 +92,9 @@ type vatSlipModel struct {
 	Usr string
 	Wad string
 }
+
+type byWad []vatSlipModel
```
elizabethengelman, line 96: 👍

PR #169, opened 2020-04-16
(VDB-1279) LogSortedOffer + LogUnsortedOffer transformers

Gslaughl:
2 nearly-identical transformers in 2 separate commits

### DIFF
--- a/transformers/integration_tests/log_unsorted_offer_test.go
+++ b/transformers/integration_tests/log_unsorted_offer_test.go
@@ -1,0 +1,58 @@
+package integration_tests
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/makerdao/vdb-mcd-transformers/test_config"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/events/log_unsorted_offer"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/libraries/shared/fetcher"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("LogUnsortedOffer Transformer", func() {
+	logUnsortedOfferConfig := event.TransformerConfig{
+		TransformerName:   constants.LogUnsortedOfferTable,
+		ContractAddresses: test_data.OasisAddresses(),
+		ContractAbi:       constants.OasisABI(),
+		Topic:             constants.LogUnsortedOfferSignature(),
+	}
+
+	It("fetches and transforms a LogUnsortedOffer event for OASIS_MATCHING_MARKET_ONE contract", func() {
+		blockNumber := int64(9243052)
+		logUnsortedOfferConfig.StartingBlockNumber = blockNumber
+		logUnsortedOfferConfig.EndingBlockNumber = blockNumber
+
+		test_config.CleanTestDB(db)
+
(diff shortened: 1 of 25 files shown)


transformers/integration_tests/log_unsorted_offer_test.go
```diff
@@ -0,0 +1,65 @@
...
+		err = transformer.Execute(eventLogs)
+		Expect(err).NotTo(HaveOccurred())
+
+		var dbResult []logUnsortedOfferModel
+		err = db.Select(&dbResult, `SELECT offer_id FROM maker.log_unsorted_offer ORDER BY offer_id`)
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(len(dbResult)).To(Equal(1))
+		Expect(dbResult[0].OfferID).To(Equal("717050"))
+	})
```
rmulhol, line 58: Assuming we haven't seen any of these on the second contract?

transformers/integration_tests/log_unsorted_offer_test.go
```diff
@@ -0,0 +1,65 @@
...
+})
+
+type logUnsortedOfferModel struct {
+	OfferID  string `db:"offer_id"`
+	HeaderID int64
+	LogID    int64 `db:"log_id"`
```
rmulhol, line 64: nbd but I think we can remove the `HeaderID` and `LogID` from this model if we're only querying for the `OfferID`. For that matter, we may not even need a model if we can just scan that value directly into a string or int64

transformers/test_data/log_sorted_offer.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/fakes"
+)
+
+var (
+	logSortedOfferOfferID, _ = new(big.Int).SetString("811648", 10)
```
rmulhol, line 17: what's the motivation for putting this string into a big int? Seems like the only usage is calling `String()` on it?

transformers/test_data/log_sorted_offer.go
```diff
@@ -0,0 +1,51 @@
...
+	logSortedOfferModel = event.InsertionModel{
+		SchemaName: constants.MakerSchema,
+		TableName:  constants.LogSortedOfferTable,
+		OrderedColumns: []event.ColumnName{
+			event.HeaderFK, event.LogFK, event.AddressFK, constants.OfferId,
+		},
+		ColumnValues: event.ColumnValues{
+			event.HeaderFK:    LogSortedOfferEventLog.HeaderID,
+			event.LogFK:       LogSortedOfferEventLog.ID,
+			constants.OfferId: logSortedOfferOfferID.String(),
```
rmulhol, line 46: this won't work 99% of the time but in this case we might be able to do `new(big.Int).SetBytes(logSortedOfferRawLog.Data).String()`

rmulhol:
Paranoid me checking in to request a migration timestamp bump if #170 is merged first 😅

transformers/test_data/log_sorted_offer.go
Gslaughl: 🤦‍♂️

transformers/integration_tests/log_unsorted_offer_test.go
Gslaughl: Good call!

transformers/integration_tests/log_unsorted_offer_test.go
Gslaughl: correct 👍

PR #171, opened 2020-04-20
LogMinSell Oasis transformer

paytonrules:
https://makerdao.atlassian.net/secure/RapidBoard.jspa?rapidView=17&projectKey=VDB&modal=detail&selectedIssue=VDB-1276


transformers/events/log_min_sell/entity.go
```diff
@@ -0,0 +1,16 @@
+package log_min_sell
+
+import (
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type LogMinSellEntity struct {
+	Id              *big.Int
```
rmulhol, line 10: I think we can probably delete this field from the entity

transformers/integration_tests/log_min_sell_test.go
```diff
@@ -0,0 +1,131 @@
...
+var _ = Describe("LogMinSell Transformer", func() {
+	config := event.TransformerConfig{
+		TransformerName:   constants.LogMinSellTable,
+		ContractAddresses: test_data.OasisAddresses(),
+		ContractAbi:       constants.OasisABI(),
+		Topic:             constants.LogMinSellSignature(),
+	}
+
+	// OASIS_MATCHING_MARKET_ONE 0x39755357759ce0d7f32dc8dc45414cca409ae24e
+	// OASIS_MATCHING_MARKET_TWO 0x794e6e91555438afc3ccf1c5076a74f42133d08d
```
rmulhol, line 27: ✂️

transformers/test_data/log_min_sell.go
```diff
@@ -0,0 +1,53 @@
...
+	logMinSellModel = event.InsertionModel{
+		SchemaName: constants.MakerSchema,
+		TableName:  constants.LogMinSellTable,
+		OrderedColumns: []event.ColumnName{
+			event.HeaderFK, event.LogFK, event.AddressFK, constants.PayGemColumn,
+			constants.MinAmountColumn},
+		ColumnValues: event.ColumnValues{
+			event.HeaderFK: LogMinSellEventLog.HeaderID,
+			event.LogFK:    LogMinSellEventLog.ID,
+			// pay gem address id
```
rmulhol, line 47: if we're adding a comment for columns that aren't included here (I dig it), I'd prob put the `address id` on a separate line to make clear that it's a separate field (and not just the address id for the pay gem)

rmulhol (changes requested):
Nice work! LGTM as soon as the migration timestamp is updated 👍

transformers/events/log_min_sell/entity.go
paytonrules: Those aren't required? Kindda surprised.

transformers/integration_tests/log_min_sell_test.go
paytonrules: 🤦

transformers/events/log_min_sell/entity.go
rmulhol: Yeah the entity is just used for unpacking blockchain data into Go, so in theory it should only have the fields that are defined on the event (pay_gem, min_amount). By convention, we've been adding important metadata that needs to be inserted into postgres (header_id, log_id, contract_address), but don't do that for the database ID since those are created automatically (and occasionally the event will have its own "Id" that the entity needs to declare in order to unpack, and we end up inserting those as offer_id/bid_id/etc).

transformers/test_data/log_min_sell.go
paytonrules: "I learned it from watching you Dad" - I mean I was trying to follow along with what you did in LogMake.

```
			event.HeaderFK: LogMakeEventLog.HeaderID,
			event.LogFK:    LogMakeEventLog.ID,
			// Oasis contract address id
			constants.OfferId:    "811648",
			constants.PairColumn: "0xcdd6659bca20e2b28ea10ead902280762ac8977c84459a152f90e561d50edf8c",
			// Maker address id
			// Pay gem address id
			// Buy gem address id
			constants.PayAmtColumn:    "64380000000000000000",
			constants.BuyAmtColumn:    "14258238600000000000000",
			constants.TimestampColumn: "1581143548",
		},
```

rmulhol: `// Oasis contract address id` 👀

rmulhol: But also do as I say not as I do! 😅

rmulhol (approved):

PR #172, opened 2020-04-20
Remove log discards in test suites

rmulhol:
- No longer necessary since logging happens higher up
- Also rename suites to be more verbose


yaoandrew (approved):
Looking good -494 lines 🥇

paytonrules (approved):
I regret that I have but one shipit to give.

PR #173, opened 2020-04-21
Drop trigger that updates Urn created timestamp

rmulhol:
- Performing very slow for Urns that change a lot
- Tests marked as pending because we should figure out a way to
  get eventual consistency for this field, either by optimizing
  the query or replacing with a separate process

### DIFF
--- a/db/migrations/20200421142131_drop_urn_created_trigger.sql
+++ b/db/migrations/20200421142131_drop_urn_created_trigger.sql
@@ -1,0 +1,75 @@
+-- +goose Up
+DROP TRIGGER urn_ink ON maker.vat_urn_ink;
+
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION maker.update_urn_inks() RETURNS TRIGGER
+AS
+$$
+BEGIN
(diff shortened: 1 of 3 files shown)


db/migrations/20200421142131_drop_urn_created_trigger.sql
```diff
@@ -0,0 +1,75 @@
...
+        PERFORM maker.update_urn_inks_until_next_diff(OLD, urn_ink_before_block(OLD.urn_id, OLD.header_id));
+        PERFORM maker.delete_obsolete_urn_state(OLD.urn_id, OLD.header_id);
+    END IF;
+    RETURN NULL;
+END
+$$
+    LANGUAGE plpgsql;
+-- +goose StatementEnd
+
+DROP FUNCTION maker.update_urn_created(INTEGER);
```
elizabethengelman, line 22: Not sure it matters, but we also have a comment on this function to omit it from postgraphile - think it's work removing the comment too?

elizabethengelman (approved):
LGTM! Looking forward to seeing what this does!

rmulhol:
Closing for now in favor of #174

PR #177, opened 2020-04-24
Add log buy enabled event transformer

yaoandrew:


transformers/events/log_buy_enabled/transformer.go
```diff
@@ -0,0 +1,68 @@
...
+
+		entity.HeaderID = log.HeaderID
+		entity.LogID = log.ID
+		entity.ContractAddress = address
+		entities = append(entities, entity)
+	}
+
+	return entities, nil
+}
+func (t Transformer) ToModels(abi string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
```
rmulhol, line 42: nbd but maybe want a line break before this function

transformers/shared/constants/signature.go
```diff
@@ -32,9 +32,11 @@ func JugFileIlkSignature() string         { return getLogNoteTopicZero(jugFileIl
 func JugFileVowSignature() string         { return getLogNoteTopicZero(jugFileVowMethod()) }
 func JugInitSignature() string            { return getLogNoteTopicZero(jugInitMethod()) }
 func LogBumpSignature() string            { return getEventTopicZero(logBumpEvent()) }
+func LogBuyEnabledSignature() string      { return getEventTopicZero(logBuyEnabledEvent()) }
 func LogItemUpdateSignature() string      { return getEventTopicZero(logItemUpdateEvent()) }
 func LogKillSignature() string            { return getEventTopicZero(logKillEvent()) }
 func LogMakeSignature() string            { return getEventTopicZero(logMakeEvent()) }
+func LogMatchingEnabledSignature() string { return getEventTopicZero(logMatchingEnabledEvent()) }
```
rmulhol, line 39: What's up with the added `LogMatchingEnabled` stuff?

rmulhol (commented):
LGTM but I'd be reluctant to merge right now (and trigger a rescan of all blocks) if there are no matching logs on mainnet. Also curious about the `LogMatchingEnabled` stuff - do we need a full transformer for that? Would have no problem rolling that into this PR if so.

Not saying we should close the issue though - I think we should merge whenever it can go in alongside another PR that will find new events

transformers/shared/constants/signature.go
yaoandrew: I think I was anticipating doing those all in one PR so I just did the work at the same time with respect to signatures.

rmulhol (approved):

transformers/test_data/log_buy_enabled.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/fakes"
+)
+
+var (
+	//TODO Updqte with real data
```
rmulhol, line 16: This totally doesn't matter but if you need to amend a commit to trigger the build you could fix the typo here

PR #178, opened 2020-04-24
Set Min Sell Transformer

paytonrules:
https://makerdao.atlassian.net/secure/RapidBoard.jspa?rapidView=17&projectKey=VDB&modal=detail&selectedIssue=VDB-1281

### DIFF
--- a/transformers/events/set_min_sell/transformer.go
+++ b/transformers/events/set_min_sell/transformer.go
@@ -1,0 +1,55 @@
+package set_min_sell
+
+import (
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
+)
+
+type Transformer struct{}
+
+func (t Transformer) ToModels(_ string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
+	var models []event.InsertionModel
+
+	for _, log := range logs {
+		err := shared.VerifyLog(log.Log, shared.FourTopicsRequired, true)
+		if err != nil {
+			return nil, err
+		}
+
+		addressID, addressErr := shared.GetOrCreateAddress(log.Log.Address.Hex(), db)
+		if addressErr != nil {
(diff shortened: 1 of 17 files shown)


transformers/events/set_min_sell/transformer.go
```diff
@@ -0,0 +1,55 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		addressID, addressErr := shared.GetOrCreateAddress(log.Log.Address.Hex(), db)
+		if addressErr != nil {
+			return nil, shared.ErrCouldNotCreateFK(addressErr)
+		}
+
+		payGemAddress := log.Log.Topics[2].Hex()
```
rmulhol, line 33: If we have the `msg.Sender` in `Topics[1]`, maybe we should add that to the transformer + migration? We haven't been doing that consistently in the past 😅 but adding it in now will be way easier than doing so after the fact with https://makerdao.atlassian.net/browse/VDB-1073

Example available at https://github.com/makerdao/vdb-mcd-transformers/blob/staging/db/migrations/20200407131156_create_median_drop.sql#L9

rmulhol (approved):
LGTM - nice work!

I know this isn't in the requirements but can you pretty please add that `msg.sender` to the transformer/migration?

transformers/events/set_min_sell/transformer.go
paytonrules: I can add it, no prob. How can I tell that Topics[1] has the sender?

rmulhol: With this implementation of `LogNote`:
```
contract DSNote {
    event LogNote(
        bytes4   indexed  sig,
        address  indexed  guy,
        bytes32  indexed  foo,
        bytes32  indexed  bar,
        uint              wad,
        bytes             fax
    ) anonymous;

    modifier note {
        bytes32 foo;
        bytes32 bar;

        assembly {
            foo := calldataload(4)
            bar := calldataload(36)
        }

        LogNote(msg.sig, msg.sender, foo, bar, msg.value, msg.data);

        _;
    }
}
```
The four fields labeled "indexed" are the topics, in order. So `Topic[0]` is the signature (`msg.sig`), `Topics[1]` is the `msg.sender`, `Topics[2]` is the first arg passed to the function that has a `note` modifier (`foo`), and `Topics[3]` is the second arg passed to the function (`bar`).
The non-indexed fields (`wad`/`msg.value` and `fax`/`msg.data`) show up in the `Data` payload of the `types.Log`.

paytonrules: 👍

PR #180, opened 2020-04-27
Vat can storage

rmulhol:
This results in an *incomplete* set of `can` keys being available to the keys loader, enabling us to transform ~1/3 of the `can` diffs in the db. Tried to set things up so that we only need to modify the `MakerStorageRepository.GetVatCanKeys()` implementation if we find a way to identify the rest of the `msg.sender` values passed to `hope`/`nope`.

### DIFF
--- a/db/migrations/20200429173048_create_vat_can_table.sql
+++ b/db/migrations/20200429173048_create_vat_can_table.sql
@@ -1,0 +1,25 @@
+-- +goose Up
+CREATE TABLE maker.vat_can
+(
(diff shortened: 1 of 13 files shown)


paytonrules (approved):

paytonrules (approved):
Apparently I viewed all these and didn't have comments, based on github.

rmulhol:
closing because this is a half measure that doesn't capture all entries on the `can` mapping

PR #182, opened 2020-04-29
Update creating an event transformer docs

paytonrules:


transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
-
-1. Generate an example raw log event, by either:
-
-   - Pulling the log directly from the Kovan deployment.
-   - Deploying the contract to a local chain and emitting the event manually.
+### Ensure there is a Contract
+If the contract isn't already present in the environment you'll need to add it
+before you can begin transforming its events. To do so:
+
+1. Get the contract signature if you don't already have it.
```
rmulhol: the term we'd want to use here is contract address

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+1. Search for the contract on [etherscan](https://etherscan.io/). Bookmark this
+   page, it is your new best friend.
+1. In each environment (at this time testing, docker and mcdTransformers) add a
```
rmulhol: no oxford comma?!

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+      deployed = 8928180
+```
+* The address is the signature you searched for, as a string.
```
rmulhol: I recognize it sounds weird to say "The address is the address" but I think the most idiomatic description might be something like "The address field should contain the hex representation of the address you used to find the contract on Etherscan," or something like that

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+* The abi can be found on the `contracts` tab in Etherscan (scroll down).
+You'll need the entire thing as a string, so it's best to use single quotes
+like above.
+* The deployed field is the block number the contract was deployed on. That
+can be found on the Contract's page, by looking for the Contract Creator
+and clicking it's transaction hash. That will take you to the block it was deployed on.
```
rmulhol: Technically it takes you to the transaction where it was deployed, which includes a link to the block that contains that transaction 🤓

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+
+### Create the Transformer
+
+To create a custom transformer you will need to create a new `Transformer`
+struct with a `ToModels` method on it which converts a `core.EventLog` object
+(the raw, untransformed data) to an `event.InsertionModel` (the domain object).
+
+Note that for this step you probably do _not_ need to create the database
+migration. This is because the only thing you'll be saving are addresses, which
```
rmulhol: It's possible you could be saving other things that are stored in other tables and referenced via foreign key, like `maker.ilks` or `maker.urns` - but I think the larger point stands that you shouldn't need a migration to make the transformer, as the only thing a transformer inserts are fields referenced by FK

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+    1. Write a test for the event signature in 
+    [`signature_test.go`](../shared/constants/signature_test.go)
+        1. To find out the event's signature you can use this [Keccak-256
+       calculator](https://emn178.github.io/online-tools/keccak_256.html). For Example:
+            * Search Etherscan for the contract `0x39755357759cE0d7f32dC8dC45414CCa409AE24e`. 
+            * Open the 'Contract' tab and in the contract source code and search
+            for the `LogItemUpdate` event. It looks like this: `event LogItemUpdate(uint id);`. 
+            * You can copy that string and paste it into the Keccak-256
+            calculator (link above). Remove the parameter names and use
+            `uint256` for the `unit` so like this: `LogItemUpdate(uint256)`.
```
rmulhol: :%s/unit/uint

also idk if it might be helpful to mention "(since a `uint` in solidity defaults to `uint256`)" - maybe too much info? same also applies for `int`/`int256`

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+            and search for it on the events tab of the contract. Make sure you 
+            prefix it with `0x`. The results should include the event you are
+            looking for, provided one already exists for it.
+    1. Make the test pass by updating the function list in `signature.go` as
+       well as updating `method.go` as needed. Both of these files are in the
+       constants package.
+    1. Create test event in [`test_data`](../test_data)
+        * Use Data from a real event in Etherscan wherever possible.
+          Particularly when it comes to Topics and the Data entries. You'll
+          thank me later.
```
rmulhol: 😂

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+        * Update constants as needed. Look at older examples of test data for
+          inspiration.
+        * You do not need a database migration at this point.
+    1. Create a new directory in `events` named after your new transformer. This
+       will be the package your transformer is stored in.
+    1. Create three files - `transformer.go`, `transformer_test.go` and
+       `<package_name>_test.go` in the new directory. See the other transformers
```
rmulhol: you can create the `<package_name>_test.go` from the command line by running `ginkgo bootstrap` in the relevant directory

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestLogMinSell(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "<PackageName>  Suite")
+}
+```
+   1. Implement the toModels function in `transformer.go` with the appropriate
```
rmulhol: :%s/toModels/ToModels

Maybe helpful to include a link to [the interface](https://github.com/makerdao/vulcanizedb/blob/staging/libraries/shared/factories/event/converter.go#L25)?

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+   testing in `transformer_test.go` of course. This means converting the raw log 
+   into a go struct. 
+   1. For custom events you can convert each EventLog entry into an entity
+   using the function `contract.UnpackLog(&entity, "<EventLogName>", log.Log)`.
+   The entity is usually defined (by you) in a file called `entity.go` in the
```
rmulhol: "by you" is probably the right answer here although in theory `abigen` may be helpful/necessary if figuring out the type mappings is challenging

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+         repo](https://github.com/dapphub/ds-note).
+       * For an example implementation look at the `flip_sol` transformer.
+   1. If, while implementing the transformer and its corresponding unit tests
+      you find you need to migrate the database see below. 
+
+### Store the data in the database
+1. Use the `make new_migration` task to create a new migration.
+    * Each event log has its own table in the database.
+    * The specific log event tables are all created in the `maker` schema.
+1. The new migration can be run by running `make test` or `make migration NAME=vulcanize_test`.
```
rmulhol: maybe worth replacing `vulcanize_test` with `<database_name>`?

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+1. To verify that the migration will work create an integration for the shiny
+   new transformer in [`integration_tests`](../integration_tests).
+1. Integration tests can be run with `make integrationtest`.
```
rmulhol, line 115: I sorta shudder at the thought of describing it in detail because it's something we need to streamline, but maybe worth covering something about what goes into writing an integration test for an event transformer?

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+
+var EventTransformerInitializer event.TransformerInitializer = event.ConfiguredTransformer{
+  Config:      shared.GetEventTransformerConfig(constants.LogDeleteTable, constants.LogDeleteSignature()),
+  Transformer: log_delete.Transformer{},
+}.NewTransformer
+```
+
+Simply replace the constants and package names with your transformer.
+
+1. Finally add your package to the list of transformerExporters in `plugins/transformerExporter.go`. Again alphabetically.
```
rmulhol: maybe extraneous but you can generate the `transformerExporter.go` from the command line with `./vulcanizedb compose --config=/path/to/config.toml`

rmulhol (approved):
Nice! 🙏 🙏 🙏 

I left a bunch of comments but they may be extraneous - curious what you think

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+struct with a `ToModels` method on it which converts a `core.EventLog` object
+(the raw, untransformed data) to an `event.InsertionModel` (the domain object).
+
+Note that for this step you probably do _not_ need to create the database
+migration. This is because the only thing you'll be saving are addresses, which
+go into the already existing address table. If you do need to create the
+migration, you'll know, because the tests won't pass. The directions here assume
+you do not, yet.
+
+1. Search for the contract on etherscan using it's signature.
```
elizabethengelman, line 55: I think this `signature` should be `address` here as well.

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+    1. Make the test pass by updating the function list in `signature.go` as
+       well as updating `method.go` as needed. Both of these files are in the
+       constants package.
+    1. Create test event in [`test_data`](../test_data)
+        * Use Data from a real event in Etherscan wherever possible.
+          Particularly when it comes to Topics and the Data entries. You'll
+          thank me later.
+        * Update constants as needed. Look at older examples of test data for
+          inspiration.
+        * You do not need a database migration at this point.
```
elizabethengelman: I wonder if this bullet point about a migration is necessary yet, since it's already mentioned above that a migration isn't likely necessary on this step.

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+   into a go struct. 
+   1. For custom events you can convert each EventLog entry into an entity
+   using the function `contract.UnpackLog(&entity, "<EventLogName>", log.Log)`.
+   The entity is usually defined (by you) in a file called `entity.go` in the
+   same package as the transformer. See the transformer in
+   `transformers/events/log_make` for an example of this method.
+   1. After converting the log entry to an entity, look up all of it's foreign
+      keys from it's data and assign them to the entity.
+   1. From the entity create an InsertionModel, which you return.
+   1. For `LogNote` events you'll need to look at the method signature of the
```
elizabethengelman: I wonder if it would be helpful to have directions for LogNote and directions for custom events as nested bullets to signify that the you need to do one or the other, not both?

Something like:
```
 1. Implement the ToModels function in `transformer.go`...
    - For custom events:
        - ...
    - For LogNoteEvents:
        - ...
```

transformers/events/DOCUMENTATION.md
```diff
@@ -2,116 +2,206 @@
...
+   method that is calling them, because LogNote events are a generic structure.
+   For example:
+       * the `tend` method is called on the [flip.sol
+       contract](https://github.com/makerdao/dss/blob/master/src/flip.sol#L123), 
+       and its method signature looks like this: `tend(uint,uint,uint)`.
+       * The first four bytes of the Keccak-256 hashed method signature will be
```
elizabethengelman: A small caveat to this is that for custom events the topic[0] will be the full Keccak-256 hashed method signature instead of just the first four bytes - that is special to LogNotes.

elizabethengelman (approved):
This is looking great! It's so nice to have accurate documentation again! 

Just a few small comments, but nothing merge blocking.

transformers/events/DOCUMENTATION.md
yaoandrew: This kind of burns me every time I'm trying to hash the function params. I think its worth it to add explicit detail at the risk of sounding like a personal request 😺

transformers/events/DOCUMENTATION.md
paytonrules: 👍

transformers/events/DOCUMENTATION.md
paytonrules: Don't tell James.

transformers/events/DOCUMENTATION.md
paytonrules: How about "The address field is the contract address you searched for, as a string."

transformers/events/DOCUMENTATION.md
paytonrules: 👍

transformers/events/DOCUMENTATION.md
paytonrules: I added some clarifying language here.

transformers/events/DOCUMENTATION.md
paytonrules: Anything that burns you repeatedly is a good thing to get written down.

transformers/events/DOCUMENTATION.md
paytonrules: 👍

transformers/events/DOCUMENTATION.md
paytonrules: Didn't know that!

transformers/events/DOCUMENTATION.md
paytonrules: Probably - I think I was running into some documentation fatigue at this point.

transformers/events/DOCUMENTATION.md
paytonrules: 👍

transformers/events/DOCUMENTATION.md
paytonrules: 👍

transformers/events/DOCUMENTATION.md
paytonrules: 👍

transformers/events/DOCUMENTATION.md
paytonrules: I added a bit of clarification here.

transformers/events/DOCUMENTATION.md
paytonrules: 👍

transformers/events/DOCUMENTATION.md
paytonrules: I added a note about `abigen`

rmulhol (changes requested):
Just dropping in to make sure we don't accidentally merge #188 as part of this PR

rmulhol (approved):
:shipit:

PR #183, opened 2020-04-30
Vdb 1292 log median price

yaoandrew:


db/migrations/20200429072513_create_log_median_price_table.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+CREATE TABLE maker.log_median_price
+    (
+        id serial primary key,
+        header_id integer NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+        log_id    BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+        val numeric,
+        age numeric,
+        UNIQUE (header_id, log_id)
```
rmulhol, line 10: Assuming this will be pointed at multiple contracts, probably want to put an `address_id` on here to know which ilk's median price we're dealing with.

Also - looks like we're missing config updates to include the transformer and specify what contracts it's watching

rmulhol (commented):
Looking good! I think we can merge when we get the config files + transformerExporter.go updated - and would ideally like an `address_id` on the table or some other way of distinguishing the ilk

db/migrations/20200429072513_create_log_median_price_table.sql
yaoandrew: Weird... I force pushed a commit with these yesterday that doesn't seem to have gone through. Let me figure that out and get an `address_id` on there. That makes sense to figure out which `ilk` it is.

rmulhol (approved):
:shipit:

PR #184, opened 2020-04-30
Add DIFF_BLOCK_FROM_HEAD_OF_CHAIN to execute docker startup script

elizabethengelman:
companion PR to https://github.com/makerdao/vulcanizedb/pull/84

### DIFF
--- a/dockerfiles/execute/startup_script.sh
+++ b/dockerfiles/execute/startup_script.sh
@@ -41,8 +41,6 @@ if test -n "$DIFF_BLOCK_FROM_HEAD_OF_CHAIN"; then
     DIFF_STARTING_BLOCK=$DIFF_BLOCK_FROM_HEAD_OF_CHAIN
 fi
 
-echo "diff starting block"
-echo $DIFF_STARTING_BLOCK
 echo "Starting transformer execution..."
 # Fire up execute
 ./vulcanizedb execute --config config.toml -d $DIFF_STARTING_BLOCK
--- a/dockerfiles/execute/startup_script.sh
+++ b/dockerfiles/execute/startup_script.sh
@@ -35,6 +35,14 @@ if [ $? -ne 0 ]; then
   exit 1
 fi
 
+DIFF_STARTING_BLOCK="-1"
+
+if test -n "$DIFF_BLOCK_FROM_HEAD_OF_CHAIN"; then
+    DIFF_STARTING_BLOCK=$DIFF_BLOCK_FROM_HEAD_OF_CHAIN
+fi
+
+echo "diff starting block"
+echo $DIFF_STARTING_BLOCK
 echo "Starting transformer execution..."
 # Fire up execute
-./vulcanizedb execute --config config.toml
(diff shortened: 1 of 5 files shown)


dockerfiles/execute/startup_script.sh
```diff
@@ -35,6 +35,14 @@ if [ $? -ne 0 ]; then
   exit 1
 fi
 
+DIFF_STARTING_BLOCK="-1"
```
elizabethengelman, line 38: Not sure if this is the best approach for handling this, so would be totally up for changing it up if folks have ideas!

go.mod
```diff
@@ -21,3 +21,5 @@ require (
 replace gopkg.in/urfave/cli.v1 => gopkg.in/urfave/cli.v1 v1.20.0
 
 replace github.com/ethereum/go-ethereum => github.com/vulcanize/go-ethereum v0.0.0-20190731183759-8e20673bd101
+
+replace github.com/makerdao/vulcanizedb => github.com/makerdao/vulcanizedb v0.0.14-rc.1.0.20200430184332-ecfa2869a661
```
elizabethengelman: Once https://github.com/makerdao/vulcanizedb/pull/84 is merged, I was thinking of cutting a new VDB version and using that on line 11 instead of replacing the version.

dockerfiles/execute/startup_script.sh
rmulhol: This approach makes sense to me 👍 

Seems like someone running the binary directly could just omit that flag, but if we're always including the flag then I like having a default. My only thought would be wondering if we could use an env var instead of a flag (so that it was fully optional and didn't need to be in the dockerfiles at all), but I think that makes it harder to set a default - and potentially means we'd need to synchronize deploying this with a modification to our configured env vars

go.mod
rmulhol: Not opposed to cutting a release and updating that way, but I think we could also solve this matter in the short term by running `go get github.com/makerdao/vulcanizedb@modify-unchecked-diffs-query`. Maybe nice to just modify the import directly instead of using a replace?

transformers/test_data/test_helpers.go
```diff
@@ -98,7 +100,9 @@ func CreateLogs(headerID int64, logs []types.Log, db *postgres.DB) []core.EventL
 	eventLogRepository := repositories.NewEventLogRepository(db)
 	insertLogsErr := eventLogRepository.CreateEventLogs(headerID, logs)
 	Expect(insertLogsErr).NotTo(HaveOccurred())
-	eventLogs, getLogsErr := eventLogRepository.GetUntransformedEventLogs()
+
+	logCount := getLogCount(db)
+	eventLogs, getLogsErr := eventLogRepository.GetUntransformedEventLogs(0, logCount)
```
rmulhol, line 105: nice catch!

rmulhol (approved):
LGTM!

Would be a fan of updating the dep via `go get` rather than a `replace` unless that runs into problems

PR #188, opened 2020-05-05
Remove urn_state type

paytonrules:
I did this as three migrations. The first to replace the all_urns function with a much simpler one. The second to remove all references to urn_state, and the third to rename obsolete_urn_state to obsolete_urn_snapshot. The code changes (non SQL edition) are almost all around tests.


db/migrations/20200504135835_remove_urn_state.sql
```diff
@@ -0,0 +1,226 @@
+-- +goose Up
+-- SQL in this section is executed when the migration is applied.
+DROP FUNCTION IF EXISTS api.all_urn_states;
+DROP FUNCTION IF EXISTS api.get_urn;
+
+DROP FUNCTION IF EXISTS api.bite_event_urn;
+CREATE OR REPLACE FUNCTION api.bite_event_urn(event api.bite_event) RETURNS api.urn_snapshot
```
rmulhol: Do we need the `DROP FUNCTION` if we're doing `CREATE OR REPLACE`?

transformers/component_tests/queries/get_urn_query_test.go
```diff
@@ -1,157 +0,0 @@
...
-	var (
-		vatRepo                vat.VatStorageRepository
-		headerRepo             repositories.HeaderRepository
-		urnOne, urnTwo         string
-		blockOne, timestampOne int
-		headerOne              core.Header
-		diffID                 int64
-	)
-
-	const getUrnQuery = `SELECT urn_identifier, ilk_identifier, ink, art, created, updated FROM api.get_urn($1, $2, $3)`
```
rmulhol: I like the idea of using the `api.urn_snapshots` table to replace this query, but do you know how we'd structure the query against that table to guarantee we're getting the most recent one by block number in GraphQL?

rmulhol: Worst come to worst I could see a case for keeping the query and it just wraps the `order by block_number desc limit 1`

transformers/component_tests/queries/urn_state_computed_columns_test.go
```diff
@@ -1,283 +0,0 @@
...
-	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
-	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
-	"github.com/makerdao/vulcanizedb/pkg/core"
-	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres/repositories"
-	"github.com/makerdao/vulcanizedb/pkg/fakes"
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-)
-
-var _ = Describe("Urn state computed columns", func() {
```
rmulhol, line 39: Just confirming - these go away because we have the exact same computed columns on the `urn_snapshot` table?

transformers/component_tests/queries/all_urn_states_query_test.go
```diff
@@ -1,218 +0,0 @@
...
-	"github.com/makerdao/vdb-mcd-transformers/transformers/storage/vat"
-	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
-	"github.com/makerdao/vulcanizedb/libraries/shared/storage/types"
-	"github.com/makerdao/vulcanizedb/pkg/core"
-	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres/repositories"
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-)
-
-var _ = Describe("Urn history query", func() {
```
rmulhol, line 21: And here it goes 💥 because we can just `select * from api.urn_snapshot where urn_identifier = x and ilk_identifier = y`?

db/migrations/20200504135835_remove_urn_state.sql
```diff
@@ -0,0 +1,226 @@
...
+$$;
+
+DROP FUNCTION IF EXISTS api.flip_bid_snapshot_urn;
+CREATE FUNCTION api.flip_bid_snapshot_urn(flip api.flip_bid_snapshot) RETURNS api.urn_snapshot
+       LANGUAGE sql STABLE
+       AS $$
+SELECT * FROM api.urn_snapshot
+WHERE ilk_identifier = (SELECT identifier FROM maker.ilks WHERE ilks.id = flip.ilk_id)
+AND urn_identifier = (SELECT identifier FROM maker.urns WHERE urns.id = flip.urn_id)
+AND block_height <= flip.block_height
```
rmulhol: I think this may need to `order by block_height desc limit 1` now that we're querying from a table that could have multiple rows with matching `urn_identifier`/`ilk_identifier`. Probably applies to queries below as well

rmulhol (commented):
Looking really good! Planning to take a closer look later, but dropping some initial thoughts

db/migrations/20200504131504_update_all_urns_function.sql
```diff
@@ -0,0 +1,80 @@
+-- +goose Up
+-- SQL in this section is executed when the migration is applied.
+
+DROP FUNCTION api.all_urns;
+
+CREATE OR REPLACE FUNCTION api.all_urns(block_height bigint DEFAULT api.max_block(), max_results integer DEFAULT NULL::integer, result_offset integer DEFAULT 0) RETURNS SETOF api.urn_snapshot
+    LANGUAGE sql STABLE
+    AS $$
+SELECT urn_identifier, ilk_identifier, all_urns.block_height, ink, coalesce(art, 0), created, updated
```
rmulhol: I think we need this to be distinct on the urn_identifier + ilk_identifier so that we're only getting the most recent state of every urn in the system

db/migrations/20200504135835_remove_urn_state.sql
paytonrules: Unfortunately since we change the signature of the function, we have to drop it first. You can't change the return value with a replace.

We probably don't need `CREATE OR REPLACE` - I was being overly-cautious..

transformers/component_tests/queries/get_urn_query_test.go
paytonrules: I guess it's a question if people are using it - we could certainly use the query you just said to keep it around.

transformers/component_tests/queries/urn_state_computed_columns_test.go
paytonrules: Yup 👍

transformers/component_tests/queries/all_urn_states_query_test.go
paytonrules: Exactly.

db/migrations/20200504131504_update_all_urns_function.sql
paytonrules: ~Does his mean you can have the same urn_identifier with a different ilk, and that should be a different urn?~

Snapshots, right, duh.

transformers/component_tests/queries/get_urn_query_test.go
paytonrules: I restored get_urn - assuming all_urns only returns the latest unique urns (ergo distinct on urn id and ilk) then get_urn should always return only 1 urn. 

I'd like to have the equivalent of ASSERT somewhere here - since if it returns more than one then something is wrong in the all_urns query. In either case as long as get_urn uses all_urns then we shouldn't need to `order by block_number desc limit`.

paytonrules: (Assuming I got it right of course).

db/migrations/20200504135835_remove_urn_state.sql
paytonrules: I restored get)urn - which should take care of that.

db/migrations/20200504131504_update_all_urns_function.sql
paytonrules: 👍

transformers/component_tests/queries/all_urns_query_test.go
```diff
@@ -115,6 +115,39 @@ var _ = Describe("Urn view", func() {
 		helper.AssertUrn(result[1], expectedUrnTwo)
 	})
 
+	It("returns only the most recent urn for multiple snapshots of the same urn", func() {
```
elizabethengelman, line 119: 👏

transformers/component_tests/queries/test_helpers/test_helpers.go
```diff
@@ -286,13 +286,13 @@ type UrnState struct {
...
 
 func AssertUrn(actual, expected UrnState) {
-	Expect(actual.UrnIdentifier).To(Equal(expected.UrnIdentifier))
-	Expect(actual.IlkIdentifier).To(Equal(expected.IlkIdentifier))
-	Expect(actual.BlockHeight).To(Equal(expected.BlockHeight))
-	Expect(actual.Ink).To(Equal(expected.Ink))
-	Expect(actual.Art).To(Equal(expected.Art))
-	Expect(actual.Created).To(Equal(expected.Created))
-	Expect(actual.Updated).To(Equal(expected.Updated))
+	Expect(actual.UrnIdentifier).To(Equal(expected.UrnIdentifier), "Urn Identifier")
```
elizabethengelman, line 289: 🎉

elizabethengelman (approved):
LGTM, nice work!

I'm a little hesitant to merge this in while we're still working on urn_snapshot reconciliation, but I'd be curious to hear what others think!

db/migrations/20200512130930_update_all_urns_function.sql
```diff
@@ -0,0 +1,83 @@
...
+
+CREATE FUNCTION api.all_urns(block_height bigint DEFAULT api.max_block(), max_results integer DEFAULT NULL::integer, result_offset integer DEFAULT 0) RETURNS SETOF api.urn_snapshot
+    LANGUAGE sql STABLE
+    AS $$
+    SELECT *
+    FROM ( SELECT DISTINCT ON (urn_identifier, ilk_identifier) urn_identifier, ilk_identifier,
+                              all_urns.block_height, ink, coalesce(art, 0), created, updated
+           FROM api.urn_snapshot
+           WHERE block_height <= all_urns.block_height
+           ORDER BY urn_identifier, ilk_identifier, updated DESC) AS latest_urns
```
rmulhol, line 14: Can we just `ORDER BY updated` here (omitting `urn_identifier, ilk_identifier`)? Also, do we need `AS latest_urns`?

db/migrations/20200512141017_remove_urn_state.sql
```diff
@@ -0,0 +1,232 @@
...
+-- SQL in this section is executed when the migration is applied.
+DROP FUNCTION IF EXISTS api.all_urn_states;
+
+DROP FUNCTION IF EXISTS api.get_urn;
+CREATE FUNCTION api.get_urn(ilk_identifier text, urn_identifier text, block_height bigint DEFAULT api.max_block()) RETURNS api.urn_snapshot
+    LANGUAGE sql STABLE STRICT
+    AS $_$
+
+SELECT *
+    FROM api.all_urns(get_urn.block_height)
```
rmulhol: Seems kinda heavyweight to query `all_urns` for one urn - might it be better to:
```
SELECT * FROM api.urn_snapshot WHERE ilk_identifier = x AND urn_identifier = y AND block_height <= n ORDER BY block_height DESC LIMIT 1
```

db/migrations/20200512141017_remove_urn_state.sql
```diff
@@ -0,0 +1,232 @@
...
+FROM api.get_urn(event.ilk_identifier, event.urn_identifier, event.block_height)
+$$;
+
+DROP FUNCTION IF EXISTS api.flip_bid_snapshot_urn;
+CREATE FUNCTION api.flip_bid_snapshot_urn(flip api.flip_bid_snapshot) RETURNS api.urn_snapshot
+       LANGUAGE sql STABLE
+       AS $$
+SELECT *
+FROM api.get_urn(
+     (SELECT identifier FROM maker.ilks WHERE ilks.id = flip.ilk_id),
```
rmulhol, line 33: Hm, this is _way_ out of scope for this PR, but it seems like `api.get_flip` should be returning a `maker.flip` and we should be deleting `api.flip_bid_snapshot` - basically a story to do exactly what you're doing here for a different place where we've got a type used in queries + a trigger table

rmulhol: Mostly just caught my eye because we have an `ilk_id` and `urn_id` on that type, instead of the identifiers - which seems odd. If we're going to have a custom type for queries, I would have thought we would have done the join to get the identifiers in the query and not exposed internal DB FKs

rmulhol (approved):
Looks good to me - a couple minor comments, would be curious to hear your thoughts, nothing merge blocking

db/migrations/20200512130930_update_all_urns_function.sql
paytonrules: Unfortunately we can't omit `urn_identifier` and `ilk_identifier` from ORDER BY because they are in the DISTINCT above. I'm not sure about latest_urns, I have to double-check.

db/migrations/20200512141017_remove_urn_state.sql
paytonrules: I was trying to be consistent with the other ones that use all_urns, but you're probably right.

db/migrations/20200512141017_remove_urn_state.sql
paytonrules: I'm not quite sure I followed that last bit - is there an action item here besides a story about flip_bid_snapshot?

rmulhol: No, sorry - kinda just thinking out loud. Will create stories for this stuff

rmulhol (approved):

PR #189, opened 2020-05-07
(VDB-1315) Add command for backfilling Urn diffs

rmulhol:
- insert diffs from every block with a frob for involved urn


backfill/frob/backfiller.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (backFiller frobBackFiller) BackFill(startingBlock int) error {
+	frobs, frobsErr := backFiller.eventsRepository.GetFrobs(startingBlock)
+	if frobsErr != nil {
+		return fmt.Errorf("error getting frobs: %w", frobsErr)
+	}
+
+	logrus.Infof("getting storage for %d grabs", len(frobs))
+	for i, frob := range frobs {
+		err := shared.FetchAndPersistDartDinkDiffs(frob.Dart, frob.Dink, frob.UrnID, frob.HeaderID, backFiller.eventsRepository, backFiller.storageRepository, backFiller.blockChain)
```
paytonrules: Maybe consider a struct for those first four params? It would prevent accidentally switching Dink/Dart and UrnID/HeaderID.

backfill/grab/backfiller_test.go
```diff
@@ -0,0 +1,241 @@
...
+		It("returns error if getting grab events fails", func() {
+			mockEventsRepository.GetGrabsError = fakes.FakeError
+
+			err := backFiller.BackFill(0)
+
+			Expect(err).To(HaveOccurred())
+			Expect(errors.Is(err, fakes.FakeError)).To(BeTrue())
+		})
+
+		It("ignores grab if dink and dart are zero", func() {
```
paytonrules: Some of these tests are duplicated and could be moved to the Shared lib function that does most of the work.

backfill/frob/backfiller.go
```diff
@@ -0,0 +1,41 @@
...
+		if err != nil {
+			return fmt.Errorf("error fetching and persisting diffs for frob %d: %w", i, err)
+		}
+	}
+	return nil
+}
```
paytonrules, line 47: If you do another backfiller beyond these two, you may consider making each backfiller a strategy. Essentially add methods for things like "error message" and "log storage" and have one place that does the Get"Things" then loop through them logic.  As opposed to copy/pasta.

backfill/repository/events_repository.go
```diff
@@ -0,0 +1,59 @@
...
+		ORDER BY headers.block_number ASC`, startingBlock)
+	return frobs, err
+}
+
+func (e eventsRepository) GetGrabs(startingBlock int) ([]Grab, error) {
+	var grabs []Grab
+	err := e.db.Select(&grabs, `SELECT header_id, urn_id, dink, dart
+		FROM maker.vat_grab
+		JOIN public.headers ON vat_grab.header_id = headers.id
+		WHERE headers.block_number >= $1`, startingBlock)
```
paytonrules: How come you don't order the Grabs but do the Frobs? I'm not sure it matters but it's odd that they are slightly different.

backfill/repository/events_repository_test.go
```diff
@@ -0,0 +1,192 @@
...
+			Expect(frobs).To(ConsistOf(frobAtStartingBlock))
+		})
+
+		It("orders results ascending by block_number", func() {
+			frobs, err := repo.GetFrobs(earlierBlock)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(frobs)).To(Equal(2))
+			Expect(frobs[0]).To(Equal(frobAtEarlierBlock))
+		})
```
paytonrules, line 108: What about the error case? Does that require shutting down the db or something?

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/onsi/gomega v1.7.0
 	github.com/pkg/errors v0.8.1 // indirect
 	github.com/sirupsen/logrus v1.2.0
+	github.com/spf13/cobra v0.0.3
```
paytonrules, line 18: image

paytonrules (approved):
This has been running in production for a while - so it's clearly working. However one thing I'd like to see would be some additions to the README explaining how to run these, and a hint on how to get started. 

Nothing huge, just something like "we use Cobra for back-fills and command line scripts that run against VDB. Here's the command to run it. "

backfill/frob/backfiller.go
rmulhol: That's a great idea - I made this mistake earlier and manually rechecked these params approx ~1 million times 😂

backfill/repository/events_repository.go
rmulhol: Probably makes sense to order both 👍 

The bigger the system gets and the more events we accumulate, I'm also starting to worry about unbounded results sets - i.e. how can we grab manageable chunks of events to backfill and proceed through from there

README.md
```diff
@@ -52,6 +54,13 @@ localhost. To allow access on Ubuntu, set localhost connections via hostname, ip
 
 (It should be noted that trusted auth should only be enabled on systems without sensitive data in them: development and local test databases)
 
+### Backfill
+This project contains one executable, `backfillUrns`.
+This command enables you to get storage data for Urns by performing lookups at blocks where we detected events indicating an Urn state change.
+To run it, you first need to `go build`.
+After that, you can run something like `./vdb-mcd-transformers backfillUrns --starting-block 10000000` - which would backfill Urn storage since block 10,000,000.
```
elizabethengelman: Super small, but I wonder if it would be worth mentioning that this commands backfills until the head of the chain?

backfill/repository/events_repository.go
```diff
@@ -0,0 +1,60 @@
...
+	var grabs []Grab
+	err := e.db.Select(&grabs, `SELECT header_id, urn_id, dink, dart
+		FROM maker.vat_grab
+		JOIN public.headers ON vat_grab.header_id = headers.id
+		WHERE headers.block_number >= $1
+		ORDER BY headers.block_number ASC`, startingBlock)
+	return grabs, err
+}
+
+func (e eventsRepository) GetHeaderByID(id int) (core.Header, error) {
```
elizabethengelman, line 56: Wondering if this method would make sense in the HeaderRepository? 🤔

backfill/repository/storage_repository.go
```diff
@@ -0,0 +1,65 @@
...
+	var urn Urn
+	err := repo.db.Get(&urn, `
+		SELECT DISTINCT urns.id AS urn_id, ilks.ilk, ilks.id AS ilk_id, urns.identifier AS urn
+		FROM maker.urns
+		    JOIN maker.ilks on maker.ilks.id = maker.urns.ilk_id
+		    WHERE urns.id = $1`, id)
+	return urn, err
+}
+
+func (repo storageRepository) InsertDiff(diff types.RawDiff) error {
```
elizabethengelman, line 41: Was trying to think of a way to reuse the `CreateStorageDiff` code here, but it looks like we'd need to rework that method to allow a `from_backfill` value to be passed in so that we could make sure these are set to `true`. So maybe it's best to keep this as is, since it's specific to backfilling.

backfill/repository/storage_repository_test.go
```diff
@@ -0,0 +1,267 @@
...
+			var ilkID int
+			ilkErr := db.Get(&ilkID, `INSERT INTO maker.ilks (ilk, identifier) VALUES ($1, $2) RETURNING id`, fakeIlk, fakeIlkIdentifier)
+			Expect(ilkErr).NotTo(HaveOccurred())
+
+			fakeUrn := test_data.RandomString(40)
+			var urnID int
+			urnErr := db.Get(&urnID, `INSERT INTO maker.urns (ilk_id, identifier) VALUES ($1, $2) RETURNING id`, ilkID, fakeUrn)
+			Expect(urnErr).NotTo(HaveOccurred())
+
+			urns, err := repo.GetUrnByID(urnID)
```
elizabethengelman: super small, but this could probably be `urn` instead of `urns`

backfill/shared/dart_dink.go
```diff
@@ -0,0 +1,148 @@
...
+	HeaderID int
+	UrnID    int
+}
+
+func FetchAndPersistDartDinkDiffs(dartDink DartDink, eventsRepository repository.EventsRepository, storageRepository repository.StorageRepository, blockChain core.BlockChain) error {
+	dartIsZero, dinkIsZero, err := dartAndDinkAreZero(dartDink.Dart, dartDink.Dink)
+	if err != nil {
+		return fmt.Errorf("error checking if dart/dink are zero: %w", err)
+	}
+	if dartIsZero && dinkIsZero {
```
elizabethengelman, line 34: I wonder if it's worth it to add a comment here about why we're returning when `dart` and `dink` are both zero? 

I've been thinking about zero-valued diffs for too long and did a double take because I was thinking these were diffs, and not frob/grab values, so its also super possible this isn't necessary 😜

backfill/shared/keys.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/makerdao/vdb-mcd-transformers/backfill/repository"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/storage/utilities"
+	"github.com/makerdao/vulcanizedb/libraries/shared/storage"
+)
+
+func GetIlkArtKey(ilk string) common.Hash {
+	return storage.GetKeyForMapping(storage.IndexTwo, ilk)
+}
+
+func GetUrnArtKey(u repository.Urn) (common.Hash, error) {
```
elizabethengelman, line 12: I wonder if it's worth making [getUrnArtKey](https://github.com/makerdao/vdb-mcd-transformers/blob/staging/transformers/storage/vat/keys_loader.go#L268) and [getUrnInkKey](https://github.com/makerdao/vdb-mcd-transformers/blob/61e6bf6da0b9b25715632fae10afddda9b59f489/transformers/storage/vat/keys_loader.go#L259) public so that we can reuse that code. If the way we get either of these keys changes, it would be nice to only have to change it in one place. Though I'm not sure how likely it is that that will ever change. 🤔

elizabethengelman (approved):
LGTM! :shipit: 

Just a few questions/comments - nothing merge block though!

backfill/repository/events_repository.go
rmulhol: Yeah definitely - can open up a PR over there to add it 👍

backfill/repository/storage_repository.go
rmulhol: Will take a look - maybe we could add a new function in the storage repository, at least, for creating a back filled diff without using the DB function that protects against dupes

PR #192, opened 2020-05-18
Vdb 1294 update docker compose config and directions

paytonrules:
Extract run-migrations from startup-script, and then reference the docker-compose setup in the directions first.

This PR has a bit of a chicken-and-egg problem as the Docker image needs to be updated before the directions are correct. I validated this using a local image, but that's not a perfect setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,14 +14,38 @@
 This repository is a collection of transformers to be used along with VDB as a plugin to fetch, transform and persist log events and storage slots of specified MCD contracts.
 
 ## Install
-### Dependencies
- - [VulcanizeDB](https://github.com/makerdao/vulcanizedb)
- - Go 1.12+
- - Postgres 11.2
+
+### Docker Setup
+
+This repository provides a docker-compose setup to manage the multiple container-based deployment for you, if at all possible you should use that setup.
+
+#### Dependencies
+ - Docker (the rest is handled by docker-compose)
+
+#### Database Initialization
+
+The database needs to be initialized once before bringing up the system. That can be by running, from the root directory, 
+
+```bash
+docker-compose -f dockerfiles/docker-compose.yml run execute ./run_migrations.sh
+```
+
+#### System Startup
+
+```bash
+docker-compose -f dockerfiles/docker-compose.yml up
+```
+
+### Manual Setup
+
+#### Dependencies
  - Ethereum Node
    - Use the patched version of [Go Ethereum](https://github.com/makerdao/go-ethereum) (1.8.23+) in order to have access to storage transformers and diffs.
    - [Parity 1.8.11+](https://github.com/paritytech/parity/releases)
-   
+ - [VulcanizeDB](https://github.com/makerdao/vulcanizedb)
+ - Go 1.12+
+ - Postgres 11.2
+    
 ### Getting the project
 Download the transformer codebase to your local local `GOPATH` via: `go get github.com/makerdao/vdb-mcd-transformers`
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -11,65 +11,85 @@
...
+
+#### Dependencies
+ - Docker (the rest is handled by docker-compose)
+
+#### Database Initialization
+
+The database needs to be initialized once before bringing up the system. That can be by running, from the root directory, 
+
+```bash
+docker-compose -f dockerfiles/docker-compose.yml run execute ./run_migrations.sh
```
rmulhol, line 30: Just to double check - have you run this locally and verified the migrations run before the containers run their own migrations? Asking because I've seen the migrations take a long time (~3 minutes) before

README.md
```diff
@@ -11,65 +11,85 @@
...
-This repository is a collection of transformers to be used along with VDB as a plugin to fetch, transform and persist log events and storage slots of
-specified MCD contracts.
+This repository is a collection of transformers to be used along with VDB as a plugin to fetch, transform and persist log events and storage slots of specified MCD contracts.
 
 ## Install
-### Dependencies
+
+### Docker Setup
+
+This repository provides a docker-compose setup to manage the multiple container-based deployment for you, if at all possible you should use that setup.
```
rmulhol, line 20: This seems like the right approach, but maybe worth mentioning the need to be sure the resources allocated to Docker are substantial - I think we'll hit lower limits really fast with the ethereum node included

README.md
```diff
@@ -11,65 +11,85 @@
...
+1. `headerSync` fetches raw Ethereum data and syncs it into VulcanizeDB's Postgres database where then it can be used for transformations. More information about the `headerSync` command can be found in the [VulcanizeDB repository](https://github.com/makerdao/vulcanizedb/blob/staging/documentation/data-syncing.md#headersync).
 
-1. `execute` uses the raw Ethereum data that has been synced into Postgres and applies transformations to configured MCD
-contract data via [event](./transformers/events) and [storage](./transformers/storage) transformers. The VulcanizeDB
-repository includes a [general description](https://github.com/makerdao/vulcanizedb/blob/staging/documentation/custom-transformers.md)
-about transformers.
+1. `execute` uses the raw Ethereum data that has been synced into Postgres and applies transformations to configured MCD contract data via [event](./transformers/events) and [storage](./transformers/storage) transformers. The VulcanizeDB repository includes a [general description](https://github.com/makerdao/vulcanizedb/blob/staging/documentation/custom-transformers.md) about transformers.
```
rmulhol, line 60: Maybe doesn't need to be part of this PR, but you now also need to run `extractDiffs` to get the storage diffs used by the storage transformers

README.md
```diff
@@ -11,65 +11,85 @@
...
 
-These core commands can be run via Docker images which is the preferred method, or they can be built and run via the
-command line interface. In either method, a postgres database will first need to be created:
+The core commands can also be run via images or built and run via the command line interface. In either method, a postgres database will first need to be created:
 1. Install Postgres
-1. Create a user for yourself that is able run migrations and add extensions.
+   1. Create a user for yourself that is able run migrations and add extensions.
 1. `createdb vulcanize_public`
-1. Migrating the database manually is unnecessary as the commands handle this.
+1. Migrate the database using the `make migrate` task in this repository.
```
rmulhol, line 66: fwiw I believe `execute` and `composeAndExecute` _do_ handle running the migrations for you, but if you went that path you would need to be sure to run those commands _first_ since the vulcanizedb migrations alone will moot out migrations in this repo.

dockerfiles/execute/startup_script.sh
```diff
@@ -1,39 +1,5 @@
...
-
-# Run the DB migrations
-echo "Connecting with: $VDB_PG_CONNECT"
-./goose -dir db/migrations postgres "$VDB_PG_CONNECT" up
-
-if [ $? -ne 0 ]; then
-  echo "Could not run migrations. Are the database details correct?"
-  exit 1
-fi
+source run_migrations.sh
```
rmulhol, line 2: So this will still run the migrations automatically, right? Just double checking we don't need to modify the CD pipeline with these changes

rmulhol (commented):
Looking good! Happy to approve as long as we can verify the existing CD pipeline will continue working as expected with these changes

README.md
paytonrules: That makes a lot of sense - probably a separate PR?

README.md
paytonrules: Just double-checked and when that command is run locally it only starts the postgres image, and the execute image. The others aren't run, and the execute image actually shuts down when run_migration is done, because it overrides the startup command.

README.md
paytonrules: You're correct - I changed the directions for this reason. It might be good to re-warn the reader here of what will happen if you do things in the wrong order.

dockerfiles/execute/startup_script.sh
paytonrules: It did locally. Is there a good CD pipeline test you can think of?

paytonrules: I was able to build the execute Docker container and run execute, and sure enough it ran the migrations.

rmulhol (approved):

PR #193, opened 2020-05-21
Add support for backfilling urns from fork events

rmulhol:


transformers/component_tests/queries/flap_bid_event_computed_columns_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
 	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
 	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore"
```
elizabethengelman, line 15: I made a note in  https://github.com/makerdao/vulcanizedb/pull/96 as well, but I'm not sure I understand the switch to importing `datastore` when we're also still importing `repositories` on the line below. Not a big deal, just wanted to understand what's going on! Also, super possible that I'm having "end of the day before a day off brain". ;p

backfill/shared/dart_dink.go
```diff
@@ -21,11 +22,31 @@ var (
...
+func NewDartDinkRetriever(blockChain core.BlockChain, eventsRepository repository.EventsRepository, headerRepository datastore.HeaderRepository, storageRepository repository.StorageRepository) DartDinkRetriever {
+	return dartDinkRetriever{
+		blockChain:        blockChain,
+		eventsRepository:  eventsRepository,
+		headerRepository:  headerRepository,
+		storageRepository: storageRepository,
+	}
+}
+
+func (retriever dartDinkRetriever) RetrieveDartDinkDiffs(dartDink DartDink) error {
```
elizabethengelman, line 49: This is kind of nitpicky, but `RetrieveDartDinkDiff` seems like the method will be returning DartDinkDiffs to me, instead of the inserting them. But I'm not sure how else to name this besides going back to `FetchAndPersist...`. Definitely not merge blocking though!

elizabethengelman (approved):

backfill/shared/dart_dink.go
rmulhol: Yeah, this was a major struggle for me. I'm honestly just not sure how to name this, but `DartDinkFetchAndPersister` is where I drew the line 😅

transformers/component_tests/queries/flap_bid_event_computed_columns_test.go
rmulhol: Yeah the idea is that for the type we import the interface from `datastore` and then when we initialize we use `NewHeaderRepository` from `repositories`, but the actual implementation of `headerRepository` in `repositories` is private

PR #201, opened 2020-06-03
(VDB-1350) Add health check to execute Dockerfile

rmulhol:
Right now this only checks that the file exists, but we could also potentially check that it includes lines indicating that both the event and storage transformers are running


elizabethengelman (approved):
lgtm!

Issue #191, opened 2020-05-14
Generic Time Series Queries


hexonaut:
Continuing on from here: https://github.com/makerdao/vdb-mcd-transformers/issues/179

I think a common use case for this data is going to be displaying time series data on these values/events.

I think we should provide a common generic api for querying stats with 3 parameters. A start datetime, an end datetime and a granularity bucket size. I've added these three for collateral price, savings dai and collateral locked:

    create index if not exists headers_block_timestamp_idx ON public.headers (block_timestamp);

    drop function if exists api.spot_poke_time;
    drop type if exists api.spot_poke_time_result cascade;
    create type api.spot_poke_time_result AS (
        date timestamptz,
        ilk_id int4,
        value numeric,
        block_number int8,
        block_timestamp numeric
    );
    create function api.spot_poke_time(start_time date, end_time date, granularity interval)
        returns setof api.spot_poke_time_result as $$
            WITH input AS (
                SELECT generate_series(start_time, end_time, granularity) AS d
            ),
            ilk_values AS (
                SELECT input.d AS date, ilk_id, MAX(s.header_id) AS header_id FROM maker.spot_poke s CROSS JOIN input LEFT JOIN public.headers h ON (h.id = s.header_id) WHERE h.block_timestamp < extract(epoch FROM (input.d + granularity)) GROUP BY input.d, ilk_id
            )
            SELECT i.date, i.ilk_id, s.value, h.block_number, h.block_timestamp FROM ilk_values i LEFT JOIN maker.spot_poke s ON (s.header_id = i.header_id AND s.ilk_id = i.ilk_id) LEFT JOIN public.headers h ON (h.id = s.header_id)
        $$ language sql stable;

    drop function if exists api.pot_pie_time;
    drop type if exists api.pot_pie_time_result cascade;
    create type api.pot_pie_time_result AS (
        date timestamptz,
        pie numeric,
        block_number int8,
        block_timestamp numeric
    );
    create function api.pot_pie_time(start_time date, end_time date, granularity interval)
        returns setof api.pot_pie_time_result as $$
            WITH input AS (
                SELECT generate_series(start_time, end_time, granularity) AS d
            ),
            pie_values AS (
[...]

PR #202, opened 2020-06-03
Added a time_ilk_snapshots custom query

hexonaut:
Time series data for the `ilk_snapshot` table.

I think this is a good general format to proceed on for any time series, bucketed queries. I didn't have time today to add a test, but I can do so if required.

General time series discussion here: https://github.com/makerdao/vdb-mcd-transformers/issues/191


db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
```diff
@@ -0,0 +1,66 @@
...
+    flip TEXT,
+    rho NUMERIC,
+    duty NUMERIC,
+    pip TEXT,
+    mat NUMERIC,
+    created TIMESTAMP,
+    updated TIMESTAMP
+);
+
+CREATE FUNCTION api.time_ilk_snapshots(ilk_identifier TEXT, bucket_start TIMESTAMP, bucket_end TIMESTAMP, bucket_interval INTERVAL)
```
rmulhol: Might we want to add some defaults here, perhaps for the `bucket_interval`?

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
```diff
@@ -0,0 +1,66 @@
...
+    RETURNS SETOF api.time_ilk_snapshot AS
+$$
+WITH buckets AS (SELECT generate_series(bucket_start, bucket_end - bucket_interval, bucket_interval) AS bucket_start)
```
rmulhol: Is the term `bucket_start` being overloaded here? Maybe we could refer to one as `time_ilk_snapshots.bucket_start` to clarify which value is coming from where?

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
rmulhol: Are there any inputs that could cause this function to fail or perform in undesirable ways? e.g. if I pass in a `bucket_start` that's greater than `bucket_end`, or pass in a `bucket_interval` that's very small?

rmulhol (commented):
Thanks for contributing this! It would be cool to see some tests in https://github.com/makerdao/vdb-mcd-transformers/tree/staging/transformers/component_tests/queries so that we can document how it works and prevent regressions in future refactors

rmulhol:
I believe the build failure is because travis doesn't grant access to private build credentials for contributions from a fork - if nothing else, one of us should be able to reopen this for you to trigger a legit build once we have tests etc

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
hexonaut: Added a 1 day default.

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
hexonaut: You are correct - bucket_start was overloaded, and it's not really accurate for the function argument case. I changed the arguments to be range_start and range_end.

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
hexonaut: I just ran some tests. If `range_start` (formerly `bucket_start`) is greater than `range_end` it will just return 0 rows. This seems like the intuitive result to me.

I also tested a very small `bucket_interval` of 1 minute over a day. Other than just taking a bit of time to process, it returns the results as expected with a lot of duplicate rows if there are no changes for several blocks.

hexonaut:
I addressed the comments above. Will look to add some tests now.

hexonaut:
Added 2 tests. Looks like schema.sql got some unintended updates from `make test` as well. I can roll those back if you like.

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
hexonaut: Marking this as resolved. Re-open if this is still something to discuss.

hexonaut:
Decided it makes sense to roll back the schema.sql changes as it is for a different db version.

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
```diff
@@ -0,0 +1,66 @@
...
+    pip TEXT,
+    mat NUMERIC,
+    created TIMESTAMP,
+    updated TIMESTAMP
+);
+
+CREATE FUNCTION api.time_ilk_snapshots(ilk_identifier TEXT, range_start TIMESTAMP, range_end TIMESTAMP, bucket_interval INTERVAL DEFAULT '1 day'::INTERVAL)
+    RETURNS SETOF api.time_ilk_snapshot AS
+$$
+WITH buckets AS (SELECT generate_series(range_start, range_end - bucket_interval, bucket_interval) AS bucket_start)
```
rmulhol: Curious about a couple things - 
1) why does the series end at `range_end - bucket_interval` instead of just `range_end`?
2) why do we select `AS bucket_start`?

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
```diff
@@ -0,0 +1,66 @@
...
+SELECT buckets.bucket_start AS bucket_start,
+       buckets.bucket_start + bucket_interval AS bucket_end,
+       bucket_interval AS bucket_interval,
```
rmulhol: Why do we need the `AS` here?

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
rmulhol: The taking a bit of time part is a little bit concerning to me, if arbitrary users can input very large ranges and and very small intervals. Do you think there are any range + interval combinations that we could enable, disabling all others, to support the most common use cases? E.g. I could see enabling a version of this query that _only_ supports, say, the last 100 days with 1 day intervals, or the last week with 1 hour intervals, etc.

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
hexonaut: 1. `generate_series()` uses inclusive start and end times, and with my use case I've been using interval aligned start/end dates which are slightly easier to iterate over when the end time is exclusive. This was a stylistic choice. Thinking on it now, it's probably more intuitive to keep the range_end as inclusive as you suggest. I'll make that change.
2. This is to name the column `bucket_start` instead of the auto-generated `generate_series` as I reference in `buckets.bucket_start`. Would you prefer another style?

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
hexonaut: I was adding it for explicit consistency, but I can alter the style to omit when it's not necessary.

db/migrations/20200603124229_create_time_ilk_snapshot_query.sql
hexonaut: Update on 1:

I just remembered my original thought process. I wanted to include all full interval periods between the start and end date, and did not want to include any last interval that is not fully contained within the date range. For example, if I did not remove a `bucket_interval` from the last date then `time_ilk_snapshots('ETH-A', '2020-01-01 00:00:00', '2020-01-31 00:00:00', '1 day'::INTERVAL)` would return with the last `buckets` row being `bucket_start = 2020-01-31` and `bucket_end = 2020-02-01` which feels unintuitive to me because the end date was specified as midnight at the start of the Jan 31st and not midnight at the start of Feb 1st.

This is really a stylistic choice, and I'm happy to adjust it to whatever you prefer.

hexonaut:
Note: I want to add a count() column as well before merging.

hexonaut:
I've added a max bucket limit of 100 to all queries. The assert returns a proper error state with a clear message to the user:

```
{
  "message": "Please limit requests to at most 100 buckets.",
  "locations": [
    {
      "line": 32,
      "column": 3
    }
  ],
  "path": [
    "timeIlkSnapshots"
  ]
}
```

hexonaut:
Moved to here: https://github.com/makerdao/vdb-mcd-transformers/pull/365

PR #204, opened 2020-06-04
Vdb 1318 CI/CD protection from out of order migrations

paytonrules:
This validates that every new migration comes after those in staging, so a staging deployment shouldn't fail. It also validates that the migrations are of a valid format.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ vdb-mcd-transformers:
 # Build is really "clean/rebuild"
 .PHONY: build
 build:
-	rm vdb-mcd-transformers
+	- rm vdb-mcd-transformers
 	go fmt ./...
 	go build
 
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,17 @@ integrationtest: | $(GINKGO) $(LINT)
 	make migrate NAME=$(TEST_DB)
 	$(GINKGO) -r transformers/integration_tests/
 
+.PHONY: validatemigrationorder
+validatemigrationorder: vdb-mcd-transformers
+	./vdb-mcd-transformers checkMigrations
+
+vdb-mcd-transformers:
+	go build
+
+# Build is really "clean/rebuild"
+.PHONY: build
 build:
+	rm vdb-mcd-transformers
 	go fmt ./...
 	go build
 
--- a/cmd/backfillUrns.go
+++ b/cmd/backfillUrns.go
@@ -15,6 +15,7 @@ import (
 	"github.com/makerdao/vulcanizedb/utils"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 type backFillInitializer func(core.BlockChain, repository.EventsRepository, repository.StorageRepository, shared.DartDinkRetriever) backfill.BackFiller
@@ -48,8 +49,22 @@ Optionally pass events to watch (fork, frob, grab) to backfill based off of cert
 
 func init() {
 	rootCmd.AddCommand(backfillUrnsCmd)
+
 	backfillUrnsCmd.Flags().IntVarP(&startingBlock, "starting-block", "s", 0, "starting block for backfilling diffs derived from urn events")
 	backfillUrnsCmd.Flags().StringSliceVarP(&eventsToBackFill, "events-to-backfill", "e", []string{"fork", "frob", "grab"}, "events to back-fill")
+	backfillUrnsCmd.Flags().String("database-name", "vulcanize_public", "database name")
+	backfillUrnsCmd.Flags().Int("database-port", 5432, "database port")
+	backfillUrnsCmd.Flags().String("database-hostname", "localhost", "database hostname")
+	backfillUrnsCmd.Flags().String("database-user", "", "database user")
+	backfillUrnsCmd.Flags().String("database-password", "", "database password")
+	backfillUrnsCmd.Flags().String("client-ipcPath", "", "location of geth.ipc file")
+
+	viper.BindPFlag("database.name", backfillUrnsCmd.Flags().Lookup("database-name"))
+	viper.BindPFlag("database.port", backfillUrnsCmd.Flags().Lookup("database-port"))
+	viper.BindPFlag("database.hostname", backfillUrnsCmd.Flags().Lookup("database-hostname"))
+	viper.BindPFlag("database.user", backfillUrnsCmd.Flags().Lookup("database-user"))
+	viper.BindPFlag("database.password", backfillUrnsCmd.Flags().Lookup("database-password"))
+	viper.BindPFlag("client.ipcPath", backfillUrnsCmd.Flags().Lookup("client-ipcPath"))
 }
 
 func backfillUrns() error {
--- a/cmd/backfillUrns.go
+++ b/cmd/backfillUrns.go
@@ -4,14 +4,21 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/makerdao/vdb-mcd-transformers/backfill"
 	"github.com/makerdao/vdb-mcd-transformers/backfill/fork"
 	"github.com/makerdao/vdb-mcd-transformers/backfill/frob"
 	"github.com/makerdao/vdb-mcd-transformers/backfill/grab"
 	"github.com/makerdao/vdb-mcd-transformers/backfill/repository"
 	"github.com/makerdao/vdb-mcd-transformers/backfill/shared"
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -71,7 +71,17 @@ integrationtest: | $(GINKGO) $(LINT)
...
+validatemigrationorder: vdb-mcd-transformers
+	./vdb-mcd-transformers checkMigrations
+
+vdb-mcd-transformers:
+	go build
+
+# Build is really "clean/rebuild"
+.PHONY: build
 build:
+	rm vdb-mcd-transformers
```
rmulhol: Is this gonna run into trouble if you haven't run it before and the `vdb-mcd-transformers` binary doesn't exist?

cmd/backfillUrns.go
```diff
@@ -34,8 +42,9 @@ var backfillUrnsCmd = &cobra.Command{
 	Use:   "backfillUrns",
 	Short: "Backfill diffs for urns, looking up diffs based on associated events",
 	Long: `Fetch diffs when events indicate the state of an Urn changed at a given block.
-Optionally pass a starting block number to backfill since a given block.
+Optionally pass a staomm
```
rmulhol: ?

cmd/checkMigrations.go
```diff
@@ -0,0 +1,230 @@
...
+	localMigrations, err := getLocalMigrations()
+
+	if err != nil {
+		return err
+	}
+
+	newMigrations := NewMigrations(localMigrations, stagingMigrations)
+	logrus.Println("New Migrations are", newMigrations)
+
+	return CheckNewMigrations(stagingMigrations, newMigrations)
```
rmulhol: Would it potentially make sense to extract a lot of this code out of the `cmd` directory and test it there?

cmd/checkMigrations_test.go
```diff
@@ -0,0 +1,130 @@
...
+		})
+
+		It("does not change the new migrations passed in", func() {
+			originalMigrations := []string{"20200429072513_last.sql"}
+			newMigrations := []string{"20200429072515_newest.sql", "20200429072512_newest.sql"}
+
+			err := cmd.CheckNewMigrations(originalMigrations, newMigrations)
+
+			Expect(err).To(HaveOccurred())
+			Expect(newMigrations).To(Equal([]string{"20200429072515_newest.sql", "20200429072512_newest.sql"}))
```
rmulhol, line 108: What's the purpose of this expectation?

cmd/checkMigrations_test.go
```diff
@@ -0,0 +1,130 @@
...
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("is an error if the new migrations are out of order", func() {
+			originalMigrations := []string{"20200429072513_last.sql"}
+			newMigrations := []string{"20200429072512_newest.sql"}
+
+			err := cmd.CheckNewMigrations(originalMigrations, newMigrations)
+
+			Expect(err).To(HaveOccurred())
```
rmulhol, line 98: nbd but could potentially be cool to assert `MatchError` if we want to be getting specific errors depending on the failure

cmd/checkMigrations_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(newMigrations).To(Equal([]string{"20200429072517_newest.sql", "20200429072515_newest.sql"}))
+		})
+
+		It("requires every new migration to start with a 14 digit timestamp", func() {
+			originalMigrations := []string{"20200429072513_last.sql"}
+			newMigrations := []string{"20200429072517_newest.sql", "2020042907251z_newest.sql"}
+
+			err := cmd.CheckNewMigrations(originalMigrations, newMigrations)
```
rmulhol, line 124: probably not worth changing this up for now but I'm wondering if this will need to be changed if/when we use `goose fix` to rename migrations

cmd/checkMigrations.go
```diff
@@ -0,0 +1,230 @@
...
+	viper.BindPFlag("branch", checkMigrationsCmd.Flags().Lookup("branch"))
+}
+
+func initGithubParams(cmd *cobra.Command, args []string) {
+	repo = viper.GetString("repo")
+	branch = viper.GetString("branch")
+}
+
+func checkMigrations() error {
+	stagingMigrations, err := getGithubFileNames()
```
rmulhol: small nit but maybe this would be better named `remoteMigrations`, if we can configure the target branch to something other than staging

cmd/checkMigrations.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	localMigrations, err := getLocalMigrations()
+
+	if err != nil {
+		return err
+	}
+
+	newMigrations := NewMigrations(localMigrations, stagingMigrations)
+	logrus.Println("New Migrations are", newMigrations)
```
rmulhol, line 90: does this look alright when it prints? Would have expected to see a `: ` after "are"

cmd/checkMigrations.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func checkMigrations() error {
+	stagingMigrations, err := getGithubFileNames()
+
+	if err != nil {
+		return err
+	}
+
+	localMigrations, err := getLocalMigrations()
```
rmulhol: nbd but I'm a fan of using different names for different errors since I think you could fail to check this error and the compiler wouldn't care - but that's not true if the error var has a different name than something that was assigned and checked previously

cmd/checkMigrations.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func getGithubFileNames() ([]string, error) {
+	url := fmt.Sprintf("http://api.github.com/repos/%s/contents/db/migrations?ref=%s", repo, branch)
+	logrus.Println("Retrieving Migration list from", url)
+
+	resp, err := http.Get(url)
+
+	if err != nil {
+		return []string{}, err
```
rmulhol: could be cool to use a wrapped error here to add context, e.g.
```
return []string{}, fmt.Errorf("error getting remote migrations: %w", err)
```

rmulhol (approved):
No blockers 👍

Makefile
paytonrules: Yes! Needs a dash in front, nice catch.

cmd/backfillUrns.go
paytonrules: 🦋 ⌨️

cmd/checkMigrations.go
paytonrules: Maybe - I debated it but decided against it because I can't imagine it being used elsewhere.

cmd/checkMigrations_test.go
paytonrules: To make sure newMigrations aren't changed in the function.

cmd/checkMigrations_test.go
paytonrules: Can you match against a regex with MatchError? I don't like to match an entire string, since the tests get flaky, but I wouldn't object to matching the 'important' part.

cmd/checkMigrations_test.go
paytonrules: It shouldn't - because that will number everything and should still be fine. 

`goose fix` could be problematic BTW - because it doesn't change the database for what migrations were already run. I learned that while starting this story.

cmd/checkMigrations_test.go
elizabethengelman: I've used [`MatchRegexp`](https://onsi.github.io/gomega/#matchregexpregexp-string-args-interface) before to match the important part of the error message, but I think that's matching the error string as opposed to the error itself. So, you'd need to do `Expect(err.Error()).To(MatchRegexp("important part"))` or something like that.

elizabethengelman (approved):
Looking good! 👍

cmd/backfillUrns.go
paytonrules: Fixed - not sure how that go thtere.

cmd/checkMigrations.go
paytonrules: 👍

cmd/checkMigrations.go
paytonrules: Yup - println is smart:

```
{
  "level": "info",
  "msg": "New Migrations are [20200500000000_test.sql]",
  "time": "2020-06-10T09:14:00-05:00"
}
```

cmd/checkMigrations.go
paytonrules: 👍

cmd/checkMigrations.go
paytonrules: 👍

cmd/checkMigrations_test.go
paytonrules: That worked - perfect, thanks!

PR #205, opened 2020-06-05
(VDB-1391) USDC-B and TUSD-A

rmulhol:


transformers/shared/constants/method.go
```diff
@@ -22,7 +22,12 @@ func CdpManagerABI() string { return getContractABI("CDP_MANAGER") }
 func FlapABI() string       { return getContractABI("MCD_FLAP") }
 func FlipABI() string {
 	return GetContractsABI([]string{
-		"MCD_FLIP_ETH_A", "MCD_FLIP_BAT_A", "MCD_FLIP_SAI",
+		"MCD_FLIP_BAT_A",
+		"MCD_FLIP_ETH_A",
+		"MCD_FLIP_SAI",
+		"MCD_FLIP_USDC_A",
+		"MCD_FLIP_USDC_B",
+		"MCD_FLIP_WBTC_A",
```
elizabethengelman, line 31: Do we need to add `MCD_FLP_TUSD_A ` to this list as well?

elizabethengelman (approved):
Just one question on `transformers/shared/constants/method.go` but otherwise, LGTM! 👍

transformers/shared/constants/method.go
rmulhol: Yes, nice catch! 🙏

PR #208, opened 2020-06-09
VDB-1397 - refactor Use CTE for all_urns query

paytonrules:
This seemed significantly faster in PGAdmin, although that may be a red herring caused by load on the staging server. In addition this fixes a bug where the query returns the block_height requested instead of the actual block_height.


rmulhol (approved):
:shipit:

PR #212, opened 2020-06-17
(VDB-1363) backfill events

rmulhol:


dockerfiles/README.md
```diff
@@ -63,6 +63,18 @@ For execution on linux, replace instances of `host.docker.internal` with `localh
           --rpc --rpcaddr "0.0.0.0" --ws --wsaddr "0.0.0.0" --statediff --syncmode full --statediff.watchedaddresses <contract address>
           --statediff.watchedaddresses <contract address>
     ```
+  
+## backfillEvents
+Dockerfile for getting events from previously checked headers when adding a new event transformer.
+
+
+### Build
+`docker build -f dockerfiles/backfill_events/Dockerfile . -t backfill_eventslatest`
```
elizabethengelman: I think the tag name is missing a colon: `%s/backfill_eventslatest/backfill_events:latest`

dockerfiles/README.md
```diff
@@ -63,6 +63,18 @@ For execution on linux, replace instances of `host.docker.internal` with `localh
...
+
+## Run
+```
+docker run -e DATABASE_NAME=vulcanize_public -e DATABASE_HOSTNAME=host.docker.internal -e DATABASE_PORT=5432 -e DATABASE_USER=user -e DATABASE_PASSWORD=pw -e CLIENT_IPCPATH=https://mainnet.infura.io/v3/token -e ENDING_BLOCK_NUMBER=1 -it backfill_event:latest
```
elizabethengelman: super tiny, but the tag above in the build step is backfill_event**s** instead of backfill_event

dockerfiles/backfill_storage/Dockerfile
```diff
@@ -17,8 +17,9 @@ RUN go build
 # build mcd with local vdb
 WORKDIR /go/src/github.com/makerdao/vdb-mcd-transformers
 RUN go mod edit -replace=github.com/makerdao/vulcanizedb=/go/src/github.com/makerdao/vulcanizedb/
-RUN make plugin PACKAGE=github.com/makerdao/vdb-mcd-transformers
-
+RUN make plugin PACKAGE=github.com/makerdao/vdb-mcd-transformers \
+    TARGET_LOCATION=$GOPATH/src/github.com/makerdao/vdb-mcd-transformers/plugins/execute/transformerExporter.go \
```
elizabethengelman, line 21: Should this `TARGET_LOCATION` be pointing at `.../plugins/backfill/...` instead of `.../plugins/execute/...`?

environments/backfillEvents.toml
```diff
@@ -0,0 +1,247 @@
+[exporter]
```
elizabethengelman, line 1: I wonder if it's possible to use the `docker.toml` file instead of creating a new one, even though `docker.toml` has config for contracts that we don't necessarily need for this command. 🤔

plugins/backfill/transformerExporter.go
```diff
@@ -0,0 +1,52 @@
+// This is a plugin generated to export the configured transformer initializers
+
+package main
+
+import (
+	deny "github.com/makerdao/vdb-mcd-transformers/transformers/events/auth/deny_initializer"
```
elizabethengelman, line 7: I think that these events are ones that we've explicitly identified as ones that need to be backfilled - is the idea to avoid backfilling all of the events, because it would likely be too time consuming?

plugins/backfill/transformerExporter.go
```diff
@@ -0,0 +1,52 @@
...
+	rely "github.com/makerdao/vdb-mcd-transformers/transformers/events/auth/rely_initializer"
```
elizabethengelman, line 8: Also I was trying to think if there could be a way to configure the backfill command so that we wouldn't have to redeploy if we discover that another event also needs to be backfilled. 🤔 It seems that we'll need to run backfills when new collaterals are deployed, so we'd need to update the config file with the new collateral contracts which would trigger a redeploy anyway, so it's probably not a big deal. 🤷

elizabethengelman (approved):
LGTM! Just a few questions/small things.

dockerfiles/README.md
rmulhol: nice catch!

dockerfiles/backfill_storage/Dockerfile
rmulhol: Not at the moment, but this is definitely a bit confusing. `plugins/backfill` is used by `backfill_events` and has no storage transformers. I wonder if it would make sense to have new exporters for both `plugins/backfill_events` and a `plugins/backfill_storage` - since we probably don't need all of the transformers from `execute` when running `backfill_storage`. Open to other ideas

environments/backfillEvents.toml
rmulhol: I think that could work, and may even be preferable for reduced complexity right now (since this config includes the lowest deployment block in `docker.toml`). The only thing that had me thinking we might want to go this route was the possibility that in the future we would only need to backfill events for transformers that had been deployed pretty recently (e.g. new collateral types). In that case, pointing at `docker.toml` would mean unnecessarily rechecking all earlier blocks for all watched events (since the starting block for backfilling is derived from the lowest deployment block in the config) - and this one would let us target only more recent blocks since the new collateral contracts were deployed. Think it's worth removing this and kicking the can down the road to a later story when we cross that bridge?

plugins/backfill/transformerExporter.go
rmulhol: Yeah, basically. I don't know if log fetching performance degrades when you query for more addresses/topics, but I think minimizing the result set we get back will help avoid attempted inserts to `public.event_logs` that are bound to fail on uniqueness constraint violations. Also, eventually I'm hoping that the set of transformers we're including here would only target very recently deployed contracts, so that there'd be less work to do scanning deployment blocks for contracts that were deployed a long time ago.

plugins/backfill/transformerExporter.go
rmulhol: Yeah, definitely a bit of a bummer to need to update this whenever we have something new that needs to be backfilled, but I couldn't really think of a way around that if we're in a situation where we truly need to check many blocks for a new event 🤔 

I am hoping that we can run this container in (potentially several) disposable resources, so that the next time we update this config we could wipe out everything that's already here (since it's being taken care of in an active backfill process) and include _only_ the new transformers requiring backfill (to spin up a new backfill process for that one).

dockerfiles/backfill_storage/Dockerfile
elizabethengelman: Splitting backfill_events and backfill_storage makes sense to me!

environments/backfillEvents.toml
elizabethengelman: That's a good point that we'll likely want to backfill with a more recent block number for some transformers. 🤔  I'm good with leaving this as is, or kicking the decision down the road. 👍

paytonrules (approved):

PR #216, opened 2020-06-26
Fix abi parsing

rmulhol:
iterating through: `for _, contractName := range contractNames[:1]` means we only ever look at the first passed ABI. Fixing that led to failures where formatting differed, so we're now parsing the ABI before comparison. This required a custom function to do the comparison since equality comparisons aren't implemented on `abi.ABI`. Decided to panic if that fails so that we can see the error message (didn't happen with the old `log.Fatal` approach if the test suite was redirecting the logs to `ioutil.Discard`).

### DIFF
--- a/transformers/shared/constants/external.go
+++ b/transformers/shared/constants/external.go
@@ -20,7 +20,9 @@ import (
 	"fmt"
 	"math"
 
-	log "github.com/sirupsen/logrus"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/makerdao/vulcanizedb/pkg/eth"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 )
 
@@ -32,7 +34,7 @@ func initConfig() {
 	}
 
 	if err := viper.ReadInConfig(); err == nil {
-		log.Info("Using config file:", viper.ConfigFileUsed())
+		logrus.Info("Using config file:", viper.ConfigFileUsed())
 	} else {
 		panic(fmt.Sprintf("Could not find environment file: %v", err))
 	}
@@ -43,7 +45,7 @@ func getEnvironmentString(key string) string {
 	initConfig()
 	value := viper.GetString(key)
 	if value == "" {
-		log.Fatalf("No environment configuration variable set for key: \"%v\"", key)
+		logrus.Fatalf("No environment configuration variable set for key: \"%v\"", key)
 	}
 	return value
 }
@@ -62,7 +64,7 @@ func GetTransformerContractNames(transformerLabel string) []string {
 	configKey := "exporter." + transformerLabel + ".contracts"
 	contracts := viper.GetStringSlice(configKey)
 	if len(contracts) == 0 {
-		log.Fatalf("No contracts configured for transformer: \"%v\"", transformerLabel)
+		logrus.Fatalf("No contracts configured for transformer: \"%v\"", transformerLabel)
 	}
 	return contracts
 }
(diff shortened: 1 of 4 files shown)


transformers/shared/constants/external.go
```diff
@@ -133,3 +153,30 @@ func GetContractAddresses(contractNames []string) (addresses []string) {
 func GetContractAddress(contract string) string {
 	return getEnvironmentString("contract." + contract + ".address")
 }
+
+func parsedABIsAreEqual(a, b abi.ABI) bool {
+	if a.Constructor.String() != b.Constructor.String() {
+		return false
+	}
+OuterMethods:
```
paytonrules, line 161: I cannot think of a better way to write this code. But I sure wish I could.

transformers/shared/constants/external_test.go
```diff
@@ -0,0 +1,35 @@
...
...se,"inputs":[{"internalType":"uint256","name":"id","type":"uint256"},{"internalType":"uint256","name":"lot","type":"uint256"},{"internalType":"uint256","name":"bid","type":"uint256"}],"name":"tend","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":false,"inputs":[{"internalType":"uint256","name":"id","type":"uint256"}],"name":"tick","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[],"name":"ttl","outputs":[{"internalType":"uint48","name":"","type":"uint48"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"vat","outputs":[{"internalType":"contract VatLike","name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"wards","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[{"internalType":"uint256","name":"id","type":"uint256"}],"name":"yank","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"}]`
```
paytonrules, line 10: Do these tests need a full IRL ABI? It's a lot of noise.

transformers/shared/constants/external.go
```diff
@@ -62,40 +64,58 @@ func GetTransformerContractNames(transformerLabel string) []string {
...
 	return contracts
 }
 
-// Get the ABI for multiple contracts from config
+// GetABIFromContractsWithMatchingABI gets the ABI for multiple contracts from config
 // Makes sure the ABI matches for all, since a single transformer may run against many contracts.
-func GetContractsABI(contractNames []string) string {
+func GetABIFromContractsWithMatchingABI(contractNames []string) string {
 	initConfig()
 	if len(contractNames) < 1 {
```
paytonrules, line 76: Does initConfig take time? You could assert before the initConfig to save that.

paytonrules (approved):
I don't see any blockers. The code is weird because it's hard to understand the point of it - but it looks correct.

transformers/shared/constants/external_test.go
rmulhol: I think the answer is, unfortunately, yes - since the actual functions are reading from the `testing.toml` config. In theory we could put a shorter fake ABI in there to simplify but, despite the file's name, we have historically had it exactly match the other config in terms of transformers/contracts. Part of the motivation there is that we also use it for integration tests.

transformers/shared/constants/external.go
rmulhol: I'm not sure whether it ever takes any significant amount of time, but it seems possible that it could - given that it's performing file reads. Could you elaborate on how you'd structure an assert? Wondering if we already achieve the same goal with the `initialized` var inside `initConfig()` (which flips to `true` the first time it's called and halts subsequent runs).

transformers/shared/constants/external.go
```diff
@@ -133,3 +153,30 @@ func GetContractAddresses(contractNames []string) (addresses []string) {
...
+		for bk, bv := range b.Methods {
+			if ak == bk {
+				if av.String() == bv.String() {
+					continue OuterMethods
+				}
+			}
+		}
+		return false
+	}
+OuterEvents:
```
elizabethengelman, line 172: What is this called - when you have `OuterMethods` and `OuterEvents` defined like this within a method? I feel like I've seen this before, but can't remember how it works.

transformers/shared/constants/external.go
```diff
@@ -62,40 +64,58 @@ func GetTransformerContractNames(transformerLabel string) []string {
...
+		if !parsedABIsAreEqual(parsedABI, nextParsedABI) {
+			panic(fmt.Sprintf("ABIs don't match for contracts: %s", contractNames))
 		}
 	}
-	return abi
+	return contractABI
+}
+
+// GetFirstABI returns the ABI from the first contract in a collection in config
+func GetFirstABI(contractNames []string) string {
```
elizabethengelman, line 98: Where is this method being used?

elizabethengelman (approved):
Looks good. Just a couple questions.

transformers/shared/constants/external.go
rmulhol: I've seen it referred to [here](https://www.ardanlabs.com/blog/2013/11/label-breaks-in-go.html) as a label break

transformers/shared/constants/external.go
rmulhol: That's now used when generating the [transformer config](https://github.com/makerdao/vdb-mcd-transformers/blob/staging/transformers/shared/config.go) - since we have events that live across contracts with different ABIs (e.g. deal on flap/flip/flop)

transformers/shared/constants/external.go
paytonrules: What I mean is - couldn't you check len(contractNames) before calling init?

transformers/shared/constants/external.go
paytonrules: It's a GOTO and golang is too chicken to admit it.

PR #220, opened 2020-07-01
Further optimize allUrns query

paytonrules:
Replace DISTINCT with GROUP BY and inner join which reduces the time for me locally against staging. 

For the original query a recent run gives me:

```sql
WITH distinct_urn_snapshots AS (
    SELECT DISTINCT ON (urn_identifier, ilk_identifier)
        urn_identifier, ilk_identifier, block_height, ink, coalesce(art, 0), created, updated
        FROM api.urn_snapshot
        WHERE block_height <= api.max_block()
        ORDER BY urn_identifier, ilk_identifier, updated DESC)
SELECT *
    FROM distinct_urn_snapshots
    LIMIT NULL::integer
    OFFSET 0
```
```
Successfully run. Total query runtime: 895 msec.
8057 rows affected.
```

The new version: 
```sql
-- My database has been migrated
SELECT * from api.all_urns();
```
```
Successfully run. Total query runtime: 321 msec.
8057 rows affected.
```

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -957,14 +957,15 @@ CREATE TABLE api.urn_snapshot (
 CREATE FUNCTION api.all_urns(block_height bigint DEFAULT api.max_block(), max_results integer DEFAULT NULL::integer, result_offset integer DEFAULT 0) RETURNS SETOF api.urn_snapshot
     LANGUAGE sql STABLE
     AS $$
-WITH distinct_urn_snapshots AS (
-    SELECT DISTINCT ON (urn_identifier, ilk_identifier)
-        urn_identifier, ilk_identifier, block_height, ink, coalesce(art, 0), created, updated
-        FROM api.urn_snapshot
-        WHERE block_height <= all_urns.block_height
-        ORDER BY urn_identifier, ilk_identifier, updated DESC)
-SELECT *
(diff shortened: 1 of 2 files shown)


db/schema.sql
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 11.6
--- Dumped by pg_dump version 11.6
+-- Dumped from database version 11.8
+-- Dumped by pg_dump version 11.8
```
rmulhol: If it's possible to generate this schema without the version bump, that would be cool - think we're running 11.6 in the deploy. nbd since I don't think this schema actually gets used in the deploy, but I'd lean toward maintaining the match if it's not too much trouble (esp so we know that what we see locally reflects what we'd expect on deploy).

rmulhol (approved):
LGTM!

Could be cool to drop the benchmarks/query plans in the PR description for additional context - I wouldn't have expected a significant difference between these implementations

db/schema.sql
paytonrules: Did not mean to do a version bump.

paytonrules: 👍  - fixed

PR #222, opened 2020-07-10
Adds address_id to log_value

yaoandrew:


rmulhol (changes requested):
Adding this temporarily since it's a breaking change

paytonrules:
@yaoandrew - mostly so it doesn't get lost, I bet you could rebase this to breaking-changes.

transformers/events/log_value/transformer.go
```diff
@@ -17,67 +17,48 @@
...
 	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
 	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
 	"github.com/makerdao/vulcanizedb/pkg/core"
 	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
-	"github.com/makerdao/vulcanizedb/pkg/eth"
 )
 
 type Transformer struct{}
 
 const Val event.ColumnName = "val"
```
rmulhol: should we delete this now?

transformers/events/log_value/transformer.go
```diff
@@ -17,67 +17,48 @@
...
+		contractAddressID, contractAddressErr := shared.GetOrCreateAddress(log.Log.Address.String(), db)
+		if contractAddressErr != nil {
+			return nil, err
+		}
 
-	var models []event.InsertionModel
-	for _, logValueEntity := range entities {
-		bigIntVal := new(big.Int).SetBytes(logValueEntity.Val[:])
+		bigIntVal := new(big.Int)
+		bigIntVal.SetBytes(log.Log.Data)
```
rmulhol, line 45: mostly just curious - what inspired you to take out parsing of the entity from the ABI? This seems more elegant - I hadn't even realized we could do this 🎉

transformers/events/log_value/transformer.go
```diff
@@ -17,67 +17,48 @@
...
-	"github.com/makerdao/vulcanizedb/pkg/eth"
 )
 
 type Transformer struct{}
 
 const Val event.ColumnName = "val"
 
-func (Transformer) toEntities(contractAbi string, logs []core.EventLog) ([]LogValueEntity, error) {
-	var entities []LogValueEntity
+func (t Transformer) ToModels(abi string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
```
rmulhol: maybe worth replacing `abi` with `_`, if it's no longer used

transformers/events/log_value/transformer.go
```diff
@@ -17,67 +17,48 @@
 package log_value
 
 import (
-	"fmt"
 	"math/big"
 
 	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
```
rmulhol: super nbd but could prob remove this whitespace

db/migrations/20191118031513_create_log_value.sql
```diff
@@ -4,6 +4,7 @@ CREATE TABLE maker.log_value
     id        SERIAL PRIMARY KEY,
     header_id INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
     log_id    BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    address_id BIGINT NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
```
rmulhol: could "Reformat Code" to fix the alignment

rmulhol (commented):
:shipit:

rmulhol (approved):

transformers/events/log_value/transformer.go
yaoandrew: 😄 I'm glad it works, I think I just noticed that during the refactor that if all I needed was the data, I could grab it from the log.

PR #223, opened 2020-07-10
VDB-1395 vow file

elizabethengelman:
- Instead of calling this transformer/table `vow_file_flapper_flopper` I called it `vow_file_auction_address` since this [file method implemntation](https://github.com/makerdao/dss/blob/master/src/vow.sol#L101) is setting the flapper or flopper address on the vow contract. Totally happy to change this to something else if folks have another idea. 

- I didn't change the location or the name of the original `vow_file` transformer, and was thinking of just waiting until we pick up [VDB-1437](https://makerdao.atlassian.net/browse/VDB-1437) to make those changes. Perhaps we could rename it to `vow_file_auction_attributes` or something since that [file method implementation](https://github.com/makerdao/dss/blob/master/src/vow.sol#L92) is setting specific attributes for either the flapper or flopper on the vow contract.


db/migrations/20200709092801_create_vow_file_auction_address_table.sql
```diff
@@ -0,0 +1,20 @@
+-- +goose Up
+CREATE TABLE maker.vow_file_auction_address
+(
+    id        SERIAL PRIMARY KEY,
+    header_id INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    log_id    BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    what      TEXT,
+    data      INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
```
rmulhol, line 9: maybe worth grabbing the msg.sender as well?

transformers/integration_tests/vow_file_auction_address_test.go
```diff
@@ -0,0 +1,92 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		eventLogs := test_data.CreateLogs(header.Id, logs, db)
+
+		tr := initializer.NewTransformer(db)
+		executeErr := tr.Execute(eventLogs)
+		Expect(executeErr).NotTo(HaveOccurred())
+
+		var dbResult []vowFileAuctionAddressModel
+		getVowFileErr := db.Select(&dbResult, `SELECT what, data from maker.vow_file_auction_address`)
```
rmulhol: Super minor but if we're expecting only one result maybe we can get away with using `db.Get` and passing a single `vowFileAuctionAddressModel`? Thinking that should mean we don't need to expect len(results) to == 0

transformers/shared/constants/method.go
```diff
@@ -173,6 +173,9 @@ func vatSlipMethod() string { return getSolidityFunctionSignature(VatABI(), "sli
 func vatSuckMethod() string { return getSolidityFunctionSignature(VatABI(), "suck") }
 func vowFessMethod() string { return getSolidityFunctionSignature(VowABI(), "fess") }
 func vowFileMethod() string { return getSolidityFunctionSignature(VowABI(), "file") }
```
rmulhol: A little surprised this doesn't need to use `getOverloadedFunctionSignature` - maybe worth using anyway to be explicit about the distinction?

transformers/test_data/vow_file.go
```diff
@@ -65,3 +65,43 @@ var VowFileModel = event.InsertionModel{
 		event.LogFK:          VowFileEventLog.ID,
 	},
 }
+
+var rawVowFileAuctionAddressLog = types.Log{
+	Address: common.HexToAddress(VowAddress()),
+	Topics: []common.Hash{
+		common.HexToHash(constants.VowFileSignature()),
```
rmulhol: Should this be `VowFileAuctionAddressSignature()`?

rmulhol (approved):
LGTM! Couple minor comments, nothing blocking - would be cool to add a story for adding the `msg.sender` if we're not doing so here

transformers/integration_tests/vow_file_auction_address_test.go
elizabethengelman: Good call!

transformers/shared/constants/method.go
elizabethengelman: 🤔  Yeah, good catch. I wonder if we just lucked out before, and the file signature we were after was returned first from the abi. Updating it to use `getOverloadedFunctionSignature` makes sense to me though.

PR #225, opened 2020-07-14
vdb 1417 sender tend

yaoandrew:


rmulhol (changes requested):
dropping this here since it's a breaking change 😬

transformers/component_tests/queries/test_helpers/test_helpers.go
```diff
@@ -633,11 +633,14 @@ func CreateFlopKick(contractAddress string, bidId int, headerId, logId int64, db
 func CreateTend(input TendCreationInput) (err error) {
 	addressID, addressErr := shared.GetOrCreateAddress(input.ContractAddress, input.DB)
 	Expect(addressErr).NotTo(HaveOccurred())
+	msgSenderAddressID, msgSenderAddressErr := shared.GetOrCreateAddress(test_data.TendEventLog.Log.Topics[1].Hex(), input.DB)
```
rmulhol: I like not spending too much time here, but wonder if we might eventually want to add the `msg.sender` as part of the `input` that's passed in - if we want to assert against it. Wouldn't rush into doing this (esp as part of this PR) until taking a longer look at where this function is called

transformers/events/tend/transformer.go
```diff
@@ -52,15 +57,16 @@ func (Transformer) ToModels(_ string, logs []core.EventLog, db *postgres.DB) ([]
 			SchemaName: constants.MakerSchema,
 			TableName:  constants.TendTable,
 			OrderedColumns: []event.ColumnName{
-				event.HeaderFK, event.AddressFK, event.LogFK, constants.BidIDColumn, constants.LotColumn, constants.BidColumn,
+				event.HeaderFK, event.AddressFK, event.LogFK, constants.MsgSenderColumn, constants.BidIDColumn, constants.LotColumn, constants.BidColumn,
```
rmulhol: super nbd but might consider putting each of these on a new line

transformers/events/tend/transformer_test.go
```diff
@@ -44,11 +44,18 @@ var _ = Describe("Tend transformer", func() {
...
 
-			var addressID int64
+			var addressID, msgSenderAddressID int64
+
 			addrErr := db.Get(&addressID, `SELECT id FROM public.addresses`)
 			Expect(addrErr).NotTo(HaveOccurred())
 			expectedModel := test_data.TendModel()
 			expectedModel.ColumnValues[event.AddressFK] = addressID
+
+			msgSenderAddressErr := db.Get(&msgSenderAddressID, `SELECT id FROM addresses WHERE address = $1`,
```
rmulhol: Noticed this while working on the deal transformer as well, but idk why we're doing raw DB queries here (and above for the `addressID`) - would be a fan of using `shared.GetOrCreateAddress` (esp since I think now the query for the `addressID` could be flaky unless the contract address == `msg.sender`)

transformers/test_data/test_helpers.go
```diff
@@ -65,8 +66,13 @@ func CreateTestHeader(db *postgres.DB) int64 {
 // Create a header sync log to reference in an event, returning inserted header sync log
 func CreateTestLog(headerID int64, db *postgres.DB) core.EventLog {
```
rmulhol: seems weird that this function would create only `tend` events - wonder if it's worth creating a new function if we need this to be a `tend` somewhere (and/or change it so we can pass in the `topic0`)

transformers/component_tests/queries/test_helpers/test_helpers.go
rmulhol: Following up - definitely don't think this merits a change in this PR. Seems like it would require a lot of changes for a pretty questionable value add

transformers/events/tend/transformer_test.go
```diff
@@ -44,11 +44,18 @@ var _ = Describe("Tend transformer", func() {
 			models, err := transformer.ToModels(constants.FlipABI(), []core.EventLog{test_data.TendEventLog}, db)
 			Expect(err).NotTo(HaveOccurred())
 
-			var addressID int64
+			var addressID, msgSenderAddressID int64
+
 			addrErr := db.Get(&addressID, `SELECT id FROM public.addresses`)
```
paytonrules: This line looks like it happens to depend on the order, so it gets the right address by luck.

transformers/test_data/test_helpers.go
yaoandrew: Ah yeah.. this thing is called a bunch of other places. I should just set the topics after.

paytonrules (approved):
This ready for merge? It looks right to me.

transformers/component_tests/queries/test_helpers/test_helpers.go
```diff
@@ -633,11 +633,14 @@ func CreateFlopKick(contractAddress string, bidId int, headerId, logId int64, db
...
+	Expect(msgSenderAddressErr).NotTo(HaveOccurred())
 	tendModel := test_data.TendModel()
 	tendLog := test_data.CreateTestLog(input.TendHeaderId, input.DB)
 	tendModel.ColumnValues[constants.BidIDColumn] = strconv.Itoa(input.BidId)
 	tendModel.ColumnValues[constants.LotColumn] = strconv.Itoa(input.Lot)
 	tendModel.ColumnValues[constants.BidColumn] = strconv.Itoa(input.BidAmount)
+	tendModel.ColumnValues[constants.MsgSenderColumn] = msgSenderAddressID
```
rmulhol: Could prob use Eric's `AssignMessageSenderID` here

transformers/component_tests/triggers/bid_event_trigger_test.go
```diff
@@ -101,11 +102,17 @@ var _ = Describe("Updating bid_event table", func() {
 		address := test_data.FlipEthAddress()
 		addressID, addressErr := shared.GetOrCreateAddress(address, db)
 		Expect(addressErr).NotTo(HaveOccurred())
+
+		fakeMsgSenderAddress := fakes.RandomString(42)
+		fakeMsgSenderAddressID, fakeMsgSenderAddressErr := shared.GetOrCreateAddress(fakeMsgSenderAddress, db)
+		Expect(fakeMsgSenderAddressErr).NotTo(HaveOccurred())
```
rmulhol: ditto here

transformers/component_tests/triggers/bid_event_trigger_test.go
```diff
@@ -341,6 +348,10 @@ var _ = Describe("Updating bid_event table", func() {
 			flipRepo.SetDB(db)
 			diffID = CreateFakeDiffRecord(db)
 
+			fakeMsgSenderAddress := fakes.RandomString(42)
+			fakeMsgSenderAddressID, fakeMsgSenderAddressErr := shared.GetOrCreateAddress(fakeMsgSenderAddress, db)
+			Expect(fakeMsgSenderAddressErr).NotTo(HaveOccurred())
```
rmulhol: same

transformers/events/tend/transformer_test.go
```diff
@@ -44,11 +44,15 @@ var _ = Describe("Tend transformer", func() {
...
-			Expect(addrErr).NotTo(HaveOccurred())
+			contractAddressID, contractAddressErr := shared.GetOrCreateAddress(test_data.TendEventLog.Log.Address.String(), db)
+			Expect(contractAddressErr).NotTo(HaveOccurred())
 			expectedModel := test_data.TendModel()
-			expectedModel.ColumnValues[event.AddressFK] = addressID
+			expectedModel.ColumnValues[event.AddressFK] = contractAddressID
+
+			msgSenderAddressID, msgSenderAddressErr := shared.GetOrCreateAddress(test_data.TendEventLog.Log.Topics[1].Hex(), db)
+			Expect(msgSenderAddressErr).NotTo(HaveOccurred())
+			expectedModel.ColumnValues[constants.MsgSenderColumn] = msgSenderAddressID
```
rmulhol: here as well

transformers/storage/repository_test.go
```diff
@@ -18,6 +18,7 @@ package storage_test
 
 import (
 	"fmt"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
```
rmulhol: If you remove the blank space on line 24/25 and run `goimports -w transformers/storage/repository_test.go`, that should clean this up a bit

transformers/storage/repository_test.go
```diff
@@ -795,9 +796,19 @@ func insertFlopKicks(blockNumber int64, kicks string, contractAddressID int64, d
...
-		headerID, bidID, 0, 0, contractAddressID, tendLog.ID,
+	tendLog.Log.Topics = []common.Hash{
+		common.HexToHash(constants.TendSignature()),
+		common.HexToHash("0x0000000000000000000000003a673843d27d037b206bb05eb1abbc7288d95e66"),
+		common.HexToHash("0x000000000000000000000000000000000000000000000000000000000000000a"),
+		common.HexToHash("0x000000000000000000000000000000000000000000000000000007bb0f7b0800"),
+	}
+
+	msgSenderAddressID, msgSenderAddressErr := shared.GetOrCreateAddress(tendLog.Log.Topics[1].Hex(), db)
+	Expect(msgSenderAddressErr).NotTo(HaveOccurred())
```
rmulhol: can prob use `AssignMessageSenderID` here too

transformers/storage/repository_test.go
```diff
@@ -795,9 +796,19 @@ func insertFlopKicks(blockNumber int64, kicks string, contractAddressID int64, d
...
 	tendLog := test_data.CreateTestLog(headerID, db)
-	_, err := db.Exec(`INSERT into maker.tend (header_id, bid_id, lot, bid, address_id, log_id)
-		VALUES($1, $2::NUMERIC, $3::NUMERIC, $4::NUMERIC, $5, $6)`,
-		headerID, bidID, 0, 0, contractAddressID, tendLog.ID,
+	tendLog.Log.Topics = []common.Hash{
+		common.HexToHash(constants.TendSignature()),
+		common.HexToHash("0x0000000000000000000000003a673843d27d037b206bb05eb1abbc7288d95e66"),
+		common.HexToHash("0x000000000000000000000000000000000000000000000000000000000000000a"),
+		common.HexToHash("0x000000000000000000000000000000000000000000000000000007bb0f7b0800"),
+	}
```
rmulhol: Depending on when this gets merged, I think you might be able to use a helper created by Eric to build a test log based off of test data (and assign events that way) - https://github.com/makerdao/vdb-mcd-transformers/pull/236/files#diff-eaee5d9c3bd36e4c40f28c33bad51985R97

rmulhol (approved):
All my stuff is no longer merge blocking 👍

PR #226, opened 2020-07-14
VDB-1419 - Add msg sender to dent table

elizabethengelman:
Wondering if we want to add msg_sender to `flip_bid_event`, `flop_bid_event`, etc postgres types now, or once we've already added msg_sender to all of the associated events.

### DIFF
--- a/transformers/events/dent/transformer_test.go
+++ b/transformers/events/dent/transformer_test.go
@@ -43,16 +43,12 @@ var _ = Describe("Dent Transformer", func() {
 		models, err := transformer.ToModels(constants.FlipABI(), []core.EventLog{test_data.DentEventLog}, db)
 		Expect(err).NotTo(HaveOccurred())
 
-		msgSender := common.HexToAddress(test_data.DentEventLog.Log.Topics[1].Hex()).Hex()
-		msgSenderId, msgSenderErr := shared.GetOrCreateAddress(msgSender, db)
-		Expect(msgSenderErr).NotTo(HaveOccurred())
-
 		address := common.HexToAddress(test_data.FlipEthAddress()).Hex()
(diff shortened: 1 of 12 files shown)


yaoandrew (approved):

transformers/events/dent/transformer_test.go
```diff
@@ -43,10 +43,18 @@ var _ = Describe("Dent Transformer", func() {
 		models, err := transformer.ToModels(constants.FlipABI(), []core.EventLog{test_data.DentEventLog}, db)
 		Expect(err).NotTo(HaveOccurred())
 
+		var msgSenderId int64
+		msgSender := common.HexToAddress(test_data.DentEventLog.Log.Topics[1].Hex()).Hex()
+		msgSenderErr := db.Get(&msgSenderId, `SELECT id FROM public.addresses WHERE address = $1`, msgSender)
```
yaoandrew: This must be a pattern in our older transformers because I had similar code in my test. @rmulhol had suggested trying to get rid of the raw SQL look up here and use `shared.GetOrCreateAddress`

transformers/events/dent/transformer_test.go
```diff
@@ -43,10 +43,18 @@ var _ = Describe("Dent Transformer", func() {
...
+		Expect(msgSenderErr).NotTo(HaveOccurred())
+
 		var addressID int64
-		addrErr := db.Get(&addressID, `SELECT id FROM public.addresses`)
+		address := common.HexToAddress(test_data.FlipEthAddress()).Hex()
+		addrErr := db.Get(&addressID, `SELECT id FROM public.addresses WHERE address = $1`, address)
```
yaoandrew: Same here... could potentially use `shared.GetOrCreateAddress`

PR #227, opened 2020-07-14
(VDB-1459) Add msg.sender for maker.deal

rmulhol:


db/migrations/20180910233720_create_deal.sql
```diff
@@ -17,12 +18,9 @@ CREATE INDEX deal_bid_id_index
     ON maker.deal (bid_id);
 CREATE INDEX deal_address_index
     ON maker.deal (address_id);
+CREATE INDEX deal_msg_sender_index
+    ON maker.deal (msg_sender);
 
 
 -- +goose Down
-DROP INDEX maker.deal_address_index;
```
paytonrules, line 23: Did you mean to remove this drop? I guess maybe it's unnecessary since you're getting rid of the columns.

transformers/component_tests/queries/test_helpers/test_helpers.go
```diff
@@ -586,12 +586,15 @@ func SetUpFlopBidContext(setupData FlopBidCreationInput) (err error) {
 func CreateDeal(input DealCreationInput) (err error) {
 	addressID, addressErr := shared.GetOrCreateAddress(input.ContractAddress, input.DB)
 	Expect(addressErr).NotTo(HaveOccurred())
+	msgSenderID, msgSenderErr := shared.GetOrCreateAddress(test_data.DealEventLog.Log.Topics[1].Hex(), input.DB)
```
paytonrules: Not a blocker but you could use the helper function I added for this in test_data, since we use it so often.

paytonrules (approved):
Could probably use my magical test function in a couple spots, but otherwise looks right.

db/migrations/20180910233720_create_deal.sql
rmulhol: Yeah just figured removing it entirely was probably more simple than adding another one for the new index. Had previously been a fan of being explicit about the drops, but it seems unnecessary if we lose them when we drop the table anyway (and makes it all to easy to forget one).

PR #228, opened 2020-07-16
Vdb 1465 vow flog

paytonrules:
Creating PR to at least see it run integration tests.

### DIFF
--- a/db/migrations/20181023141856_create_vow_flog.sql
+++ b/db/migrations/20181023141856_create_vow_flog.sql
@@ -20,8 +20,5 @@ CREATE INDEX vow_flog_msg_sender_index
 
 
 -- +goose Down
-DROP INDEX maker.vow_flog_era_index;
-DROP INDEX maker.vow_flog_log_index;
-DROP INDEX maker.vow_flog_header_index;
 
 DROP TABLE maker.vow_flog;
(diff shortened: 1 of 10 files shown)


db/migrations/20181023141856_create_vow_flog.sql
```diff
@@ -14,6 +15,8 @@ CREATE INDEX vow_flog_log_index
     ON maker.vow_flog (log_id);
 CREATE INDEX vow_flog_header_index
     ON maker.vow_flog (header_id);
+CREATE INDEX vow_flog_msg_sender_index
+    ON maker.vow_flog (msg_sender);
```
elizabethengelman, line 19: Super small thing - but I wonder if we want to remove the `DROP INDEX` statements below, or add `DROP INDEX maker.vow_flog_msg_sender_index` for consistency.

I believe that all indexes are dropped when the associated table is dropped, so it's probably safe to just remove those DROPs.

elizabethengelman (approved):
LGTM!

db/migrations/20181023141856_create_vow_flog.sql
paytonrules: 👍 done.

PR #229, opened 2020-07-16
VDB-1443 msg sender for jug file

elizabethengelman:


transformers/events/jug_file/base/transformer.go
```diff
@@ -37,17 +37,24 @@ func (Transformer) ToModels(_ string, logs []core.EventLog, _ *postgres.DB) ([]e
 		what := shared.DecodeHexToText(log.Log.Topics[2].Hex())
 		data := shared.ConvertUint256HexToBigInt(log.Log.Topics[3].Hex())
 
+		msgSender := shared.GetChecksumAddressString(log.Log.Topics[1].Hex())
```
paytonrules, line 40: Why do you need GetCheckSumAddressString? I don't think I've seen this elsewhere.

transformers/events/jug_file/base/transformer_test.go
```diff
@@ -29,7 +30,14 @@ import (
...
 
 var _ = Describe("Jug file base transformer", func() {
-	var transformer = base.Transformer{}
+	var (
+		transformer = base.Transformer{}
+		db          = test_config.NewTestDB(test_config.NewTestNode())
+	)
+
+	BeforeEach(func() {
+		test_config.CleanTestDB(db)
```
paytonrules, line 39: 🎉 - although now I think I may need to add this to some of my tests.....

transformers/events/jug_file/ilk/transformer.go
```diff
@@ -35,6 +35,12 @@ func (Transformer) ToModels(contractAbi string, logs []core.EventLog, db *postgr
 			return nil, verifyErr
 		}
 
+		msgSender := shared.GetChecksumAddressString(log.Log.Topics[1].Hex())
+		msgSenderId, msgSenderErr := shared.GetOrCreateAddress(msgSender, db)
```
paytonrules: Probably want msgSenderID - the linter will get it eventually. Not a blocker.

transformers/events/jug_file/ilk/transformer_test.go
```diff
@@ -62,13 +62,15 @@ var _ = Describe("Jug file ilk transformer", func() {
 
 	It("converts a log to a model", func() {
 		models, err := transformer.ToModels(constants.JugABI(), []core.EventLog{test_data.JugFileIlkEventLog}, db)
-
 		Expect(err).NotTo(HaveOccurred())
+
 		var ilkID int64
 		ilkErr := db.Get(&ilkID, `SELECT id FROM maker.ilks where ilk = $1`, test_data.JugFileIlkEventLog.Log.Topics[2].Hex())
```
paytonrules: Not technically your code - but you could replace this with 		

```
ilkID, ilkIDErr := shared.GetOrCreateIlk(test_data.JugFileIlkEventLog.Log.Topics[2].Hex(), db)
Expect(ilkIDErr).NotTo(HaveOccurred())
```

paytonrules (approved):

transformers/events/jug_file/base/transformer.go
elizabethengelman: I just added this as a helper so avoid converting the topic hex to an address, and then back to hex again, eg: 
 `common.HexToAddress(log.Log.Topics[1].Hex()).Hex()`.

Not sure if this is a pattern we want to use, but thought I'd see how it looked. Happy to change it back if that makes more sense.

PR #236, opened 2020-07-20
Vdb 1463 Flip

paytonrules:


db/migrations/20180912171047_create_cat_file.sql
```diff
@@ -34,6 +35,8 @@ CREATE INDEX cat_file_flip_log_index
     ON maker.cat_file_flip (log_id);
 CREATE INDEX cat_file_flip_ilk_index
     ON maker.cat_file_flip (ilk_id);
+CREATE INDEX cat_file_flip_msg_sender_index
+    ON maker.cat_file_flip (msg_sender);
```
elizabethengelman, line 42: I wonder if we want to either include a `DROP INDEX maker.cat_file_chop_msg_sender_index;` or remove the `DROP INDEX` statements from this migration for consistency.

elizabethengelman (approved):
This LGTM once the build passes. I looked at it briefly, and it seems like it could be failing on an integration test that times out because the build in general seems to be slow today.

db/migrations/20180912171047_create_cat_file.sql
paytonrules: Looks like somebody (Rob) took care of this on a rebase.

PR #238, opened 2020-07-21
(VDB-1475) Add msg.sender for Yank

rmulhol:


transformers/events/yank/transformer.go
```diff
@@ -32,6 +32,10 @@ func (Transformer) ToModels(_ string, logs []core.EventLog, db *postgres.DB) (re
 		if validationErr != nil {
 			return nil, validationErr
 		}
+		msgSenderID, msgSenderErr := shared.GetOrCreateAddress(log.Log.Topics[1].Hex(), db)
+		if msgSenderErr != nil {
+			return nil, shared.ErrCouldNotCreateFK(msgSenderErr)
```
paytonrules, line 37: Would it make sense to move the ErrCouldNotCreateFK into GetOrCeateAddress?

Not that it's a blocker, but I notice we are reproducing that error everywhere - and it always needs to be the same thing.

paytonrules (approved):

transformers/events/yank/transformer.go
rmulhol: That makes sense to me - would ideally decouple that change from this story, though, since it's going to affect a lot of places (also not sure if that change would ultimately need to be made in vdb)

PR #241, opened 2020-07-23
Vdb 1477 add msg sender to pot file

elizabethengelman:


transformers/events/pot_file/dsr/transformer.go
```diff
@@ -8,15 +8,23 @@ import (
...
 	var results []event.InsertionModel
 	for _, log := range logs {
 		verifyErr := shared.VerifyLog(log.Log, shared.FourTopicsRequired, shared.LogDataNotRequired)
 		if verifyErr != nil {
 			return nil, verifyErr
 		}
+		msgSender := shared.GetChecksumAddressString(log.Log.Topics[1].Hex())
+		msgSenderID, msgSenderErr := shared.GetOrCreateAddress(msgSender, db)
+		if msgSenderErr != nil {
+			return nil, msgSenderErr
```
rmulhol: not sure if the value add is worth revising, but I think in a lot of places we use `shared.ErrCouldNotCreateFK` to wrap errors like this

rmulhol (approved):
:shipit:

PR #242, opened 2020-07-23
Add msg sender to integration tests

elizabethengelman:


transformers/integration_tests/tend_test.go
```diff
@@ -76,13 +76,19 @@ var _ = Describe("Tend EventTransformer", func() {
...
 
 		var flipTend tendModel
-		err = db.Get(&flipTend, `SELECT address_id, bid, bid_id, lot FROM maker.tend`)
+		err = db.Get(&flipTend, `SELECT address_id, msg_sender, bid, bid_id, lot FROM maker.tend`)
 		Expect(err).NotTo(HaveOccurred())
 
 		flipAddressID, addrErr := shared.GetOrCreateAddress(test_data.FlipEthAddress(), db)
 		Expect(addrErr).NotTo(HaveOccurred())
+
+		msgSender := shared.GetChecksumAddressString("0x00000000000000000000000000abe7471ec9b6953a3bd0ed3c06c46f29aa4280")
```
yaoandrew, line 85: Oh interesting. Are you using `shared.GetChecksumAddressString` because we might get back mixed case if you get it direct from the event log topics?

yaoandrew (approved):

transformers/integration_tests/tend_test.go
elizabethengelman: Yep, exactly. I grabbed this directly from etherscan... and instead of converting it to the checksum version using that in the test, or converting it inline, I just decided to create this helper.

PR #244, opened 2020-07-23
Vdb 1514 Backfill Storage

paytonrules:
- Suspect I'm missing something. We'll need to get info from Sava on how he wants to deploy this.


environments/backfillStorage.toml
```diff
@@ -0,0 +1,196 @@
...
+    name     = "transformerExporter"
+    save     = false
+    transformerNames = [
+        "knc_flip",
+        "tusd_flip",
+        "usdc_a_flip",
+        "usdc_b_flip",
+        "wbtc_flip",
+        "zrx_flip",
+        "median_knc",
```
rmulhol, line 12: one of us is going to need to rebase to get `median_mana` added to this list

image

environments/backfillStorage.toml
```diff
@@ -0,0 +1,196 @@
...
+    home     = "github.com/makerdao/vulcanizedb"
+    name     = "transformerExporter"
+    save     = false
+    transformerNames = [
+        "knc_flip",
+        "tusd_flip",
+        "usdc_a_flip",
+        "usdc_b_flip",
+        "wbtc_flip",
+        "zrx_flip",
```
rmulhol, line 11: ah same for renaming these flip addresses

environments/backfillStorage.toml
```diff
@@ -0,0 +1,196 @@
...
+        deployed = 10322969
```
rmulhol, line 196: I think we might be able to remove a bunch of the addresses here that aren't used by the included transformers, but probably good to keep them available? Also not sure if that would break the build due to failed config reading - but if so maybe that's worth thinking about (in another ticket of course)

rmulhol (approved):
🚀

environments/backfillStorage.toml
paytonrules: Almost certainly, but I felt like cross checking addresses with contracts was likely to be error-prone. So I figured it won't hurt to keep them.

PR #246, opened 2020-07-24
Vdb 1466 maker spot file mat

paytonrules:


elizabethengelman (approved):
Is there an integration test that we'd want to add a msg_sender assertion?

Otherwise, this LGTM!

paytonrules:
@elizabethengelman - Good point. Why don't I do a PR for an integration test of all 3 spot-file types when they are done.

PR #247, opened 2020-07-27
Vdb 1468  - Add Message Sender to Spot File Par

paytonrules:


transformers/events/spot_file/par/transformer_test.go
```diff
@@ -14,6 +15,7 @@ import (
 
 var _ = Describe("Spot file par transformer", func() {
 	var transformer = par.Transformer{}
+	db := test_config.NewTestDB(test_config.NewTestNode())
```
rmulhol: nbd but I think we probably want to include a `BeforeEach` here to clean the test DB before each `It`. Also I _think_ it's good to explicitly declare the `db` a `var`, but don't remember why 🤔

rmulhol (approved):

transformers/events/spot_file/par/transformer_test.go
paytonrules: 👍

PR #250, opened 2020-07-27
Integration tests for spot file

paytonrules:
This updates two integration tests for the Spot File contract, the Par event still doesn't have an example on Etherscan.

This is draft because I don't want to merge it until the other Spot File transformers are merged.


rmulhol (approved):
🚢

PR #234, opened 2020-07-20
VDB-1437 Changes naming for vow file auction attributes

yaoandrew:
NOTE: This contains breaking changes as we are renaming tables in an old migration.

PR #251, opened 2020-07-27
Vdb 1472 vow file msg sender

yaoandrew:
This was branched off of #234. Will rebase before merging.


db/migrations/20200709092801_create_vow_file_auction_address_table.sql
```diff
@@ -12,7 +12,7 @@ CREATE TABLE maker.vow_file_auction_address
 
 CREATE INDEX vow_file_auction_address_header_index
     ON maker.vow_file_auction_address (header_id);
-CREATE INDEX vvow_file_auction_address_log_index
```
paytonrules, line 15: vvvvvvvvvvow......

scripts/check_config.sh
```diff
@@ -14,6 +14,8 @@ fi
 
 for ((i=0; i<${#exportedTransformers[@]}; i++)); do
   if [ ${exportedTransformers[$i]} != ${configTransformers[$i]} ]; then
+    echo config contains: ${configTransformers[$i]}
```
paytonrules: 🎉

.travis.yml
```diff
@@ -17,7 +17,6 @@ before_install:
 script:
 - set -e
 - scripts/check_config.sh
-- env GO111MODULE=on make validatemigrationorder
```
rmulhol, line 20: Makes sense this is necessary, but do you think we should create a story to restore this at some point in the future?

transformers/events/vow_file/auction_attributes/transformer.go
```diff
@@ -26,28 +27,39 @@ import (
...
-		err := shared.VerifyLog(log.Log, shared.FourTopicsRequired, shared.LogDataRequired)
-		if err != nil {
-			return nil, err
+		verifyLogErr := shared.VerifyLog(log.Log, shared.FourTopicsRequired, shared.LogDataRequired)
+		if verifyLogErr != nil {
+			return nil, verifyLogErr
+		}
+
+		msgSenderAddress := common.HexToAddress(log.Log.Topics[1].Hex()).Hex()
+		msgSenderAddressId, msgSenderAddressErr := shared.GetOrCreateAddress(msgSenderAddress, db)
```
paytonrules: Golint gets sad when Id isn't `ID`

transformers/events/vow_file/auction_attributes/transformer.go
```diff
@@ -26,28 +27,39 @@ import (
...
+		if msgSenderAddressErr != nil {
+			return nil, msgSenderAddressErr
```
paytonrules: The error is supposed to be wrapped in a `ErrCouldNotCreateFK` error like so `return nil, shared.ErrCouldNotCreateFK(msgSenderErr)`

I wonder if we could make `GetOrCreateAddress` since we are doing that everywhere.

transformers/events/vow_file/auction_attributes/transformer_test.go
```diff
@@ -42,10 +50,17 @@ var _ = Describe("Vow file transformer", func() {
 	})
 
 	It("converts a log to a model", func() {
-		models, err := transformer.ToModels(constants.VowABI(), []core.EventLog{test_data.VowFileEventLog}, nil)
+		models, toModelsErr := transformer.ToModels(constants.VowABI(), []core.EventLog{test_data.VowFileAuctionAttributesEventLog}, db)
+		Expect(toModelsErr).NotTo(HaveOccurred())
+
+		var msgSenderAddressID int64
+		msgSenderAddressErr := db.Get(&msgSenderAddressID, `SELECT id from addresses WHERE address = $1`,
```
paytonrules: Don't you want to use GetOrCreateAddress here?

transformers/integration_tests/vow_file_auction_attributes_test.go
```diff
@@ -37,23 +37,23 @@ var _ = Describe("VowFile LogNoteTransformer", func() {
...
 		initializer = event.ConfiguredTransformer{
-			Config:      vowFileConfig,
-			Transformer: vow_file.Transformer{},
+			Config:      vowFileAuctionAttributesConfig,
+			Transformer: auction_attributes.Transformer{},
 		}
 	})
 
-	It("fetches and transforms a Vow.file event", func() {
+	It("fetches and transforms a Vow.file auction attributes event", func() {
```
paytonrules, line 56: You could check for a msgSender value as well here too. I don't think it's consistent, but i know many tests are checking.

paytonrules (approved):
A few minor things - looks good!

transformers/integration_tests/vow_file_auction_attributes_test.go
```diff
@@ -71,17 +71,24 @@ var _ = Describe("VowFile LogNoteTransformer", func() {
 		err = tr.Execute(eventLogs)
 		Expect(err).NotTo(HaveOccurred())
 
-		var dbResult []vowFileModel
-		err = db.Select(&dbResult, `SELECT what, data from maker.vow_file`)
+		var msgSenderId int64
+		msgSenderAddress := common.HexToAddress(eventLogs[0].Log.Topics[1].String()).Hex()
```
rmulhol: I think we should replace `eventLogs[0].Log.Topics[1].String()` with `0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB` so that we know we're matching the actual address and not failing to parse the topic

PR #245, opened 2020-07-24
(VDB-1540) Add new auction contracts to beta

rmulhol:
- [X] Flap
- [X] Flip BAT-A
- [X] Flip ETH-A
- [x] Flip KNC-A
- [x] Flip MANA-A
- [x] Flip TUSD-A
- [X] Flip USDC-A
- [X] Flip USDC-B
- [x] Flip WBTC-A
- [x] Flip ZRX-A
- [X] Flop

PR #252, opened 2020-07-27
(VDB-1540) Add new auctions addresses

rmulhol:
Note: this is largely the same code as #245 - targeting `breaking-changes` instead of `beta`


dockerfiles/docker-compose.yml
```diff
@@ -101,19 +101,40 @@ services:
...
+      "--statediff.watchedaddresses", "0x92645a34d07696395b6e5b8330b000D0436A9aAD", # CONTRACT_MCD_FLIP_ZRX_A_1.0.9_ADDRESS
+      "--statediff.watchedaddresses", "0x4d95a049d5b0b7d32058cd3f2163015747522e99", # CONTRACT_MCD_FLOP_1.0.1_ADDRESS
+      "--statediff.watchedaddresses", "0xA41B6EF151E06da0e34B009B86E828308986736D", # CONTRACT_MCD_FLOP_1.0.9_ADDRESS
       "--statediff.watchedaddresses", "0x19c0976f590d67707e62397c87829d896dc0f1f1", # CONTRACT_MCD_JUG_ADDRESS
       "--statediff.watchedaddresses", "0x197e90f9fad81970ba7976f33cbd77088e5d7cf7", # CONTRACT_MCD_POT_ADDRESS
       "--statediff.watchedaddresses", "0x65c79fcb50ca1594b025960e539ed7a9a6d434a3", # CONTRACT_MCD_SPOT_ADDRESS
       "--statediff.watchedaddresses", "0x35d1b3f3d7966a1dfe207aa4514c12a259a0492b", # CONTRACT_MCD_VAT_ADDRESS
       "--statediff.watchedaddresses", "0xa950524441892a31ebddf91d3ceefa04bf454466", # CONTRACT_MCD_VOW_ADDRESS
-      "--statediff.watchedaddresses", "0x81FE72B5A8d1A857d176C3E7d5Bd2679A9B85763", # CONTRACT_OSM_ETH_ADDRESS
-      "--statediff.watchedaddresses", "0xb4eb54af9cc7882df0121d26c5b97e802915abe6", # CONTRACT_OSM_BAT_ADDRESS
```
elizabethengelman, line 117: How come we're removing these OSM addresses from the watched address?

environments/backfillEvents.toml
```diff
@@ -28,43 +28,60 @@
         type = "eth_event"
         repository = "github.com/makerdao/vdb-mcd-transformers"
         migrations = "db/migrations"
-        contracts = ["MCD_FLAP", "MCD_FLIP_BAT_A", "MCD_FLIP_ETH_A", "MCD_FLIP_KNC_A", "MCD_FLIP_SAI",
-                     "MCD_FLIP_TUSD_A", "MCD_FLIP_USDC_A", "MCD_FLIP_USDC_B", "MCD_FLIP_WBTC_A", "MCD_FLIP_ZRX_A",
-                     "MCD_FLOP"]
+        contracts = ["MCD_FLAP_1.0.0", "MCD_FLAP_1.0.9", "MCD_FLIP_BAT_A_1.0.0", "MCD_FLIP_BAT_A_1.0.9",
```
elizabethengelman, line 31: Since the prod environment will be syncing from the beginning, I wonder if we potentially don't need anything configured in the backfill config at the beginning. Totally cool with punting on this decision/conversation too, as it's kind of separate from this PR.

dockerfiles/docker-compose.yml
rmulhol: Thinking that if we don't have storage transformers for the OSM contracts, then we might as well not watch those addresses in the node - to reduce extraneous rows in the `public.storage_diff` table that might slow down processing. Think we can add it back and/or backfill if we decide to write those transformers

environments/backfillEvents.toml
rmulhol: Yeah hoping we don't need to backfill, so could wipe this file - was mostly just adding compatibility (the version of this config currently on `breaking-changes` breaks the build for this PR since the directories are no longer recognized)

dockerfiles/docker-compose.yml
elizabethengelman: 👍 Totally makes sense.

transformers/events/DOCUMENTATION.md
```diff
@@ -27,7 +27,7 @@ If the contract isn't already present in the environment you'll need to add it b
    
     ``` toml
     [contract]
-    [contrac t.MCD_FLIP_ETH_A]
```
elizabethengelman, line 30: looks like this should be updated from `contrac t` -> `contract`

transformers/integration_tests/deny_test.go
```diff
@@ -21,49 +21,104 @@ var _ = Describe("Deny transformer", func() {
...
+		denyIntegrationTest(int64(8928163), test_data.FlapV100Address(), usrAddress, usrAddress)
 	})
 
-	Context("Flip ETH deny events", func() {
+	Context("Flap v1.0.9 deny events", func() {
+		usrAddress := "0xdDb108893104dE4E1C6d0E47c42237dB4E617ACc"
+		denyIntegrationTest(int64(10510886), test_data.FlapV109Address(), usrAddress, usrAddress)
+	})
+
+	Context("Flip BAT v1.0.0 deny events", func() {
```
elizabethengelman, line 34: Small thing - but should we have a test for Flip BAT v1.0.9 deny events too?

elizabethengelman (approved):
whew, thanks for taking care of these changes! 😅 🙏

transformers/integration_tests/deny_test.go
rmulhol: Yes! Nice catch 👍

paytonrules (approved):

PR #255, opened 2020-07-29
(VDB-1448) Use numeric indexes for core VDB migrations

rmulhol:
- prevent core migrations from being interspersed with plugin-specific
  changes - keep them all at the beginning


paytonrules (approved):
👍

PR #259, opened 2020-08-03
VDB-1368: Add Dockerfile for extract diffs cmd

elizabethengelman:
This PR needs to work with [this VDB PR](https://github.com/makerdao/vulcanizedb/pull/119). Before the VDB PR is merged in, pass `VDB_VERSION=VDB-1368-add-new-geth-patch-as-diff-source` as a build arg while building this container, eg: 

```
docker build -f dockerfiles/extract_diffs/Dockerfile . -t extract --build-arg VDB_VERSION=VDB-1368-add-new-geth-patch-as-diff-source
```


dockerfiles/extract_diffs/startup_script.sh
```diff
@@ -0,0 +1,71 @@
...
+./goose -dir db/migrations postgres "$VDB_PG_CONNECT" up
+
+if [ $? -ne 0 ]; then
+  echo "Could not run migrations. Are the database details correct?"
+  exit 1
+fi
+
+# Run extractDiffs
+echo "Running extractDiffs..."
+./vulcanizedb extractDiffs --watchedAddresses=0x78f2c2af65126834c51822f56be0d7469d7a523e \
```
elizabethengelman: Unfortunately, I don't think that we can add a comment for each address telling which contract it belongs to. Not sure if that's ideal, so I may play around with just using a .toml config file like we're doing with execute.

yaoandrew (approved):
LGTM - I was able to build the container locally. Of course I didn't read your note at first so I got an error when I tried to run it without `--build-arg` set🤣

elizabethengelman:
I've added extract diffs to the travis deploy script since the last review and would love another set of eyes!

environments/extractDiffs.toml
```diff
@@ -0,0 +1,125 @@
...
+        deployed = 10322821
+    [contract.OASIS_MATCHING_MARKET_ONE]
```
rmulhol: Might be able to remove the Oasis + OSM contracts?

dockerfiles/extract_diffs/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/dockerfiles/extract_diffs/startup_script.sh .
+COPY --from=builder /go/src/github.com/makerdao/vulcanizedb/db/migrations/* db/migrations/
+COPY --from=builder /go/src/github.com/pressly/goose/cmd/goose/goose goose
+# needed for waiting until postgres is ready before starting from docker-compose
+COPY --from=builder /go/src/github.com/makerdao/vulcanizedb/dockerfiles/wait-for-it.sh .
+
+HEALTHCHECK CMD test -f /tmp/connection
+
+# need to execute with a shell to access env variables
+CMD ["./startup_script.sh"]
```
rmulhol, line 56: One thing that bit me recently was forgetting to update permissions on the `startup_script.sh` file - if you haven't run this locally, maybe worth verifying that won't be an issue here

.travis/deploy.sh
```diff
@@ -28,6 +28,9 @@ fi
 message BUILDING EXECUTE DOCKER IMAGE
 docker build -f dockerfiles/execute/Dockerfile . -t makerdao/vdb-execute:$TAG
 
+message BUILDING EXTRACT-DIFFS DOCKER IMAGE
+docker build -f dockerfiles/extract_diffs/Dockerfile . -t makerdao/vdb-extract-diffs:$TAG
```
rmulhol, line 32: Since this is moving over here, do we want a corresponding PR to remove it from vdb?

rmulhol (approved):
A couple questions, but LGTM 👍

.travis/deploy.sh
elizabethengelman: 👍 I think that https://github.com/makerdao/vulcanizedb/pull/119 should take care of this.

environments/extractDiffs.toml
elizabethengelman: Updated in c0947f8

dockerfiles/extract_diffs/Dockerfile
elizabethengelman: I think this should be okay - I've changed the script's permissions, and can run the docker container on my machine.

PR #260, opened 2020-08-05
VDB-1425: Update VDB version

elizabethengelman:
- Updates VDB to use https://github.com/makerdao/vulcanizedb/pull/120
- This is part of the story to remove hashed storage keys. I think all of the hashing is done in VDB, and the only change that is happening here is to update the plugin integration test. But would love another set of eyes on it!


rmulhol (approved):
:shipit:

paytonrules (approved):

PR #262, opened 2020-08-10
Update permissions for backfill_events startup script

elizabethengelman:
changing the permissions for the backfill_events startup script so that it is executable


rmulhol (approved):
🙏

paytonrules (approved):

PR #266, opened 2020-08-11
Update imports to match vdb

rmulhol:
- new Geth version


transformers/events/new_cdp/transformer.go
```diff
@@ -19,61 +19,39 @@ package new_cdp
...
 )
 
 type Transformer struct{}
 
-func (Transformer) toEntities(contractAbi string, logs []core.EventLog) ([]NewCdpEntity, error) {
-	var entities []NewCdpEntity
+func (t Transformer) ToModels(_ string, logs []core.EventLog, _ *postgres.DB) ([]event.InsertionModel, error) {
+	var models []event.InsertionModel
 	for _, log := range logs {
-		var entity NewCdpEntity
```
elizabethengelman, line 36: Trying to wrap my head around this change... looks like we're not using `UnpackLog` anymore, and doing it manually now. Is this just because `UnpackLog` isn't able to unpack the ID into `big.Int` anymore? Wondering if this is going to affect other transformers too?

elizabethengelman (approved):
🙏

transformers/events/new_cdp/transformer.go
rmulhol: The issue seemed to be, as far as I can tell, that `UnpackLog` doesn't deal well with cases where all of the payload is in the `topics` and not the `data` - but the `data` payload is non-empty. We were hitting this error: https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/argument.go#L95

On a closer look, though, it seems like the problem might be that our test data erroneously includes a non-empty data payload - not seeing any `data` for similar logs on Etherscan. Thinking maybe it makes sense to clean up the test data, but I'm on the fence about whether we should restore `toEntities` (since parsing this payload is fairly trivial without that tooling) 🤔

elizabethengelman: Yeah, that totally makes to me to stay with not using `UnpackLog` for this one. Also good news that it may be a test data issue - I was thinking that it would be surprising that nothing else would have been affected by this change.

PR #269, opened 2020-08-13
Add ContractAddressID to cat storage tables

paytonrules:
A few things that make this larger than might be expected.

* The data_generator needed a hack because CatIlkChop and CatIlkLump now require an addressID, but the rest of the ilk tables don't.
* InsertFieldWithIlkAndAddress was added to shared/utlities, because not every InsertFieldWithIlk needs an address
* I changed MappingResWithAddress to take an int64. Since this is just in test code, it should be fine, but it caused a large number of cascading changes I didn't anticipate.

### DIFF
--- a/data_generator/data_generator.go
+++ b/data_generator/data_generator.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/jmoiron/sqlx"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
 	"github.com/makerdao/vdb-mcd-transformers/transformers/storage/cat"
 	"github.com/makerdao/vdb-mcd-transformers/transformers/storage/jug"
 	"github.com/makerdao/vdb-mcd-transformers/transformers/storage/test_helpers"
@@ -397,17 +398,30 @@ func (state *GeneratorState) insertInitialIlkData(ilkId int64) error {
 		cat.InsertCatIlkLumpQuery,
 	}
 
+	contractAddressID, contractAddressErr := shared.GetOrCreateAddress(getRandomAddress(), state.db)
+	if contractAddressErr != nil {
+		return fmt.Errorf("error inserting random contract address: %w", contractAddressErr)
+	}
+
 	for _, intInsertSql := range intInsertions {
-		_, err := state.pgTx.Exec(intInsertSql, state.currentDiffID, state.currentHeader.Id, ilkId, rand.Int())
+		// HACK because the cat queries now take an address_id. If the other storage queries start taking an
+		// address_id, remove this hack
+		var err error
+		if intInsertSql == cat.InsertCatIlkChopQuery || intInsertSql == cat.InsertCatIlkLumpQuery {
+			_, err = state.pgTx.Exec(intInsertSql, state.currentDiffID, state.currentHeader.Id, contractAddressID, ilkId, rand.Int())
+		} else {
+			_, err = state.pgTx.Exec(intInsertSql, state.currentDiffID, state.currentHeader.Id, ilkId, rand.Int())
+		}
+
 		if err != nil {
-			return fmt.Errorf("error inserting initial ilk data: %v", err)
+			return fmt.Errorf("error inserting initial ilk data: %w", err)
 		}
 	}
 	_, flipErr := state.pgTx.Exec(cat.InsertCatIlkFlipQuery,
-		state.currentDiffID, state.currentHeader.Id, ilkId, test_data.AlreadySeededRandomString(10))
+		state.currentDiffID, state.currentHeader.Id, contractAddressID, ilkId, test_data.AlreadySeededRandomString(10))
 
 	if flipErr != nil {
-		return fmt.Errorf("error inserting initial ilk data: %v", flipErr)
+		return fmt.Errorf("error inserting flip ilk data: %v", flipErr)
 	}
 
 	return nil
--- a/transformers/component_tests/cat/configured_transformer_test.go
+++ b/transformers/component_tests/cat/configured_transformer_test.go
@@ -38,20 +38,21 @@ import (
 
 var _ = Describe("Executing the transformer", func() {
 	var (
-		db                = test_config.NewTestDB(test_config.NewTestNode())
-		contractAddress   = test_data.CatAddress()
-		keccakOfAddress   = types.HexToKeccak256Hash(contractAddress)
-		repository        = cat.StorageRepository{ContractAddress: contractAddress}
-		storageKeysLookup = storage.NewKeysLookup(cat.NewKeysLoader(&mcdStorage.MakerStorageRepository{}, contractAddress))
-		transformer       = storage.Transformer{
+		db              = test_config.NewTestDB(test_config.NewTestNode())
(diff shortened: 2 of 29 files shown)


data_generator/data_generator.go
```diff
@@ -397,17 +398,30 @@ func (state *GeneratorState) insertInitialIlkData(ilkId int64) error {
...
+	if contractAddressErr != nil {
+		return fmt.Errorf("error inserting random contract address: %w", contractAddressErr)
+	}
+
 	for _, intInsertSql := range intInsertions {
-		_, err := state.pgTx.Exec(intInsertSql, state.currentDiffID, state.currentHeader.Id, ilkId, rand.Int())
+		// HACK because the cat queries now take an address_id. If the other storage queries start taking an
+		// address_id, remove this hack
+		var err error
+		if intInsertSql == cat.InsertCatIlkChopQuery || intInsertSql == cat.InsertCatIlkLumpQuery {
```
rmulhol, line 410: I guess we never deal with `cat_ilk_flip` in here?

transformers/component_tests/cat/configured_transformer_test.go
```diff
@@ -136,22 +145,26 @@ var _ = Describe("Executing the transformer", func() {
...
 	Describe("ilk", func() {
 		var (
-			ilk    string
-			ilkID  int64
-			ilkErr error
+			ilk                string
+			ilkID              int64
+			ilkErr             error
+			contractAddressID  int64
+			contractAddressErr error
```
rmulhol: nbd but I feel like `ilkErr` and `contractAddressErr` could be declared within the `BeforeEach` so that they're not in scope for subsequent `It` statements - seems like we only care about them in the expectations on lines 165 + 167

transformers/component_tests/flap/configured_transformer_test.go
```diff
@@ -189,7 +189,7 @@ var _ = Describe("Executing the flap transformer", func() {
 			var wardsResult test_helpers.MappingResWithAddress
 			err := db.Get(&wardsResult, `SELECT diff_id, header_id, address_id, usr AS key, wards.wards AS value FROM maker.wards`)
 			Expect(err).NotTo(HaveOccurred())
-			Expect(wardsResult.AddressID).To(Equal(strconv.FormatInt(flapAddressID, 10)))
+			Expect(wardsResult.AddressID).To(Equal(flapAddressID))
```
rmulhol: 😤

transformers/shared/utilities.go
```diff
@@ -124,6 +124,37 @@ func InsertFieldWithIlk(diffID, headerID int64, ilk, variableName, query, value
 	return tx.Commit()
 }
 
+func InsertFieldWithIlkAndAddress(diffID, headerID int64, address, ilk, variableName, query, value string, db *postgres.DB) error {
+	tx, txErr := db.Beginx()
+	if txErr != nil {
+		return fmt.Errorf("error beginning transaction: %w", txErr)
+	}
+
+	addressID, addressErr := GetOrCreateAddress(address, db)
```
rmulhol: Maybe worth using `GetOrCreateAddressInTransaction` here, and rolling back on failure?

transformers/storage/cat/repository.go
```diff
@@ -51,23 +51,38 @@ func (repository *StorageRepository) SetDB(db *postgres.DB) {
 }
 
 func (repository *StorageRepository) insertLive(diffID, headerID int64, live string) error {
-	_, err := repository.db.Exec(insertCatLiveQuery, diffID, headerID, live)
+	addressID, addressErr := shared.GetOrCreateAddress(repository.ContractAddress, repository.db)
```
rmulhol: Given the number of times that these inserts can happen, I think it might be cool to have a singleton-ish helper function that memoizes the addressID the first time it's called - so that we don't need to go to the DB for it every time

transformers/storage/cat/repository.go
```diff
@@ -51,23 +51,38 @@ func (repository *StorageRepository) SetDB(db *postgres.DB) {
...
+	if addressErr != nil {
+		return fmt.Errorf("Could not retrieve address id for %s, error: %w", repository.ContractAddress, addressErr)
```
rmulhol: incredibly minor but I think the linter complains about capitalization in error strings

transformers/component_tests/cat/configured_transformer_test.go
```diff
@@ -136,22 +145,26 @@ var _ = Describe("Executing the transformer", func() {
 			var wardsResult test_helpers.MappingResWithAddress
 			err := db.Get(&wardsResult, `SELECT diff_id, header_id, address_id, usr AS key, wards.wards AS value FROM maker.wards`)
 			Expect(err).NotTo(HaveOccurred())
-			Expect(wardsResult.AddressID).To(Equal(strconv.FormatInt(catAddressID, 10)))
+			Expect(wardsResult.AddressID).To(Equal(catAddressID))
 			test_helpers.AssertMapping(wardsResult.MappingRes, wardsDiff.ID, header.Id, strconv.FormatInt(userAddressID, 10), "1")
```
rmulhol: Definitely not a blocker but could be cool to introduce a `AssertMappingWithAddress` helper so that we don't need to remember to always include a separate expectation for the `AddressID` if we're doing this more often

rmulhol (approved):
LGTM! A few questions, nothing blocking

transformers/storage/cat/repository.go
rmulhol: fwiw we already do what I'm suggesting here: https://github.com/makerdao/vdb-mcd-transformers/blob/staging/transformers/storage/median/repository.go#L292

data_generator/data_generator.go
paytonrules: Oddly no, but I guess since it's test data.

transformers/storage/cat/repository.go
paytonrules: Oh bite me linter. why can no haz capitals?

transformers/shared/utilities.go
paytonrules: 👍  - I'm also assuming that eventually these two functions will coalesce into one. Is that fair?

paytonrules: Did not know about that but it makes a lot of sense, since this is in a transaction.

transformers/storage/cat/repository.go
paytonrules: That makes sense to me - originally I had this passing around the ID but I didn't care for that.

transformers/component_tests/cat/configured_transformer_test.go
paytonrules: Heh - I flip/flopped on that. I can't remember why I stayed with this version.

transformers/component_tests/cat/configured_transformer_test.go
paytonrules: 👍

elizabethengelman (approved):

transformers/shared/utilities.go
paytonrules: I ended up just passing the addressID here, because we didn't use this method as often as I originally thought. That meant it used the cached version in the cat storage repository.

PR #270, opened 2020-08-13
Add eth_node_id to storage_diffs_table

paytonrules:
Synchronize with vulcanizedb. 

This meant pointing to the staging version of vulcanizedb so that PersistedDiff would have the EthNodeID.


elizabethengelman (approved):
🎉

PR #272, opened 2020-08-18
(VDB-1082) bigint primary keys for big tables

rmulhol:


db/migrations/20180815031512_create_tend.sql
```diff
@@ -2,13 +2,13 @@
...
-    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
     log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
     bid_id     NUMERIC NOT NULL,
     lot        NUMERIC,
     bid        NUMERIC,
-    address_id INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
-    msg_sender INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    address_id BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    msg_sender BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
```
elizabethengelman, line 11: Do you think it's worth updating the header id to be a `BIGSERIAL` (on the header table), and then a `BIGINT` here?

elizabethengelman (approved):
👍

db/migrations/20180815031512_create_tend.sql
elizabethengelman: Also, now just realizing that this would be more a change for the vdb repo.

rmulhol: I'm thinking about skipping that one just because I think we'd need to be syncing multiple networks to have any chance of exhausting the range of ints available, and I feel like supporting that would require more deliberate thought/work elsewhere 🤔

PR #276, opened 2020-08-21
(VDB-1424) Remove hashed addresses

rmulhol:
- companion PR for changes to vulcanizedb: https://github.com/makerdao/vulcanizedb/pull/121
- don't need hashed addresses with Geth version that emits raw address


transformers/component_tests/cat/configured_transformer_test.go
```diff
@@ -39,17 +38,16 @@ import (
...
 	)
 
 	BeforeEach(func() {
-		storageKeysLookup := storage.NewKeysLookup(cat.NewKeysLoader(&mcdStorage.MakerStorageRepository{}, contractAddress))
-		repository := cat.StorageRepository{ContractAddress: contractAddress}
+		storageKeysLookup := storage.NewKeysLookup(cat.NewKeysLoader(&mcdStorage.MakerStorageRepository{}, contractAddress.Hex()))
+		repository := cat.StorageRepository{ContractAddress: contractAddress.Hex()}
 		transformer = storage.Transformer{
-			Address:           common.HexToAddress(contractAddress),
+			Address:           common.HexToAddress(contractAddress.Hex()),
```
yaoandrew: What type is `contractAddress`? I'm not sure I follow all the converting here. Isn't it already an `address`?

yaoandrew (approved):
Looks good to me 🚢 . Just had a non-blocking question on some type conversions.

transformers/component_tests/cat/configured_transformer_test.go
rmulhol: yes, nice catch!

PR #280, opened 2020-08-31
Add versioning to cat contracts [staging]

elizabethengelman:
_cut from `staging`_

I think that this is a slightly different situation than the last time we added versions to our contracts (auction contracts) since at that time the ABIs were unchanged (I think?), or at least the interface of the contracts we cared about didn't change. Whereas now, we need to account for a new cat ABI. But would love another set of eyes on that.


transformers/shared/constants/method.go
```diff
@@ -83,19 +84,19 @@ func VatABI() string  { return getContractABI("MCD_VAT") }
 func VowABI() string  { return getContractABI("MCD_VOW") }
 
 func auctionFileMethod() string { return getSolidityFunctionSignature(FlipABI(), "file") }
-func biteMethod() string        { return getSolidityFunctionSignature(CatABI(), "Bite") }
+func biteMethod() string        { return getSolidityFunctionSignature(Cat100ABI(), "Bite") }
```
elizabethengelman: I'm kind of wondering if we want to use `Cat110ABI` instead here, just to rely on the most recent contract? They should be the same, so it probably doesn't matter too much 🤔

paytonrules: Why wouldn't we use the latest? Strikes me as odd to use the old one, but I am a Ether-dumb-dumb.

paytonrules (approved):
Seems straightforward, mostly follow the compiler stuff, but I don't see where the new version of the contract is used.

It all looks right but I feel like I'm missing something.

elizabethengelman:
Yeah, it's not being used in this PR - normally I would have done it as part of #284 but broke out adding the new contract into a separate PR so we could use it for the other transformers that depended on it.

transformers/shared/constants/method.go
elizabethengelman: I think using the latest one makes sense - for many of the cat methods, the ABI didn't change, so it doesn't matter too much. But might as well use the most recent one. I'll push up a commit shortly!

PR #282, opened 2020-08-31
VDB-1599 cat file box [beta]

elizabethengelman:


transformers/integration_tests/cat_file_test.go
```diff
@@ -36,168 +37,212 @@ import (
...
-		catFileConfig.Topic = constants.CatFileVowSignature()
-		catFileConfig.StartingBlockNumber = vowBlockNumber
-		catFileConfig.EndingBlockNumber = vowBlockNumber
-
-		initializer := event.ConfiguredTransformer{
-			Config:      catFileConfig,
-			Transformer: vow.Transformer{},
+	Describe("1.1.0 Cat Contract", func() {
+		var catFileConfig = event.TransformerConfig{
+			ContractAddresses: []string{test_data.Cat110Address()},
```
elizabethengelman, line 203: I think this is a good example of how this may be confusing/unclear to future developers - like for testing `cat_file_box` we _have_ to use the 1.1.0 version. But for the other events, we could use either contract, and I'm not sure that this is really clear anywhere in the code. I don't have a good idea on how to clarify this some, but just wanted to point it out.

elizabethengelman: Also, just thinking that maybe it would be good to include tests for the other events, but for 1.1.0. 🤔

transformers/events/cat_file/box/transformer_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	//"github.com/ethereum/go-ethereum/core"
+	"github.com/makerdao/vdb-mcd-transformers/test_config"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/events/cat_file/box"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+
+	//"github.com/makerdao/vdb-mcd-transformers/transformers/events/cat_file/box"
```
yaoandrew: Do we need these commented out imports?

transformers/integration_tests/cat_file_test.go
yaoandrew: Totally. I'm confused 🤪  just thinking about this. So for something like `cat_file_vow` that exists in both versions and is unchanged, we won't duplicate those tests for 1.1.0? And since `cat_file_box` was introduced in 1.1.0 we'll only have tests for it there?  We should probably see how this plays out as we make changes.

transformers/shared/constants/signature_test.go
```diff
@@ -42,6 +42,10 @@ var _ = Describe("Signature constants", func() {
 		Expect(CatFileVowSignature()).To(Equal("0xd4e8be8300000000000000000000000000000000000000000000000000000000"))
 	})
 
+	It("generates cat file box signature", func() {
```
yaoandrew: Should this be in a different place alphabetically? This is probably super OCD and obviously non-blocking 😅

yaoandrew (dismissed):
LGTM all non-blocking comments :shipit:  🚀

transformers/events/cat_file/box/transformer_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	BeforeEach(func() {
+		test_config.CleanTestDB(db)
+	})
+
+	It("converts a box log to a model", func() {
+		models, toModelsErr := transformer.ToModels(constants.Cat110ABI(), []core.EventLog{test_data.CatFileBoxEventLog}, db)
+		Expect(toModelsErr).NotTo(HaveOccurred())
+
+		contractAddressID, msgSenderErr := shared.GetOrCreateAddress(test_data.Cat110Address(), db)
```
yaoandrew, line 31: :100: I like the way this reads much better. I was putting something like this in my tests - `shared.GetOrCreateAddress(test_data.CatFileBoxEventLog.Log.Address.Hex(), db)`

transformers/events/cat_file/box/transformer_test.go
elizabethengelman: nice catch! 🙈

transformers/integration_tests/cat_file_test.go
elizabethengelman: Yeah, we'd end up duplicating tests for 1.1.0... which isn't ideal either. I'm good with holding off with making that change for now, and seeing how the rest of these changes pan out before making a decision.

yaoandrew (dismissed):

plugins/backfill/transformerExporter.go
```diff
@@ -13,7 +14,9 @@ type exporter string
 var Exporter exporter
 
 func (e exporter) Export() ([]event.TransformerInitializer, []storage.TransformerInitializer, []interface1.ContractTransformerInitializer) {
-	return []event.TransformerInitializer{},
+	return []event.TransformerInitializer{
+			cat_file_box.EventTransformerInitializer,
```
rmulhol: I think this needs to be removed from this file if we don't want to kick off a backfill

transformers/events/cat_file/box/transformer_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/makerdao/vdb-mcd-transformers/test_config"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/events/cat_file/box"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
rmulhol, line 14: super nbd but I don't think these newlines are necessary in the imports

rmulhol (approved):
:shipit:

PR #283, opened 2020-09-01
(VDB-1605) cat 1.1.0 storage

rmulhol:


transformers/component_tests/queries/all_ilks_query_test.go
```diff
@@ -47,7 +47,7 @@ var _ = Describe("All Ilks query", func() {
 				headerOne.Timestamp, batIlkValuesTwo)
 
 			var actualIlks []ilkHelpers.IlkSnapshot
-			err := db.Select(&actualIlks, `SELECT ilk_identifier, rate, art, spot, line, dust, chop, lump, flip, rho, duty, pip, mat, created, updated FROM api.all_ilks()`)
+			err := db.Select(&actualIlks, `SELECT ilk_identifier, rate, art, spot, line, dust, chop, lump, flip, rho, duty, pip, mat, dunk, created, updated FROM api.all_ilks()`)
```
elizabethengelman, line 50: Making sure I understand how we're handling this - we're expecting that we'd see `lump` be 0 for the new contract, and `dunk` be 0 for the old contract. Is that right?

transformers/storage/cat/repository_test.go
```diff
@@ -34,7 +34,7 @@ var _ = Describe("Cat storage repository", func() {
 
 	BeforeEach(func() {
 		test_config.CleanTestDB(db)
-		repo = cat.StorageRepository{ContractAddress: test_data.CatAddress()}
+		repo = cat.StorageRepository{ContractAddress: test_data.CatV100Address()}
```
elizabethengelman: Super small thing, but I wonder if we should use CatV110Address for these new storage slot tests, since they are only in the new contract? Not that it probably matters for these specific tests, but it may be helpful for documentation purposes?

elizabethengelman (approved):

transformers/component_tests/queries/all_ilks_query_test.go
rmulhol: I think we'd expect those values to be `null` on the contract where they aren't present, although it looks a little different on the actual `ilk_snapshot` table depending on the history of the ilk - e.g. if a `lump` was assigned before a `dunk`, then `lump` = `n` and `dunk` = `null`. Once `dunk` is assigned, `lump` = `n` and `dunk` = `m`. If `lump` was never assigned before `dunk` (for a new ilk), then `lump` = `null` and `dunk` = `n`, etc

rmulhol: also, if we insert a fake diff to wipe the `lump`, then we'd expect `lump` to go to whatever it's assigned to in the fake diff

transformers/storage/cat/repository_test.go
rmulhol: As it stands, this repository includes fields from both contracts (e.g. both `lump` and `dunk`). Happy to make the change, but I guess neither contract has all of the values that are included in these tests 🤔

PR #284, opened 2020-09-01
VDB-1599 cat file box [staging]

elizabethengelman:


plugins/backfill/storage/transformerExporter.go
```diff
@@ -35,28 +14,6 @@ var Exporter exporter
...
-			flip_wbtc_a_v1_0_9.StorageTransformerInitializer,
-			flip_zrx_a_v1_0_8.StorageTransformerInitializer,
-			flip_zrx_a_v1_0_9.StorageTransformerInitializer,
-			flop_v1_0_9.StorageTransformerInitializer,
-			median_knc.StorageTransformerInitializer,
-			median_mana.StorageTransformerInitializer,
-			median_wbtc.StorageTransformerInitializer,
-			median_zrx.StorageTransformerInitializer,
-		},
+		[]storage.TransformerInitializer{},
```
paytonrules, line 17: Did you mean to empty this out?

transformers/events/cat_file/box/transformer_test.go
```diff
@@ -0,0 +1,51 @@
+package box_test
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	//"github.com/ethereum/go-ethereum/core"
```
paytonrules, line 6: Commented imports.

transformers/shared/constants/method.go
```diff
@@ -83,19 +84,22 @@ func VatABI() string  { return getContractABI("MCD_VAT") }
...
 
 func auctionFileMethod() string { return getSolidityFunctionSignature(FlipABI(), "file") }
-func biteMethod() string        { return getSolidityFunctionSignature(CatABI(), "Bite") }
+func biteMethod() string        { return getSolidityFunctionSignature(Cat110ABI(), "Bite") }
+func catFileBoxMethod() string {
+	return getOverloadedFunctionSignature(Cat110ABI(), "file", []string{"bytes32", "uint256"})
+}
 func catFileChopLumpMethod() string {
-	return getOverloadedFunctionSignature(CatABI(), "file", []string{"bytes32", "bytes32", "uint256"})
+	return getOverloadedFunctionSignature(Cat100ABI(), "file", []string{"bytes32", "bytes32", "uint256"})
```
paytonrules, line 92: Just double checking that this is right - this is Cat100 not 110 because there is no chop lump in the new contract correct?

paytonrules (approved):
A couple questions that probably reflect more on my ignorance than the code, and you'll probably want to remove commented imports, otherwise 👍

plugins/backfill/storage/transformerExporter.go
elizabethengelman: I did - was thinking that it would be easier to do a backfill for all of the new cat transformers in one go, than having to change it for each new transformer.

transformers/events/cat_file/box/transformer_test.go
elizabethengelman: 🙈  nice catch

transformers/shared/constants/method.go
elizabethengelman: Yep, that's correct. Though, I think that Cat110 has a file method that has the same signature, so this method could probably be renamed to `catFileChopDunkMethod()` as part of the VDB-1632 work.

PR #285, opened 2020-09-01
VDB 1627 Only manage plugin migrations

paytonrules:
The transformer should only manage the plugin migrations, and is not responsible for the public schema. In addition to just pulling out the migrations:

- The Make tasks all pass the "-table" flag when migrating the schema.
- The docker files all use go run to ensure they are using the right version of goose
- A copy of the schema dump from vulcanizedb is used for testing. Updating vulcanize can be done with a handy make task.
- Compose now takes a schema flag.


Makefile
```diff
@@ -113,34 +107,41 @@ checkmigname:
...
 ## Apply all migrations not already run
 .PHONY: migrate
-migrate: $(GOOSE) checkdbvars
+migrate: checkdbvars
+	psql $(NAME) -c 'CREATE SCHEMA IF NOT EXISTS maker;'
 	cd db/migrations;\
-	  $(GOOSE) postgres "$(CONNECT_STRING)" up
-	pg_dump -O -s $(CONNECT_STRING) > db/schema.sql
+	  $(GOOSE) -table "maker.goose_db_version" postgres "$(CONNECT_STRING)" up
+	pg_dump -n 'maker' -n 'api' -O -s $(CONNECT_STRING) > db/schema.sql
```
rmulhol, line 127: Not a responsibility of this PR but I wonder if we should create a story to remove/rename the `api` schema - seems like that could get confusing with multiple plugins + every schema include in the actual API 🤔

dockerfiles/execute/run_migrations.sh
```diff
@@ -29,6 +29,8 @@ fi
 # Run the DB migrations
 echo "Connecting with: $VDB_PG_CONNECT"
 ./goose -dir db/migrations postgres "$VDB_PG_CONNECT" up
+psql $VDB_PG_CONNECT -c 'CREATE SCHEMA IF NOT EXISTS maker;'
+./goose -table maker.goose_db_version -dir db/vdb-mcd-transformers/migrations postgres "$VDB_PG_CONNECT" up
```
rmulhol, line 33: So just to make sure I understand - we need some vdb to image to run its migrations and setup the public schema before this fires, right? Since we're not loading the public schema? (Totally makes sense if so, since I think we prob don't want to be loading the public schema here, just thinking about order dependency in docker-compose etc)

dockerfiles/extract_diffs/Dockerfile
```diff
@@ -20,6 +15,8 @@ RUN git clone https://github.com/makerdao/vulcanizedb.git
 WORKDIR /go/src/github.com/makerdao/vulcanizedb
 RUN git checkout $VDB_VERSION
 RUN go build
+# Build migration tool
+RUN go install -tags='no_mysql no_sqlite3 no_mssql no_redshift' github.com/pressly/goose/cmd/goose
```
rmulhol: Just tagging here something we discussed on the VDB PR that we need to confirm this sticks to a version of goose that supports the `-table` flag (and doesn't accidentally upgrade to not supporting it if they remove/rename that in a subsequent release)

transformers/integration_tests/plugin_test.go
```diff
@@ -41,7 +41,8 @@ import (
 )
 
 var eventConfig = config.Plugin{
-	Home: "github.com/makerdao/vdb-mcd-transformers",
+	Home:   "github.com/makerdao/vulcanizedb",
```
rmulhol: This is interesting to me and something I probably missed on the VDB PR side - why is `Home` no longer the plugin repo? Should we omit if it's always going to be `github.com/makerdao/vulcanizedb`?

transformers/integration_tests/plugin_test.go
```diff
@@ -167,6 +176,7 @@ var _ = Describe("Plugin test", func() {
 			goPath, soPath, pathErr = eventConfig.GetPluginPaths()
 			Expect(pathErr).ToNot(HaveOccurred())
 			g, initErr = p2.NewGenerator(eventConfig, dbConfig)
+			g.SetMigrationManager(&StubMigrationManager{})
```
rmulhol, line 179: Why is this necessary? Was it running migrations before?

rmulhol (commented):
A few questions but in general this looks like a super awesome and elegant way to isolate the plugin schema 🥇 🚀

dockerfiles/execute/run_migrations.sh
paytonrules: Line 31 above loads the public schema, just like before. If we want to take it out I'd be okay with that - when we run the docker compose setup there are two containers trying to migrate public. One "wins" and it works itself out with retries, but I don't like it.

transformers/integration_tests/plugin_test.go
paytonrules: It would try to run migrations - but would determine they were finished. This doesn't work now since in test the public schema isn't technically migrated, just loaded.

 I tried for way too long to get this to play nice via various tricks, before I just realized I didn't want this in tests anyway.

transformers/integration_tests/plugin_test.go
paytonrules: Home is how the plugin found its public migration schema. Maybe that's a bad name too, because it didn't seem to use it for anything else.

So this worked before because of the duplicated migrations. That said this change might not be necessary once I stubbed out the migrations in tests.

dockerfiles/execute/run_migrations.sh
rmulhol: oh snap I didn't realize there were two `up` statements here 👍

transformers/integration_tests/plugin_test.go
rmulhol: I mean, if we don't need it, ... ✂️

transformers/integration_tests/plugin_test.go
rmulhol: yeah I mean nice catch, tests running migrations is ... :donotwant:

Makefile
paytonrules: I can add a story to Jira if you want. API definitely seems confused.

dockerfiles/extract_diffs/Dockerfile
paytonrules: I went ahead and changed the docker containers to all explicitly specify their versions, just to be safe.

transformers/integration_tests/plugin_test.go
paytonrules: It's ✂️

Makefile
elizabethengelman: 👍 really like the idea of updating the `api` schema at some point

dockerfiles/backfill_events/Dockerfile
```diff
@@ -4,9 +4,6 @@ RUN apk --update --no-cache add make git g++ linux-headers
 
 # Build migration tool
 WORKDIR /go
-RUN GO111MODULE=auto go get -u -d github.com/pressly/goose/cmd/goose
-WORKDIR /go/src/github.com/pressly/goose/cmd/goose
-RUN GO111MODULE=auto go build -a -ldflags '-s' -tags='no_mysql no_sqlite' -o goose
```
elizabethengelman, line 7: Just to make sure I'm thinking about this correctly - we don't need goose for this command since it's not going to be doing any migration - is that right?

I wonder if we can remove the `COPY --from=builder /go/src/github.com/makerdao/vdb-mcd-transformers/db/migrations/* db/migrations/` line below as well?

dockerfiles/extract_diffs/Dockerfile
```diff
@@ -31,11 +28,10 @@ RUN apk update \
         && apk upgrade \
         && apk add --no-cache \
         ca-certificates \
-	    bash \
+        bash \
+        postgresql-client \
         && update-ca-certificates 2>/dev/null || true
 
-ARG CONFIG_FILE=environments/extractDiffs.toml
```
elizabethengelman, line 37: Curious why we're removing this config file arg?

elizabethengelman (commented):
looking good, just a couple of questions 👍

dockerfiles/backfill_events/Dockerfile
paytonrules: Sure can, nice catch!

dockerfiles/extract_diffs/Dockerfile
paytonrules: We aren't - it's also on line 42. This one is a duplicate.

rmulhol (approved):
approving since the deadline passed 🕐

PR #288, opened 2020-09-01
VDB-1598: cat claw staging [staging]

elizabethengelman:


db/migrations/20200901135641_create_cat_claw.sql
```diff
@@ -0,0 +1,21 @@
+-- +goose Up
+CREATE TABLE maker.cat_claw
+(
+    id        SERIAL PRIMARY KEY,
+    header_id INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    address_id BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    log_id    BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    msg_sender BIGINT NOT NULL REFERENCES addresses (id) ON DELETE CASCADE,
+    rad      NUMERIC,
```
paytonrules, line 9: Why is this rad and not litter?

transformers/integration_tests/cat_claw_test.go
```diff
@@ -0,0 +1,77 @@
...
+		addressID, addressErr := shared.GetOrCreateAddress(test_data.Cat110Address(), db)
+		Expect(addressErr).NotTo(HaveOccurred())
+
+		msgSender := "0xF32836B9E1f47a0515c6Ec431592D5EbC276407f"
+		msgSenderID, msgSenderErr := shared.GetOrCreateAddress(msgSender, db)
+		Expect(msgSenderErr).NotTo(HaveOccurred())
+
+		Expect(dbResult.AddressID).To(Equal(addressID))
+		Expect(dbResult.MsgSender).To(Equal(msgSenderID))
+		Expect(dbResult.Rad).To(Equal("164878299999999999999707170900220955677367419868"))
```
paytonrules, line 69: Not a big enough number.

paytonrules (approved):
I don't understand why it's rad and not litter, but I think that's my confusion not yours.

👍

db/migrations/20200901135641_create_cat_claw.sql
elizabethengelman: It's `rad` since that is what the argument for the `claw` method: https://github.com/makerdao/dss/blob/master/src/cat.sol#L178

PR #293, opened 2020-09-03
Remove Oasis

rmulhol:
- remove Oasis tables from schema
- remove Oasis transformer code + config
- remove Oasis test data + integration tests

WIP until we're confident about running Oasis stuff from a separate repo


transformers/shared/constants/keys.go
```diff
@@ -45,43 +45,28 @@ const (
 	ArtColumn       event.ColumnName = "art"
 	BidColumn       event.ColumnName = "bid"
 	BidIDColumn     event.ColumnName = "bid_id"
-	BuyAmtColumn    event.ColumnName = "buy_amt"
-	BuyGemColumn    event.ColumnName = "buy_gem"
 	CdpColumn       event.ColumnName = "cdp"
 	DartColumn      event.ColumnName = "dart"
 	DataColumn      event.ColumnName = "data"
 	DinkColumn      event.ColumnName = "dink"
-	DustColumn      event.ColumnName = "dust"
```
yaoandrew, line 54: Is this only used in the Oasis transformers? I'm just thinking there's a vat_file `what = "dust"`, but maybe there aren't any `event.ColumnName ` entries for dust?

yaoandrew (approved):

transformers/shared/constants/keys.go
rmulhol: I believe this file should only need to contain columns that are used in the oasis transformers (and their schema). I would have thought that we could also get rid of `dink`/`dart` and some of the other column names in this file

rmulhol: Whoops I totally misunderstood this - we need `dink`/`dart` because this is MCD not Oasis 🤦 

I'm assuming the `dust` column wasn't actually used anywhere (as you suggested, since it's just a `what` value that gets parsed as a string)

rmulhol:
Might need to coordinate this with tech ops but I wonder if we want to make sure any names like `execute` or `vdb-execute` in docker stuff are renamed to `vdb-mcd-execute` or similar, in light of the additional `execute` image (may also apply to `extract-diffs`?)

paytonrules:
Mentioned this in the Oasis PR but let's get the docker images renamed as part of the deployment stories.

PR #294, opened 2020-09-04
Vdb 1632 [beta] cat file dunk

yaoandrew:


transformers/integration_tests/cat_file_test.go
```diff
@@ -203,6 +204,46 @@ var _ = Describe("Cat File transformer", func() {
 			ContractAddresses: []string{test_data.Cat110Address()},
 			ContractAbi:       constants.Cat110ABI(),
 		}
```
yaoandrew, line 206: Not sure what we want to do here. Do we leave out a test for `chop_lump` since its unchanged between versions?

transformers/integration_tests/cat_file_test.go
```diff
@@ -203,6 +204,46 @@ var _ = Describe("Cat File transformer", func() {
...
+			msgSenderID, err := shared.GetOrCreateAddress("0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB", db)
+			Expect(err).NotTo(HaveOccurred())
+			err = db.Get(&dbResult, `SELECT address_id, msg_sender, what, data FROM maker.cat_file_chop_dunk`)
+			Expect(err).NotTo(HaveOccurred())
+			addressID, err := shared.GetOrCreateAddress(test_data.Cat110Address(), db)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(dbResult.AddressID).To(Equal(addressID))
+			Expect(dbResult.MsgSenderID).To(Equal(msgSenderID))
+			Expect(dbResult.What).Should(Or(Equal("dunk"), Equal("chop")))
```
yaoandrew, line 243: This is the only block with `dunk` events, but it also has `chop` events (same signature). Not sure if there's a better assertion here.

rmulhol (changes requested):
I think we might not want a new transformer if its got the same address + topic0 as `cat_file_chop_lump` - seems like the existing transformer/table will capture these events as is.

Could see a case for renaming the transformer/table + adding an integration test for dunk events, but would like to think through anything beyond that a little bit further - at least for beta

transformers/integration_tests/cat_file_test.go
rmulhol: I would use a `db.Select` to get both results and assert that `len(dbResults` == 2 - so that you know we're not just matching the chop event

rmulhol:
Closing in favor of separate approach without new transformer (since address and signature match existing `cat_file_chop_lump` transformer)

PR #295, opened 2020-09-04
Added a time_frob_totals query

hexonaut:
Follows the same format as the other time series queries.

There is some weird indexing behaviour with the headers table. You may wonder why I'm having to precompute and do a weird `AVG(id) + (COUNT(id)/2.0) - 0.5` to calculate the max id. If you replace that with MAX(id) then it will try to order the table and slow down the query massively. I get my solution is a little hackish, but it works until a better solution comes along.

Note: Still need to add tests.


hexonaut:
Update: Managed to find a much better solution then the AVG + COUNT workaround.

hexonaut:
I've added a max bucket limit of 100 to all queries. The assert returns a proper error state with a clear message to the user:

```
{
  "message": "Please limit requests to at most 100 buckets.",
  "locations": [
    {
      "line": 32,
      "column": 3
    }
  ],
  "path": [
    "timeFrobTotals"
  ]
}
```

hexonaut:
Moved to here: https://github.com/makerdao/vdb-mcd-transformers/pull/365

PR #297, opened 2020-09-08
Update cat file transformer with chop lump dunk [beta edition]

yaoandrew:
This is a new PR which modifies the existing cat file chop lump transformer to save `cat1.1.0` dunk events.

### DIFF
--- a/db/migrations/20180912171047_create_cat_file.sql
+++ b/db/migrations/20180912171047_create_cat_file.sql
@@ -1,23 +1,21 @@
 -- +goose Up
-CREATE TABLE maker.cat_file_chop_lump_dunk
+CREATE TABLE maker.cat_file_chop_lump
 (
-    id         SERIAL PRIMARY KEY,
-    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
-    address_id BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
-    msg_sender BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
-    log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
-    ilk_id     INTEGER NOT NULL REFERENCES maker.ilks (id) ON DELETE CASCADE,
-    what       TEXT,
-    data       NUMERIC,
+    id        SERIAL PRIMARY KEY,
+    header_id INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    log_id    BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    ilk_id    INTEGER NOT NULL REFERENCES maker.ilks (id) ON DELETE CASCADE,
+    what      TEXT,
+    data      NUMERIC,
     UNIQUE (header_id, log_id)
 );
 
-CREATE INDEX cat_file_chop_lump_dunk_header_index
-    ON maker.cat_file_chop_lump_dunk (header_id);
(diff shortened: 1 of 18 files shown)


db/migrations/20180912171047_create_cat_file.sql
```diff
@@ -1,21 +1,23 @@
 -- +goose Up
-CREATE TABLE maker.cat_file_chop_lump
+CREATE TABLE maker.cat_file_chop_lump_dunk
```
rmulhol: unfortunately, we can't modify an earlier migration if we're looking to have this applied to `beta`. This should work for `staging` though 👍

rmulhol (changes requested):
Needs to be addressed via backwards-compatible changes (new migration and/or leaving the table name unchanged [leaving the table unchanged actually isn't that bad of an option since we can always alias in postgraphile])

transformers/component_tests/queries/all_ilk_file_events_query_test.go
```diff
@@ -49,17 +50,24 @@ var _ = Describe("Ilk File Events Query", func() {
 		timestampOne = int(rand.Int31())
 		headerOne = createHeader(blockOne, timestampOne, headerRepo)
 		logOneId = test_data.CreateTestLog(headerOne.Id, db).ID
+		msgSender = test_data.Cat110Address()
```
rmulhol: I wonder if we might want to use a different (potentially random) address here - the contract itself is unlikely to be the `msg.sender`, and I could see some tricky-to-debug test situations if the contract address and message sender IDs are the same

db/migrations/20180912171047_create_cat_file.sql
yaoandrew: Oh cool. I didn't realize we could alias it in postgraphile. That sounds like a better bet.

transformers/component_tests/queries/all_ilk_file_events_query_test.go
yaoandrew: Yeah.. makes sense. I think I spaced out on assigning the contract address as the msgSender.

transformers/events/cat_file/chop_lump_dunk/transformer_test.go
```diff
@@ -44,8 +45,7 @@ var _ = Describe("Cat file chop lump transformer", func() {
 			models, err := transformer.ToModels(constants.Cat100ABI(), []core.EventLog{test_data.CatFileChopEventLog}, db)
 			Expect(err).NotTo(HaveOccurred())
 
-			var ilkID int64
-			ilkErr := db.Get(&ilkID, `SELECT id FROM maker.ilks where ilk = $1`, test_data.CatFileChopEventLog.Log.Topics[2].Hex())
+			ilkID, ilkErr := shared.GetOrCreateIlk(test_data.CatFileChopEventLog.Log.Topics[2].Hex(), db)
```
elizabethengelman, line 48: 👏

elizabethengelman (approved):

transformers/shared/constants/method.go
```diff
@@ -102,6 +102,9 @@ func catClawMethod() string     { return getSolidityFunctionSignature(Cat110ABI(
 func catFileBoxMethod() string {
 	return getOverloadedFunctionSignature(Cat110ABI(), "file", []string{"bytes32", "uint256"})
 }
+func catFileChopDunkMethod() string {
```
rmulhol: Seems like we could just rename the `catFileChopLumpMethod()` to `catFileChopLumpDunkMethod()`?

rmulhol (approved):
Would prefer to remove the `CatFileChopDunkSignature()` method + private functions it calls (and get rid of the test) since I think we can use/rename `CatFileChopLumpSignature()` - otherwise LGTM 👍

PR #298, opened 2020-09-09
(VDB-1606) Add Flip v1.1.0 storage transformers

rmulhol:
- include cat field

### DIFF
--- a/environments/docker.toml
+++ b/environments/docker.toml
@@ -11,22 +11,31 @@
         "flap_v1_0_9",
         "flip_bat_a_v1_0_0",
         "flip_bat_a_v1_0_9",
+        "flip_bat_a_v1_1_0",
         "flip_eth_a_v1_0_0",
         "flip_eth_a_v1_0_9",
+        "flip_eth_a_v1_1_0",
         "flip_knc_a_v1_0_8",
         "flip_knc_a_v1_0_9",
+        "flip_knc_a_v1_1_0",
         "flip_mana_a_v1_0_9",
+        "flip_mana_a_v1_1_0",
         "flip_sai_v1_0_0",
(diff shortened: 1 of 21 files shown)


transformers/storage/bid_storage.go
```diff
@@ -17,16 +17,18 @@
...
-	Ilk    = "ilk"
-	Gem    = "gem"
-	Beg    = "beg"
-	Pad    = "pad"
-	Ttl    = "ttl"
-	Tau    = "tau"
-	Kicks  = "kicks"
-	Live   = "live"
-	Vow    = "vow"
+	Beg   = "beg"
```
yaoandrew, line 20: 🥇

yaoandrew (approved):
LGTM 🚀

environments/docker.toml
```diff
@@ -11,22 +11,31 @@
...
+        "flip_eth_a_v1_1_0",
         "flip_knc_a_v1_0_8",
         "flip_knc_a_v1_0_9",
+        "flip_knc_a_v1_1_0",
         "flip_mana_a_v1_0_9",
+        "flip_mana_a_v1_1_0",
         "flip_sai_v1_0_0",
         "flip_tusd_a_v1_0_7",
         "flip_tusd_a_v1_0_9",
+        "flip_tusd_a_V1_1_0",
```
elizabethengelman: super small, but I think we want this `V1` to be `v1` so that it matches with the exporter?

elizabethengelman (approved):
Looks good - just that one comment about the flip_tusd transformer name in docker.toml.

environments/docker.toml
rmulhol: Nice catch! 👍

PR #300, opened 2020-09-10
(VDB-1594) Add 1.1.0 addresses [staging]

rmulhol:


transformers/shared/constants/method.go
```diff
@@ -48,6 +48,19 @@ func FlipABI() string {
 		"MCD_FLIP_ZRX_A_1.0.9",
 	})
 }
+func FlipV110ABI() string {
+	return GetABIFromContractsWithMatchingABI([]string{
+		"MCD_FLIP_BAT_A_1.1.0",
+		"MCD_FLIP_ETH_A_1.1.0",
+		"MCD_FLIP_KNC_A_1.1.0",
+		"MCD_FLIP_MANA_A_1.1.0",
+		"MCD_FLIP_TUSD_A_1.1.0",
```
yaoandrew, line 57: Do we need USDT and PAXUSD added here or is that in another PR?

yaoandrew (approved):
Just had a question on the newer collateral additions. Otherwise LGTM. 🚀

environments/docker.toml
```diff
@@ -336,11 +336,13 @@
         repository = "github.com/makerdao/vdb-mcd-transformers"
```
elizabethengelman, line 406: Do we also need to add these new contract versions to the storage transformers list as well?

elizabethengelman: Also, I think we'd need new storage transformer initializers for those as well.

transformers/shared/constants/method.go
elizabethengelman: I am adding USDT and PAXUSD and #316 😄

environments/docker.toml
rmulhol: will be included when #302 and #303 are merged 👍

PR #301, opened 2020-09-11
Vdb 1597 flip file cat [beta]

yaoandrew:
This adds a new transformer for the `flip` contract to track `file_cat `events.


environments/docker.toml
```diff
@@ -444,6 +445,15 @@
         migrations = "db/migrations"
         contracts = ["MCD_FLAP_1.0.0", "MCD_FLAP_1.0.9"]
         rank = "0"
+    [exporter.flip_file_cat]
+        path = "transformers/events/flip_file/cat/initializer"
+        type = "eth_event"
+        repository = "github.com/makerdao/vdb-mcd-transformers"
+        migrations = "db/migrations"
+        contracts = ["MCD_FLIP_BAT_A_1.1.0", "MCD_FLIP_ETH_A_1.1.0", "MCD_FLIP_KNC_A_1.1.0", "MCD_FLIP_MANA_A_1.1.0",
```
rmulhol: Popping in to note that this might need to be updated if merged after #299

transformers/events/flip_file/cat/transformer.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	"github.com/makerdao/vulcanizedb/pkg/datastore/postgres"
+)
+
+type Transformer struct{}
+
+func (Transformer) ToModels(_ string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
+	var results []event.InsertionModel
+	for _, log := range logs {
+		err := shared.VerifyLog(log.Log, shared.FourTopicsRequired, shared.LogDataRequired)
```
rmulhol: Why is `LogDataRequired`? Looks like we're getting all info from the topics

transformers/events/flip_file/cat/transformer.go
```diff
@@ -0,0 +1,59 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		what := shared.DecodeHexToText(log.Log.Topics[2].Hex())
+		data := shared.ConvertUint256HexToBigInt(log.Log.Topics[3].Hex())
```
rmulhol: I think we want to decode the `data` as an address?

transformers/test_data/flip_file.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/makerdao/vulcanizedb/pkg/fakes"
+)
+
+var rawFlipFileCatLog = types.Log{
+	Address: common.HexToAddress(FlipEthV110Address()),
+	Topics: []common.Hash{
+		common.HexToHash(constants.FlipFileCatSignature()),
+		common.HexToHash("0x000000000000000000000000be8e3e3618f7474f8cb1d074a26affef007e98fb"),
+		common.HexToHash("0x6361740000000000000000000000000000000000000000000000000000000000"),
+		common.HexToHash("0x0000000000000000000000000000000000000000000000000e4b4b8af6a70000"),
```
rmulhol: If we're grabbing test data from another event, `cat_file_vow` might be a good choice (same signature, topics[3] is an address)

rmulhol (changes requested):
Need to decode the `data` payload as an `address` instead of a big int

rmulhol (approved):

transformers/events/flip_file/cat/transformer_test.go
```diff
@@ -0,0 +1,58 @@
...
+		Expect(addressErr).NotTo(HaveOccurred())
+
+		msgSenderID, msgSenderErr := shared.GetOrCreateAddress("0xbe8e3e3618f7474f8cb1d074a26affef007e98fb", db)
+		Expect(msgSenderErr).NotTo(HaveOccurred())
+
+		expectedFlipFileCat := test_data.FlipFileCatModel()
+		expectedFlipFileCat.ColumnValues[event.AddressFK] = addressID
+		expectedFlipFileCat.ColumnValues[constants.MsgSenderColumn] = msgSenderID
+		expectedFlipFileCat.ColumnValues[constants.WhatColumn] = "cat"
+		expectedFlipFileCat.ColumnValues[constants.DataColumn] = "0xA950524441892A31ebddF91d3cEEFa04Bf454466"
```
rmulhol: Super minor but you could probably assign the `What` and `Data` values on the `FlipFileCatModel()` to avoid needing to re-assign them here (as long as they're not referencing a FK that changes between test runs)

transformers/test_data/config_values.go
```diff
@@ -40,6 +40,26 @@ func FlipAddresses() []string {
...
+	flipAddresses := constants.GetContractAddresses([]string{
+		"MCD_FLIP_BAT_A_1.1.0",
+		"MCD_FLIP_ETH_A_1.1.0",
+		"MCD_FLIP_KNC_A_1.1.0",
+		"MCD_FLIP_MANA_A_1.1.0",
+		"MCD_FLIP_TUSD_A_1.1.0",
+		"MCD_FLIP_USDC_A_1.1.0",
+		"MCD_FLIP_USDC_B_1.1.0",
+		"MCD_FLIP_WBTC_A_1.1.0",
+		"MCD_FLIP_ZRX_A_1.1.0",
```
rmulhol, line 54: I think we can probably shore this up after the backfill process is running, but in theory this should include the PAXUSD + USDT addresses

yaoandrew: ugg.. forgot about this file.

PR #304, opened 2020-09-14
Adds dunk to cat_file event transformer [staging]

yaoandrew:


transformers/integration_tests/cat_file_test.go
```diff
@@ -266,6 +266,48 @@ var _ = Describe("Cat File transformer", func() {
...
+			msgSenderID, err := shared.GetOrCreateAddress("0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB", db)
+			Expect(err).NotTo(HaveOccurred())
+			err = db.Select(&dbResults, `SELECT address_id, msg_sender, what, data FROM maker.cat_file_chop_lump_dunk`)
+			Expect(err).NotTo(HaveOccurred())
+			addressID, err := shared.GetOrCreateAddress(test_data.Cat110Address(), db)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(dbResults[0].AddressID).To(Equal(addressID))
+			Expect(dbResults[0].MsgSender).To(Equal(msgSenderID))
+			Expect(dbResults[0].What).Should(Or(Equal("dunk"), Equal("chop")))
```
rmulhol: Sorry this came up earlier and I think I provided a bad answer, but I think we might want to do a sort here - similarly to what we do for [`vat_move`](https://github.com/makerdao/vdb-mcd-transformers/blob/staging/transformers/integration_tests/vat_move_test.go#L72). Probably not blocking but would enable us to be sure we were asserting against a result where `What` == `dunk` (and specify a single `data` value)

rmulhol (approved):
Looks like this needs a rebase and I had one minor comment but overall LGTM 👍

PR #306, opened 2020-09-14
Vdb 1638 insert zero cat lump

elizabethengelman:


cmd/assignIlkLumpToZero.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var blockNumber int
+var assignIlkLumpToZeroCmd = &cobra.Command{
+	Use:   "assignIlkLumpToZero",
+	Short: "Create a diff with a zero ilk lump value.",
+	Long: `Inserts an artificial storage_diff with the storage_key for cat_ilk_lump, and a
+storage_value of zero for all ilks for the given block height. This storage_diff will then be picked up in the execute
```
rmulhol: Noting for posterity that we're looking at changing this to enable setting `lump` to zero at different block heights for different ilks

cmd/assignIlkLumpToZero.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+var assignIlkLumpToZeroCmd = &cobra.Command{
+	Use:   "assignIlkLumpToZero",
+	Short: "Create a diff with a zero ilk lump value.",
+	Long: `Inserts artificial cat_ilk_lump storage_diffs with storage_values of zero for the given ilks at the given block
+height. The storage_diffs will then be picked up in the execute process to be transformed into maker.cat_ilk_lump records.
+The block-number (b) and ilks (i) arguments are required.
+
+Note: the ilk argument(s) passed in need to be the hex representation of the ilk.
```
elizabethengelman, line 22: I wanted to call this out - we'll need to pass the ilks in as their hex representation, i.e. `0x4241542d41000000000000000000000000000000000000000000000000000000`, `0x4554482d41000000000000000000000000000000000000000000000000000000`. 

Though not ideal, I thought it was easier than converting the ilkIdentifiers (i.e. `ETH-A`, `BAT-A`) into these hex representations. Another idea I had was to look up the ilks from the db, based on their identifiers. Not sure which is easiest/makes the most sense but up for some ideas!

rmulhol: Totally on board with your reasoning here, especially since this is a write operation that only we will run (and it only needs to be ran a few times)

cmd/assignIlkLumpToZero.go
```diff
@@ -0,0 +1,120 @@
...
+		logrus.Infof("Successfully created a zero-value ilk lump diffs for block %d", blockNumber)
+		return
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(assignIlkLumpToZeroCmd)
+	assignIlkLumpToZeroCmd.Flags().IntVarP(&blockNumber, "block-number", "b", -1, "the block associated with artificial ilk lump diffs")
+	assignIlkLumpToZeroCmd.MarkFlagRequired("block-number")
+	assignIlkLumpToZeroCmd.Flags().StringSliceVarP(&ilks, "ilks", "i", []string{}, "the ilk to set lump to zero")
```
rmulhol, line 46: This is interesting - so for a string slice we don't need to pass `--ilks ilk1,ilk2,ilk3` but can do `--ilk ilk1 --ilk ilk2 --ilk ilk3`?

cmd/assignIlkLumpToZero.go
```diff
@@ -0,0 +1,120 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/storage/cat/v1_0_0"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
+	storage2 "github.com/makerdao/vulcanizedb/libraries/shared/storage"
```
rmulhol: super nbd but wondering if we might be able to remove the alias on this import?

cmd/assignIlkLumpToZero.go
```diff
@@ -0,0 +1,120 @@
...
+
+func NewZeroValueDiffGenerator(db postgres.DB) ZeroValueDiffGenerator {
+	return ZeroValueDiffGenerator{
+		HeaderRepository: repositories.NewHeaderRepository(&db),
+		DiffRepo:         storage2.NewDiffRepository(&db),
+	}
+}
+
+func (generator ZeroValueDiffGenerator) CreateZeroValueIlkLumpDiff(blockNumber int, ilks []string) error {
+	fmt.Println("ilks", ilks)
```
rmulhol: maybe worth removing and/or using logging tools with a more descriptive message?

cmd/assignIlkLumpToZero.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func zeroOutIlkLump() error {
+	blockChain := getBlockChain()
+	db := utils.LoadPostgres(databaseConfig, blockChain.Node())
+	generator := NewZeroValueDiffGenerator(db)
+	return generator.CreateZeroValueIlkLumpDiff(blockNumber, ilks)
+}
+
+type ZeroValueDiffGenerator struct {
```
rmulhol: Maybe worth pulling out this struct into a separate namespace outside of the `cmd` directory?

cmd/assignIlkLumpToZero_test.go
```diff
@@ -0,0 +1,87 @@
...
+				StorageKey:    common.HexToHash(ethALumpKey),
+				StorageValue:  common.Hash{},
+			}
+
+			batAIlkLumpRawDiff := types.RawDiff{
+				HashedAddress: common.HexToHash(test_data.Cat110Address()),
+				BlockHash:     fakes.FakeHash,
+				BlockHeight:   blockNumber,
+				StorageKey:    common.HexToHash(batALumpKey),
+				StorageValue:  common.Hash{},
```
rmulhol: would potentially like to see a test that an empty hash propagates to a zero value on the `cat_ilk_lump` table, though not sure the best way to do that without running execute. Maybe just for documentation's sake it would make sense to include a test that `HexToInt(common.Hash{}.Hex()) == 0`?

rmulhol (approved):
LGTM!

Don't think any of my comments are blocking, though would be curious to make sure we're certain that an empty hash is correctly transformed to 0 on the `cat_ilk_lump` table

cmd/assignIlkLumpToZero.go
elizabethengelman: Yikes, nice catch! 🤦‍♀️  On line 90 there's a log line with the ilks that have been passed in that is a better message.

cmd/assignIlkLumpToZero.go
elizabethengelman: Nice catch - I need to change the example to to `--ilks ilk1 --ilks ilk2`. But yes, as far as I can tell, cobra command handles slice by requiring that you pass each element in the slice as a separate flag. Kind of like how we were handling watched addresses in the old geth patch version.

cmd/assignIlkLumpToZero.go
elizabethengelman: Yeah, I was thinking that too, but I'm not sure where would be best to put it. Thinking that some of the checkMigrations code could be pulled into it's own file. Any ideas?

rmulhol: not merge blocking so I'd be fine to defer, but I think my default rn would be to create a new top level directory similar to `backfill/` or `data_generator/`

PR #307, opened 2020-09-15
Vdb 1597 flip file cat [staging]

yaoandrew:
This has been rebased with #300. Probably makes sense to wait until that's in staging and this has been rebased again.


transformers/events/flip_file/cat/transformer.go
```diff
@@ -0,0 +1,59 @@
...
+func (Transformer) ToModels(_ string, logs []core.EventLog, db *postgres.DB) ([]event.InsertionModel, error) {
+	var results []event.InsertionModel
+	for _, log := range logs {
+		err := shared.VerifyLog(log.Log, shared.FourTopicsRequired, shared.LogDataRequired)
+		if err != nil {
+			return nil, err
+		}
+
+		what := shared.DecodeHexToText(log.Log.Topics[2].Hex())
+		data := shared.ConvertUint256HexToBigInt(log.Log.Topics[3].Hex())
```
rmulhol: mentioned this on the beta PR but this should probably be converted to an address rather than an int

also looks like this PR has several extra commits that hopefully go away after #300 is merged

transformers/events/flip_file/cat/flip_file_cat_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package cat_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestBox(t *testing.T) {
```
elizabethengelman: Super small, but I think this should be `TestFileCat` instead of `TestBox` - not merge blocking!

db/migrations/20200915204154_create_flip_file_cat.sql
```diff
@@ -0,0 +1,23 @@
+-- +goose Up
+CREATE TABLE maker.flip_file_cat
+(
+    id         SERIAL PRIMARY KEY,
+    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    address_id INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    msg_sender INTEGER NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    what       TEXT,
+    data       TEXT,
```
elizabethengelman: Not necessarily required, but I wonder if we could make this column a FK to the address table?

elizabethengelman (commented):
Looking good! I'll probably take another look once #300 is merged in, so I can more easily see what this PR is adding.

transformers/events/flip_file/cat/transformer_test.go
```diff
@@ -0,0 +1,60 @@
...
+		Expect(addressErr).NotTo(HaveOccurred())
+
+		msgSenderID, msgSenderErr := shared.GetOrCreateAddress("0xbe8e3e3618f7474f8cb1d074a26affef007e98fb", db)
+		Expect(msgSenderErr).NotTo(HaveOccurred())
+
+		dataColumnID, dataColumnErr := shared.GetOrCreateAddress("0xA950524441892A31ebddF91d3cEEFa04Bf454466", db)
+		Expect(dataColumnErr).NotTo(HaveOccurred())
+
+		expectedFlipFileCat := test_data.FlipFileCatModel()
+		expectedFlipFileCat.ColumnValues[event.AddressFK] = addressID
```
elizabethengelman: Not merge blocking, but I think we can use the `AssignMessageSenderID` and `AssignAddressID` helper methods here.

transformers/shared/constants/signature.go
```diff
@@ -27,6 +27,7 @@ func DealSignature() string                { return getLogNoteTopicZero(dealMeth
 func DentSignature() string                { return getLogNoteTopicZero(dentMethod()) }
 func DenySignature() string                { return getLogNoteTopicZero(denyMethod()) }
 func FlapKickSignature() string            { return getEventTopicZero(flapKickMethod()) }
+func FlipFileCatSignature() string         { return getEventTopicZero(flipFileCatMethod()) }
```
elizabethengelman: I think that we'd want to use `getLogNoteTopicZero` method here instead since the file method has the note modifier, instead of emitting a custom event.

That would make the returned signature: `0xd4e8be8300000000000000000000000000000000000000000000000000000000`

transformers/integration_tests/flip_file_cat_test.go
```diff
@@ -0,0 +1,54 @@
...
+	})
+
+	flipFileCatConfig := event.TransformerConfig{
+		TransformerName:   constants.FlipFileCatTable,
+		ContractAddresses: test_data.FlipV110Addresses(),
+		ContractAbi:       constants.FlipV110ABI(),
+		Topic:             constants.FlipFileCatSignature(),
+	}
+
+	XIt("fetches and transforms a Flip File Cat event", func() {
```
elizabethengelman: I think that there was a cat file event at 10769102, so we can probably un-pend this integration test.

elizabethengelman (changes requested):
The only thing that should probably be updated before merging this in is how we're getting the FlipFileCatSignature.

transformers/integration_tests/flip_file_cat_test.go
yaoandrew: I didn't see anything on the `MCD_FLIP_BAT_A_1.1.0`, `MCD_FLIP_ETH_A_1.1.0`, `MCD_FLIP_KNC_A_1.1.0`, or `MCD_FLIP_PAXUSD_A_1.1.1` contracts. I've been assuming this event would be from calling `function file(bytes32 what, address data)` where `what = "cat"` on one of those `Flip` contracts

elizabethengelman: Doh, you're right. I was looking at the wrong contract. Sorry about that!

elizabethengelman (approved):

PR #310, opened 2020-09-15
Add liq 1.2, PAXUSD and USDT contracts for backfill events process

elizabethengelman:
TODO:
- [x] add the ending block number to the docker script
- [x] make sure that new liq 1.2 transformers are added to config file

### DIFF
--- a/dockerfiles/backfill_events/startup_script.sh
+++ b/dockerfiles/backfill_events/startup_script.sh
@@ -1,10 +1,4 @@
 #!/bin/sh
-
-if test -z "$ENDING_BLOCK_NUMBER"
-then
-    echo ENDING_BLOCK_NUMBER is required and no value was given
-    exit 1
-fi
-echo "Starting event backfill from block $ENDING_BLOCK_NUMBER"
+echo "Starting event backfill from block 10875761"
 # Fire up execute
-./vulcanizedb backfillEvents --config config.toml -e $ENDING_BLOCK_NUMBER
+./vulcanizedb backfillEvents --config config.toml -e 10875761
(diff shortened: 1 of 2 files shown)


dockerfiles/backfill_events/startup_script.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+ENDING_BLOCK_NUMBER=10875761
```
rmulhol: I'm wondering if we might want to remove the `ENDING_BLOCK_NUMER` var from this file entirely and just hard code `10875761` as the value passed to `-e` ... not sure if an env var in the deploy could end up overriding the value we're assigning here

environments/backfillEvents.toml
```diff
@@ -157,6 +488,10 @@
...
         deployed = 8928152
+    [contract.MCD_VAT]
```
rmulhol, line 491: Are we adding the CDP_MANAGER and MCD_VAT here to avoid forgetting to add them in the future?

rmulhol (approved):
LGTM - minor questions but I don't think they're blocking

environments/backfillEvents.toml
elizabethengelman: Yep, just noticed that they were missing, so I though I'd add them to avoid confusion. 

Also, I think this may be leftover from in the past where we had to make sure to configure all contract addresses, otherwise loading the config would fail. I feel like this isn't the case anymore, but wanted to make sure not to fail for that reason.

dockerfiles/backfill_events/startup_script.sh
elizabethengelman: oh yeah, good point - like if we have an ENDING_BLOCK_NUMBER set in the environment already, it could overwrite this value. i'll update shortly.

PR #316, opened 2020-09-17
Vdb 1519 add usdt and pax staging

elizabethengelman:
cut from #300


environments/extractDiffs.toml
```diff
@@ -43,6 +43,10 @@
         address  = "0xec25Ca3fFa512afbb1784E17f1D414E16D01794F"
```
elizabethengelman, line 99: reminder: need to update these to use the versioned names

paytonrules (approved):
My ignorance is talking here - but how come you didn't have to write any actual behavior? It appears to all be config and magic numbers.

rmulhol (approved):
:shipit:

elizabethengelman:
@paytonrules this PR is just config updates because the new collaterals consist of new flip, median and osm contracts (for each collateral). And the way that those transformers are created, we can just add a new contracts for them to watch.

paytonrules:
@elizabethengelman Thanks, how do you know what the numbers are for those in the tests? Are those block numbers for actual data?

None of this is meant to be a criticism of your code, I just want to understand this better.

elizabethengelman:
@paytonrules Yeah, totally! Those numbers are block numbers, and then values from actual events. I went onto etherscan and found events on the contracts and then just added those to the integration tests.

PR #320, opened 2020-09-23
Restore the Validationmigrationorder Check

paytonrules:
Just turns it back on.


rmulhol:
LGTM but would hold off on merging until pretty much everything else is through

yaoandrew (approved):
LGTM :shipit:

PR #324, opened 2020-09-28
VDB-1653 - add COMP, LINK, LRC to beta

elizabethengelman:


transformers/integration_tests/integration_tests_suite_test.go
```diff
@@ -25,7 +25,7 @@ var (
 
 var _ = BeforeSuite(func() {
 	testConfig := viper.New()
-	testConfig.SetConfigName("testing")
+	testConfig.SetConfigName("docker")
 	testConfig.AddConfigPath("$GOPATH/src/github.com/makerdao/vdb-mcd-transformers/environments/")
 	err := testConfig.ReadInConfig()
```
rmulhol: How does the DB get set for the integration tests?

environments/docker.toml
```diff
@@ -1660,6 +1799,14 @@
...
...s","name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[],"name":"start","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":false,"inputs":[{"internalType":"uint16","name":"ts","type":"uint16"}],"name":"step","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":false,"inputs":[],"name":"stop","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[],"name":"stopped","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[],"name":"void","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"wards","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"zzz","outputs":[{"internalType":"uint64","name":"","type":"uint64"}],"payable":false,"stateMutability":"view","type":"function"}]'
```
rmulhol: maybe it makes sense to rename this file if it's being used for tests as well?

rmulhol (approved):
Awesome refactor! Have you been able to run `compose` with just the one config file? Wondering if we might want to update something in the docs to explain how to use env vars for providing a DB connection when the main config is only transformers + addresses

transformers/integration_tests/integration_tests_suite_test.go
elizabethengelman: Great question - we're calling `test_config.SetTestConfig` in the 0.go file, which sets the db info. I think we must be loading 0.go somewhere in setting up the integration test suite. I'll try to see if I can call `test_config.SetTestConfig` more explicitly so that it's clearer.

elizabethengelman: Ohhhh, so on line 47, we're calling `db = test_config.NewTestDB(blockChain.Node())` which initializes the test_config.go file, which sets the test config via `SetTestConfig`. I'll plan to add a comment explaining this.

environments/docker.toml
elizabethengelman: 👍 that makes sense!

elizabethengelman:
@rmulhol Just tested out compose and execute, and they both work. Made a small update to the README, but happy to expand upon that as well.

PR #325, opened 2020-09-28
Vdb 1653 backfill events

elizabethengelman:
Cut from #324. Still need to update the to block number in the startup script, once #324 is deployed.


rmulhol (commented):
Need to update the exporter, otherwise LGTM

rmulhol (approved):

PR #326, opened 2020-09-29
Use buster not alpine as base image

paytonrules:
This fixes the infamous "execute hangs on linking to plugin" error . There's also a couple of small tweaks to the makefile to make it a little easier to use.

I believe I said I was going to use stretch a few times - but in fact stretch isn't supported after go 14.


rmulhol (approved):
🥇

PR #328, opened 2020-09-30
Vdb 1654 new collaterals to staging

elizabethengelman:


environments/backfillStorage.toml
```diff
@@ -206,31 +230,3 @@
...
         deployed = 9417183
-    [contract.OSM_BAT]
```
elizabethengelman, line 209: Removing these from from this config file because we're not currently tracking OSM storage slots, so these would be unused.

README.md
```diff
@@ -148,9 +153,7 @@ shell's `$PATH`.
 [Postgraphile](https://www.graphile.org/postgraphile/) is used to expose GraphQL endpoints for our database schemas, this is described in detail [here](https://github.com/makerdao/vulcanizedb/blob/staging/documentation/postgraphile.md).
 
 ### Tests
-- Set the ipc path to a Kovan node either by:
-    - replacing the empty `ipcPath` in the `environments/testing.toml` with a path to a full node's eth_jsonrpc endpoint (e.g. local geth node ipc path or infura url)
-    - Or, setting the CLIENT_IPCPATH environment variable
+- Set the ipc path to a Kovan node either by setting the CLIENT_IPCPATH environment variable.
```
rmulhol: Not introduced by this PR but maybe we could remove the reference to Kovan since all integration tests point at mainnet now? Also seems like we can delete the word "either"

test_config/test_config.go
```diff
@@ -28,29 +28,36 @@ import (
...
-	TestConfig = viper.New()
-	TestConfig.SetConfigName("testing")
-	TestConfig.AddConfigPath("$GOPATH/src/github.com/makerdao/vdb-mcd-transformers/environments/")
-	err := TestConfig.ReadInConfig()
-	if err != nil {
-		log.Fatal(err)
+func SetTestConfig() {
+	viper.AddConfigPath("$GOPATH/src/github.com/makerdao/vdb-mcd-transformers/environments/")
+
+	viper.SetConfigName("testDatabase")
```
rmulhol, line 44: 🥇 had been wondering about `TestConfig`...

test_config/test_config.go
```diff
@@ -28,29 +28,36 @@ import (
...
+	readConfigErr := viper.MergeInConfig()
+	if readConfigErr != nil {
+		log.Fatal(readConfigErr)
 	}
-	ipc := TestConfig.GetString("client.ipcPath")
-	hn := TestConfig.GetString("database.hostname")
-	port := TestConfig.GetInt("database.port")
-	name := TestConfig.GetString("database.name")
+
+	ipc := viper.GetString("client.ipcPath")
```
rmulhol, line 56: Not your responsibility in this PR but I'm wondering if we can either remove the `ipc` + `TestClient` or use the `ipc` derived here in the places (like integration test config) where we're depending on this to grab the DB config

transformers/integration_tests/integration_tests_suite_test.go
```diff
@@ -24,26 +24,22 @@ var (
...
-	Expect(ipc).NotTo(BeEmpty(), "testing.toml IPC path or $CLIENT_IPCPATH env variable need to be set")
+	Expect(ipc).NotTo(BeEmpty(), "$CLIENT_IPCPATH env variable need to be set")
 
 	rpcClient, ethClient, clientErr := getClients(ipc)
 	Expect(clientErr).NotTo(HaveOccurred())
 	var blockChainErr error
 	blockChain, blockChainErr = getBlockChain(rpcClient, ethClient)
 	Expect(blockChainErr).NotTo(HaveOccurred())
 
+	// the init function in test_config set the config files for the test database and transformers
```
rmulhol, line 42: Clarifying the above, wondering if it would make sense to also have something like `blockchain = test_config.NewTestBlockchain()` that uses the `ipc` we have as a result of calling `init` in the `test_config`

rmulhol: Though maybe that's a bad idea if it means we'd fail silently if there were no ipc path? 🤔

rmulhol (approved):
🚀

test_config/test_config.go
elizabethengelman: I need to think on this a bit, so will probably do it as part of a new PR.

PR #340, opened 2020-10-07
Update deps and docs for prod

rmulhol:
- point at vdb v0.1.0
- modify default branch to point at prod in docker and elsewhere


paytonrules (approved):
🥳

PR #341, opened 2020-10-08
Update docker compose

elizabethengelman:
updates docker-compose.yml to use rpc pubsub geth patch (instead of the statediff service patch)

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -74,7 +74,7 @@ services:
     restart: always
 
   postgraphile:
-    image: makerdao/vdb-postgraphile
+    image: makerdao/vdb-postgraphile:latest
     ports:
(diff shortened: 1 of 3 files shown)


dockerfiles/docker-compose.yml
```diff
@@ -93,58 +93,10 @@ services:
       - 30303
     command: [
       "--verbosity", "2",
-      "--rpc",
-      "--rpcaddr", "0.0.0.0",
+      "--http",
```
elizabethengelman, line 96: these are updates based on warnings when running geth -  the `rpc` flags are being deprecated in favor of `http`.

paytonrules: 👍

paytonrules (approved):

rmulhol (approved):

PR #342, opened 2020-10-09
Remove shared functions inserting addresses

rmulhol:
- no need for wrappers with the same signature as the underlying
  function

This is a bit of a preparatory refactor for extracting shared code to a separate repo. Since the code that these shared functions were using already lives in vDB, it seems like we can probably just depend on that underlying code.


elizabethengelman (approved):
👍

PR #344, opened 2020-10-20
VDB-1550 Collateral Generator

elizabethengelman:


cmd/newCollateralGenerator.go
```diff
@@ -0,0 +1,240 @@
...
+	collateral             types.Collateral
+	medianContractRequired bool
+	osmContractRequired    bool
+)
+
+var addNewCollateralCmd = &cobra.Command{
+	Use:     "addNewCollateral",
+	Short:   "",
+	Long:    ``,
+	Example: ``,
```
paytonrules: Don't you want an example? I get the lack of a long and short.

elizabethengelman: Nice call - just pushed an update.

cmd/newCollateralGenerator.go
```diff
@@ -0,0 +1,101 @@
...
+		return collateralErr
+	}
+
+	contracts, contractsErr := prompter.GetContractDetails()
+	if contractsErr != nil {
+		return contractsErr
+	}
+
+	configUpdater := config.NewConfigUpdater(collateral, contracts, prompter.MedianRequired, prompter.OsmRequired)
+	configFileName := "mcdTransformers"
```
paytonrules: Is this always mcdTransformers? If so maybe it should be a constant.

generators/new_collateral/config/parser_test.go
```diff
@@ -0,0 +1,53 @@
...
+		It("reads in the contracts", func() {
+			expectedContracts := types.Contracts{
+				"MCD_CAT_1_0_0": test_data.Cat100Contract,
+				"MCD_CAT_1_1_0": test_data.Cat110Contract,
+			}
+			config, parseErr := configParser.ParseCurrentConfig(testConfigFilePath, testConfigFileName)
+			Expect(parseErr).NotTo(HaveOccurred())
+			Expect(config.Contracts).To(Equal(expectedContracts))
+		})
+	})
```
paytonrules, line 139: What about the expected errors?

cmd/newCollateralGenerator.go
rmulhol: agree with the sentiment - would also consider making the config file configurable via CLI flag with this as the default. Trying to think about how we can make it obvious that renaming that file will require a change here 🤔

cmd/newCollateralGenerator.go
```diff
@@ -0,0 +1,101 @@
...
+	configFilePath := helpers.GetEnvironmentsPath()
```
rmulhol: also wondering if we might want to be able to specify the path, or maybe just loudly warn that this command depends on the config file living in `$GOPATH/src/github.com/makerdao/vdb-mcd-transformers/environments/`

cmd/newCollateralGenerator.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	newCollateralGenerator := new_collateral.NewCollateralGenerator{
+		ConfigFileName:       configFileName,
+		ConfigFilePath:       configFilePath,
+		ConfigParser:         config.Parser{},
+		ConfigUpdater:        configUpdater,
+		InitializerGenerator: &initializerWriter,
+	}
+
+	fmt.Println(fmt.Sprintf("Adding %s to %s", collateral.Name, helpers.GetFullConfigFilePath(configFileName, configFilePath)))
```
rmulhol: maybe worth using logging facilities here?

generators/new_collateral/prompts/prompts_wrapper.go
```diff
@@ -0,0 +1,153 @@
+package prompts
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
```
paytonrules, line 8: Neat!

generators/new_collateral/config/parser.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/BurntSushi/toml"
+	"github.com/makerdao/vdb-mcd-transformers/generators/new_collateral/helpers"
+	"github.com/makerdao/vdb-mcd-transformers/generators/new_collateral/types"
+	"github.com/mitchellh/mapstructure"
+)
+
+type IParse interface {
+	ParseCurrentConfig(configFilePath, configFileName string) (types.TransformersConfig, error)
+}
+type Parser struct{}
```
rmulhol: Really minor, but I wonder if we might want to name the interface `Parser` and then not export the `struct{}` (e.g. make it `parser`) and then use a constructor to return that private implementation

cmd/newCollateralGenerator.go
```diff
@@ -0,0 +1,101 @@
...
+	addErr := newCollateralGenerator.UpdateConfig()
+	if addErr != nil {
+		return addErr
```
rmulhol: maybe worth decorating the errors with `fmt.Errorf("here's where we failed: %w", err)` throughout to make it clear where we failed if something goes wrong/bubbles up

generators/new_collateral/initializer/generator.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+type Generator struct {
+	Collateral                types.Collateral
+	MedianInitializerRequired bool
+}
+
+var initializerFileName = "initializer.go"
+
+func (g *Generator) GenerateFlipInitializer() error {
```
rmulhol, line 31: Do the methods in this file require pointer receivers?

generators/new_collateral/test_data/testConfig.toml
```diff
@@ -0,0 +1,34 @@
...
...":"bytes32"}],"name":"ilks","outputs":[{"internalType":"address","name":"flip","type":"address"},{"internalType":"uint256","name":"chop","type":"uint256"},{"internalType":"uint256","name":"lump","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"live","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[{"internalType":"address","name":"usr","type":"address"}],"name":"rely","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[],"name":"vat","outputs":[{"internalType":"contract VatLike","name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"vow","outputs":[{"internalType":"contract VowLike","name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"wards","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"}]'
```
paytonrules: Do the tests need the entire ABIs?

generators/new_collateral/initializer/generator.go
```diff
@@ -0,0 +1,70 @@
...
+	initializer := g.createInitializer(g.Collateral.GetFlipContractName(), "flip")
+	//create the path to the initializer file
+	path := g.Collateral.GetAbsoluteFlipStorageInitializersDirectoryPath()
+	mkDirErr := os.MkdirAll(path, os.ModePerm)
```
rmulhol, line 35: could go either way here but I wonder if we might want to consider extracting/wrapping these filesystem IO operations and mocking them in the these tests? Not sure if that gets super complex when calling `Save` on the object returned by `jen.NewFile` 🤔

generators/new_collateral/config/parser.go
```diff
@@ -0,0 +1,85 @@
...
+	}, nil
+}
+
+func parseExporterMetaData(tomlConfig types.TransformersConfigForToml) (types.ExporterMetaData, error) {
+	home, homeOk := tomlConfig.Exporter["home"].(string)
+	name, nameOk := tomlConfig.Exporter["name"].(string)
+	save, saveOk := tomlConfig.Exporter["save"].(bool)
+	schema, schemaOk := tomlConfig.Exporter["schema"].(string)
+	if !homeOk || !nameOk || !saveOk || !schemaOk {
+		return types.ExporterMetaData{}, errors.New("error asserting exporter meta data types")
```
rmulhol: also super minor but I wonder if it'd be helpful to format in the various `ok` values so that we can see what's missing in the error

generators/new_collateral/config/parser.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	var transformerNames []string
+	decodeErr := mapstructure.Decode(tomlConfig.Exporter["transformerNames"], &transformerNames)
```
rmulhol, line 62: do we want to check `tomlConfig.Exporter["transformerNames"]` with a similar `ok` pattern? Not sure if we could run into null pointer errors or silent failures here

generators/new_collateral/config/parser.go
```diff
@@ -0,0 +1,85 @@
...
+		TransformerNames: transformerNames,
+	}, nil
+}
+
+func parseTransformerExporters(tomlConfig types.TransformersConfigForToml) (types.TransformerExporters, error) {
+	delete(tomlConfig.Exporter, "home")
+	delete(tomlConfig.Exporter, "name")
+	delete(tomlConfig.Exporter, "save")
+	delete(tomlConfig.Exporter, "schema")
+	delete(tomlConfig.Exporter, "transformerNames")
```
rmulhol: this makes me nervous 😅 

could you help me understand what we're doing here?

rmulhol (commented):
Looking good! Have a few questions, but super excited about where this is headed 🎉 🚀

generators/new_collateral/initializer/generator_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+		fileInfo, fileErr := os.Stat(testMedianCollateralFullPath)
+		Expect(os.IsNotExist(fileErr)).To(BeFalse())
+		Expect(fileInfo.IsDir()).To(BeFalse())
+
+		removeTestFile := os.RemoveAll(testMedianCollateralPath)
+		Expect(removeTestFile).NotTo(HaveOccurred())
+	})
+
+	It("doesn't create a median initializer if it is not configured to do so", func() {
```
paytonrules: Kinda weird there's not a test for the actual contents of the file, but testing strings can suck so 🤷

generators/new_collateral/config/parser.go
paytonrules: That would be more go-ish wouldn't it?

generators/new_collateral/config/parser.go
paytonrules: Gonna take a whack to make sure I understand - this removes it from an in memory map right?

paytonrules (approved):
I have a few questions, but all minor stuff, and this all looks really good.

I'd lean towards merging soon, because even if there are errors we still have to review the code that's generated. A bug isn't likely to cause anything other than inconvenience for us.

cmd/newCollateralGenerator.go
elizabethengelman: Great point - also could see this being helpful if we ever want to extend this command work for backfill config files.

generators/new_collateral/config/parser.go
elizabethengelman: Nice call. 

I was a bit on the fence on wether `parser{}` is really necessary at all since we're not setting any state on it. I decided to put `ParseCurrentConfig` on a parser receiver, so I could create an interface, and use that for testing. But up for other ideas too!

generators/new_collateral/config/parser.go
elizabethengelman: TL;DR: Yep, like Eric said, this piece is removing these keys from an in-memory map. The reason I'm doing that is so it is possible to iterate through the rest of the key-value pairs in this map, which are `transformerExporters`, and then decode those into `TransformerExporter` structs. Since the `transformerExporters` have dynamic keys, it makes it a bit more challenging to auto-decode them as the config file is currently structured.

Longer explanation:
I think that this parser could be cleaned up quite a bit if we change the config file structure - currently we basically have `exporterMetadata` (e.g. "name", "save", "home", etc) and the `transformerExporters` (e.g. exporter.flip_eth_a_v1_1_0, etc) in the same `exporter` map in the config file. But since the `transformerExporter` keys are dynamic, I don't think that we can automatically decode the whole `exporter` map into a struct since the decoder won't know what field to put those `transformerExporters` in. The workaround I'm using in the parser, is to decode the whole `exporter` map into a `map[string]interface{}`, where the values are either the `exporterMetadata` fields, or the `transformerExporters` names.

Then `parseExporterMetaData` and `parseTransformerExporters` are the hacky methods to separate out that generic `exporter` map into two different maps, which can then be automatically decoded into `ExporterMetaData` and `TransformerExporters` structs, which make things ways easier to deal with while we're adding new transformers.

That ended up being longer than I meant it to, and would be happy to chat more about it outside of this PR! 😄

generators/new_collateral/config/parser.go
elizabethengelman: I just tested it out, and it looks like if `tomlConfig.Exporter["transformerNames"]` were nil, `mapstructure` would just decode it into an empty `[]string`, which I think is ok. Happy to reconsider this, and add a check since we probably don't want an empty slice of transformer names, but OTHO it shouldn't break anything.

generators/new_collateral/config/parser.go
elizabethengelman: addressed in 0018720

generators/new_collateral/config/parser.go
elizabethengelman: addressed in 0018720bd0c7d45029d0e522e0f9e015f202fee8

elizabethengelman: addressed in 0018720bd0c7d45029d0e522e0f9e015f202fee8

generators/new_collateral/config/parser.go
elizabethengelman: wrote a test for it in 938de7a859b2a550b95a0efe656128c314b15f6e

generators/new_collateral/config/parser_test.go
elizabethengelman: good call! added tests in 938de7a859b2a550b95a0efe656128c314b15f6e

I'd be interested in your thoughts about other ways to test some of the functionality in the parser - the easiest way I could think of was to export some of the previously private methods, but not sure if maybe there's a better way to handle this that I'm not seeing at the moment.

generators/new_collateral/test_data/testConfig.toml
elizabethengelman: nope, probably not! i can cut them down.  addressed in ef697ee899f73ccafa9704e28a8b72a1618be8b9

cmd/newCollateralGenerator.go
elizabethengelman: I had initially used `fmt.Println` for this so that we get some feedback in the command line while running this command, rather than needing to tail the logs to make sure that it's working. But the success/error messages end up being put in the logs, so we need to tail those anyway to see any potential errors. 🤷

generators/new_collateral/initializer/generator.go
elizabethengelman: good call! updated in ef697ee899f73ccafa9704e28a8b72a1618be8b9

cmd/newCollateralGenerator.go
elizabethengelman: Actually, just moved some of this code into the generator pkg in 152b8c83f787e7627c451df9695463dee98e8b65, and so as to not print out a lot of output during tests, I've switched to logging vs printing for these.

cmd/newCollateralGenerator.go
elizabethengelman: addresses in 663448e6dd0071b7917c35e47b3d5e53cc9071e4

generators/new_collateral/config/parser.go
elizabethengelman: Just made a change to just skip the metadata keys, rather than removing them from the map: 110429f714de510edf1148fc4b7d65f7c6772dc0

generators/new_collateral/initializer/generator_test.go
elizabethengelman: good call - added a test in ce46cfa

cmd/newCollateralGenerator.go
elizabethengelman: makes sense - added in 4ba6331

generators/new_collateral/initializer/generator.go
elizabethengelman: I really like this idea, but may hold off on it for a bit - now that I've added a test for the contents of the initializer, I kind of like being able to see the output of calling `Save` on the `jen.NewFile`. I also think that having a `jen` wrapper may be cool but was struggling to find a good way to do that without needing to mock out each of the render methods, eg `Var()`, `Id()`, `Qual()`, etc.

generators/new_collateral/test_data/testConfig.toml
```diff
@@ -0,0 +1,34 @@
...
+        migrations = "db/migrations"
+        rank = "0"
+    [exporter.cat_file_vow]
+        path = "transformers/events/cat_file/vow/initializer"
+        type = "eth_event"
+        repository = "github.com/makerdao/vdb-mcd-transformers"
+        migrations = "db/migrations"
+        contracts = [
+            "MCD_CAT_1.0.0",
+            "MCD_CAT_1.1.0"
```
rmulhol: I might be missing some nuance but shouldn't the separator be `_` and not `.` here?

generators/new_collateral/test_data/test_data.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	CatFileVowExporterMap = map[string]interface{}{
+		"path":       CatFileVowExporter.Path,
+		"type":       CatFileVowExporter.Type,
+		"repository": CatFileVowExporter.Repository,
+		"migrations": CatFileVowExporter.Migrations,
+		"rank":       CatFileVowExporter.Rank,
+		"contracts": []interface{}{
+			"MCD_CAT_1.0.0", "MCD_CAT_1.1.0",
```
rmulhol: Similar question on `_` vs `.` here

generators/new_collateral/types/collateral.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (c Collateral) GetMedianContractName() string {
+	// example: MEDIAN_ETH_B
+	name := strings.ToUpper(strings.Replace(c.Name, "-", "_", -1))
+	return fmt.Sprintf("MEDIAN_%s", name)
+}
+
+func (c Collateral) GetOsmContractName() string {
+	// example: MEDIAN_ETH_B
```
rmulhol: :%s/MEDIAN/OSM

generators/new_collateral/types/collateral.go
```diff
@@ -0,0 +1,66 @@
...
+func (c Collateral) GetMedianInitializerDirectory() string {
+	// example: median_eth_b
+	name := strings.ToLower(strings.Replace(c.Name, "-", "_", -1))
+	return fmt.Sprintf("median_%s", name)
+}
+
+func (c Collateral) GetFlipContractName() string {
+	// example: MCD_FLIP_ETH_B_1_1_3
+	name := strings.ToUpper(strings.Replace(c.Name, "-", "_", -1))
+	version := fmt.Sprintf("%s", strings.Replace(c.Version, ".", "_", -1))
```
rmulhol, line 52: Just making sure I understand - do we need this additional line because these contracts have a version? Will changes be required in this namespace if we add versions to other contracts in the future?

generators/new_collateral/initializer/generator.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (g Generator) GenerateMedianInitializer() error {
+	if g.MedianInitializerRequired {
+		initializer := g.createInitializer(g.Collateral.GetMedianInitializerDirectory(), g.Collateral.GetMedianContractName(), "median")
+
+		path := g.GetAbsoluteMedianStorageInitializersDirectoryPath()
+		mkDirErr := os.MkdirAll(path, os.ModePerm)
+		if mkDirErr != nil {
+			return mkDirErr
```
rmulhol, line 55: Don't want to go overboard here (don't need to do this as part of this PR) - but I think at least in the medium/long term it'd be cool if wrapped _every_ error in `fmt.Errorf` to clarify exactly where an error bubbled up from if we're logging it and exiting the command

generators/new_collateral/config/updater.go
```diff
@@ -0,0 +1,278 @@
...
+	addContractsToExportersErr := cg.addContractsToEventExporters()
+	if addContractsToExportersErr != nil {
+		return addContractsToExportersErr
+	}
+	cg.addContracts()
+
+	return nil
+}
+
+func (cg *Updater) copyInitialConfig() error {
```
rmulhol: nbd but maybe want to rename the receiver to something other than `cg` now that this is called an `Updater`?

generators/new_collateral/config/parser_test.go
```diff
@@ -0,0 +1,140 @@
...
+		})
+
+		It("returns an error if it fails to decode the file", func() {
+			_, parseErr := configParser.ParseCurrentConfig(testConfigFilePath, badTestConfigFile)
+			Expect(parseErr).To(HaveOccurred())
+			Expect(parseErr).To(MatchError("open ../test_data/non-existent-file.toml: no such file or directory"))
+		})
+	})
+
+	Context("ParseExporterMetadata", func() {
```
rmulhol, line 50: nbd and probably unnecessary, but is there a reason we can't make these private and test them through `ParseCurrentConfigFile`?

rmulhol (approved):
LGTM!

Found myself asking questions but I think they're mostly nitpicks and would probably be best addressed in a separate story

generators/new_collateral/test_data/testConfig.toml
elizabethengelman: Yep, this was a copy pasta error I think - it should be `_`. It raises a good point though, as of https://github.com/makerdao/vulcanizedb/pull/147 we now have a way to catch if a `contract.MCD...` entry has contract name formatted incorrectly, but we still don't have a way to catch if we have a list of contracts for a exporter formatted incorrectly.

generators/new_collateral/types/collateral.go
elizabethengelman: Yep, that's right - this was just a way to encapsulate formatting the contract name + it's version. If we decide to add versions to Median and OSM, we'll need to add the collateral version to the `GetOsmContractName` and `GetMedianContractName` methods.

generators/new_collateral/initializer/generator.go
elizabethengelman: 👍  planning to do this in a new PR

generators/new_collateral/config/parser_test.go
elizabethengelman: 👍  planning to take a look at this when working on getting a new PR together

PR #345, opened 2020-10-20
Import shared utilities

rmulhol:
- remove functions that have been extracted to separate repo


go.mod
```diff
@@ -3,15 +3,16 @@ module github.com/makerdao/vdb-mcd-transformers
...
 	github.com/makerdao/vulcanizedb v0.1.0
-	github.com/onsi/ginkgo v1.14.0
-	github.com/onsi/gomega v1.10.1
-	github.com/sirupsen/logrus v1.2.0
+	github.com/onsi/ginkgo v1.14.1
+	github.com/onsi/gomega v1.10.2
+	github.com/sirupsen/logrus v1.7.0
 	github.com/spf13/cobra v0.0.3
-	github.com/spf13/viper v1.3.2
+	github.com/spf13/viper v1.7.1
```
rmulhol: A little nervous about these deps updating, but also wondering if we might want to go ahead and update them on the vulcanizedb side since this happened automatically

transformers/component_tests/queries/test_helpers/test_helpers.go
```diff
@@ -25,7 +25,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/makerdao/vdb-mcd-transformers/transformers/shared"
+	shared2 "github.com/makerdao/vdb-mcd-transformers/transformers/shared"
```
rmulhol: happy to replace `shared2` with a better name in places where we need both the separate repo and a mcd-version specific version from here, but also wondering if we might want to rename that package altogether and/or move functions to packages with more specific names

go.mod
paytonrules: Your nerves were right - you probably saw that this failed to start the execute container.

transformers/component_tests/queries/test_helpers/test_helpers.go
paytonrules: In the interim you could use transformers_shared.

transformers/integration_tests/median_lift_test.go
```diff
@@ -7,8 +7,9 @@ import (
 	"github.com/lib/pq"
 	"github.com/makerdao/vdb-mcd-transformers/test_config"
 	"github.com/makerdao/vdb-mcd-transformers/transformers/events/median_lift"
-	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	constants2 "github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
```
paytonrules: Maybe transformersConstants - instead of constants2

paytonrules (approved):
I think the compiler did an excellent job of reviewing 99% of this code.

PR #346, opened 2020-10-21
Vdb 1647 - Add a healthcheck on the execute container

paytonrules:
We use Github actions because travis has trouble with docker-in-docker.


.github/workflows/check_execute
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python
+import subprocess, sys, argparse, time
+
+parser = argparse.ArgumentParser(description="Check the startup of the execute container")
```
rmulhol: nbd but if we can configure the image name, maybe we don't want to specify what container is having its startup checked? Or maybe we don't need to make the image name configurable? No changes required here, just thinking out loud

.github/workflows/check_execute
```diff
@@ -0,0 +1,58 @@
...
+parser.add_argument("image", help="name of the image you're running")
+parser.add_argument("--client_ipcpath", help="IPC Path of the geth container", required=True)
+parser.add_argument("--database_hostname", help="hostname of the postgres instance", default="postgres")
+parser.add_argument("--database_name", help="name of the postgres database", default="postgres")
+parser.add_argument("--database_port", help="port of the postgres database", default="5432")
+parser.add_argument("--database_user", help="username of the postgres database user", default="postgres")
+parser.add_argument("--database_password", help="password of the postgres database user", default="postgres")
+parser.add_argument("--wait_time", help="seconds to wait for the container to be ready", default="60", type=int)
```
rmulhol, line 12: This strikes me as a potentially low number - any ideas on how long it's taking for the container to be ready right now?

.github/workflows/test-execute.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      -
+        name: Build execute image
+        uses: docker/build-push-action@v2
+        with:
+          file: ./dockerfiles/execute/Dockerfile
+          push: false
+          tags: makerdao/execute:latest
```
rmulhol: I get that this is configured to `push: false`, but I'd still prefer to use a different tag so that if we mess that up we don't push/override a `latest` image. Maybe something like `makerdao/execute:gh_actions_test_version`?

rmulhol (approved):
Thanks for finding a way to get this done!

For posterity: we're putting this in Github Actions instead of Travis bc `docker run` hits an error on `linking plugin` from mismatched imports between the transformers and vdb. This has proven difficult to reproduce locally (or even in other phases of the Travis pipeline than `script`), and we haven't found a way to resolve - but the issue does not come up with this approach.

.github/workflows/check_execute
paytonrules: About 30 seconds. Surprised me too - you can see it in the logs on the actions tab.

.github/workflows/check_execute
paytonrules: I can interpolate the image. Seems like the right call there.

.github/workflows/test-execute.yml
paytonrules: 👍  - I meant to change the name in the first place.

PR #349, opened 2020-10-21
Vdb 1669 add eth b beta

elizabethengelman:
FYI: the config file and transporterExporter.go files were not created with the generator, but i think that it should work once the generator code is finalized and ported over to beta.


test_config/test_config.go
```diff
@@ -48,15 +48,17 @@ func SetTestConfig() {
 	viper.AddConfigPath("$GOPATH/src/github.com/makerdao/vdb-mcd-transformers/environments/")
 
 	viper.SetConfigName("testDatabase")
-	mergeConfigErr := viper.ReadInConfig()
-	if mergeConfigErr != nil {
-		log.Fatal(mergeConfigErr)
+	readConfigerr := viper.ReadInConfig()
```
rmulhol, line 51: s/readConfigerr/readConfigErr

also, nice catch!

rmulhol (approved):
LGTM!

idk why this change prompted this thought, and we def don't need to resolve it as part of this PR - but I wonder if we might be able to create collections of addresses in the config (e.g. `MCD_FLIP_ADDRESSES`) so that instead of having to copy `MCD_FLIP_ETH_B_1_!_3` for each transformer, we could add it in one place and use that collection everywhere

PR #350, opened 2020-10-22
Vdb 1674 backfill eth b (beta)

elizabethengelman:
To be merged in after #349

### DIFF
--- a/environments/backfillEvents.toml
+++ b/environments/backfillEvents.toml
@@ -2,7 +2,89 @@
     home     = "github.com/makerdao/vulcanizedb"
     name     = "transformerExporter"
     save     = true
-    transformerNames = []
+    transformerNames = [
+        "auction_file",
+        "deal",
+        "dent",
+        "deny",
+        "flip_file_cat",
+        "flip_kick",
+        "rely",
+        "tend",
+        "tick",
+        "yank"
+    ]
+
+    [exporter.auction_file]
+        path = "transformers/events/auction_file/initializer"
+        type = "eth_event"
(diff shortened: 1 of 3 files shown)


environments/backfillEvents.toml
```diff
@@ -2,7 +2,86 @@
...
     save     = true
-    transformerNames = []
+    transformerNames = [
+        "auction_file",
+        "deal",
+        "dent",
+        "deny",
+        "flip_file_cat",
+        "flip_kick",
+        "rely",
```
rmulhol, line 12: looks like maybe missing some transformers?

elizabethengelman: 🤦‍♀️  nice catch. just pushed up a fix.

rmulhol (approved):
approving bc I don't _think_ having some transformers missing from the `transformerNames` should matter as long as they're in the exporter and they're config info is in the toml, though may be good to add `tend`, `tick`, and `yank` to that list before merging

paytonrules:
At the risk of being contrary just to be contrary ... oh wait I like doing that.  

If transformerNames don't need to be in the list - why have them? Maybe get rid of it entirely?

elizabethengelman:
@paytonrules that's a really good point - I think that we need them when we're relying on the vdb `compose` to generate `tranformersExporter.go`. We iterate through those names to know what the keys are for the exporters. But, after playing around with the [toml parser](https://github.com/BurntSushi/toml) we're using, I am starting to think it's possible to get those keys in another way. This is a good reminder that I should create a new story to try to address that in the future.

PR #353, opened 2020-10-26
Parse STORAGEDIFFS_SOURCE in extract-diffs

rmulhol:
- required to connect to non-default node, since this value is now a
  command-specific flag


paytonrules (approved):

elizabethengelman (approved):
🙏

PR #355, opened 2020-10-27
Move docker images to pre-built image bases

paytonrules:
Also fix logging db connection in run_migrations


elizabethengelman (approved):
looks good! really like this change! 👏

PR #357, opened 2020-10-28
(VDB-1634) Add event initializer tests

rmulhol:
Looking for early feedback on this general pattern for testing event initializers. If it looks good, I'll plan to add a new one for every transformer.

Key notes:
- only asserting against name, topic, starting block, and one address - should we be more exhaustive?
- trying to hedge against copy-paste errors by using raw values (we typically use `const` or `var` vals in the src). Does this make sense/seem like it will result in added-value?
- tests are in a separate directory/package from the initializer - this was necessary to force loading of the test config, so that the initializer `var` can be parsed (including config-dependent vals)


elizabethengelman (approved):
Looks good!

I definitely think these sorts of tests would help us catch instances where we don't configure an event transformer properly. Probably outside the scope of this story - but I keep wondering if there's a way to make sure that the test suite fails if we forget add an initializer altogether. That's probably a very specific edge case, so maybe we should't put too much effort into testing it, but it's a mistake that I've made once which is why I bring it up. 😀

yaoandrew (approved):
Looks good to me. 🚢 
I think using the raw values makes sense. I've had some IDE auto complete error with the wrong `const` in place.

rmulhol:
> Looks good!
> 
> I definitely think these sorts of tests would help us catch instances where we don't configure an event transformer properly. Probably outside the scope of this story - but I keep wondering if there's a way to make sure that the test suite fails if we forget add an initializer altogether. That's probably a very specific edge case, so maybe we should't put too much effort into testing it, but it's a mistake that I've made once which is why I bring it up. 😀

I like this idea a lot, but I'm wondering if we might want to approach it as a separate story - my first though about how to implement would be to do some filesystem traversal in CI and verify that every directory path terminates with a `*initializer.go` file. Don't have a great idea for how to do this sort of verification with tests 🤔 

That said, I do think it'd be worth pondering a bit more here, even in the context of this PR. Specifically, I'm wondering if/how we can encourage ourselves to keep these tests up to date if new transformers are added

PR #211, opened 2020-06-15
Add time-series, bucketed auction totals for flip/flop/flap auctions

hexonaut:
Continuing on with adding time series queries. Uses the `bid_event` trigger table to produce aggregate totals of the auctions.

Sample Request:

	{
	  timeFlipBidTotals (ilkIdentifier:"ETH-A", rangeStart:"2020-01-01", rangeEnd:"2020-06-01", bucketInterval:{months:1}) {
	    nodes {
	      bucketStart,
	      bucketEnd,
	      lotEnd,
	      bidAmountEnd
	    }
	  },
	  timeFlapBidTotals (rangeStart:"2020-01-01", rangeEnd:"2020-06-01", bucketInterval:{months:1}) {
	    nodes {
	      bucketStart,
	      bucketEnd,
	      lotEnd,
	      bidAmountEnd
	    }
	  },
	  timeFlopBidTotals (rangeStart:"2020-01-01", rangeEnd:"2020-06-01", bucketInterval:{months:1}) {
	    nodes {
	      bucketStart,
	      bucketEnd,
	      lotEnd,
	      bidAmountEnd
	    }
	  },
	  timeBiteTotals (ilkIdentifier:"ETH-A", rangeStart:"2020-01-01", rangeEnd:"2020-06-01", bucketInterval:{months:1}) {
	    nodes {
	      bucketStart,
	      bucketEnd,
	      ink,
	      art,
	      tab
	    }
	  }
	}

Sample Response:

	{
	  "data": {
	    "timeFlipBidTotals": {
	      "nodes": [
	        {
	          "bucketStart": "2020-01-01T00:00:00",
	          "bucketEnd": "2020-02-01T00:00:00",
	          "lotEnd": "210000000000000000000000000000000000000000000000000",
	          "bidAmountEnd": "594631296312757266648"
	        },
	        {
	          "bucketStart": "2020-02-01T00:00:00",
	          "bucketEnd": "2020-03-01T00:00:00",
	          "lotEnd": "360000000000000000000000000000000000000000000000000",
	          "bidAmountEnd": "594631296312757266648"
	        },
	        {
	          "bucketStart": "2020-03-01T00:00:00",
	          "bucketEnd": "2020-04-01T00:00:00",
	          "lotEnd": "210000000000000000000000000000000000000000000000000",
	          "bidAmountEnd": "419044134033302441615"
	        },
	        {
	          "bucketStart": "2020-04-01T00:00:00",
[...]

PR #360, opened 2020-11-03
Vdb 1410 add time series bucketed auction totals

paytonrules:
Take the PR here https://github.com/makerdao/vdb-mcd-transformers/pull/211 and set it up for Beta.


db/migrations/20201007080800_create_time_auction_totals_query.sql
```diff
@@ -0,0 +1,251 @@
...
+DECLARE
+    r api.time_bid_total%rowtype;
+BEGIN
+    ASSERT EXTRACT(EPOCH FROM (range_end - range_start)) / EXTRACT(EPOCH FROM bucket_interval) <= 100, 'Please limit requests to at most 100 buckets.';
+    
+    FOR r IN 
+        WITH buckets AS (
+            SELECT generate_series(range_start, range_end - bucket_interval, bucket_interval) AS bucket_start,
+            extract(epoch FROM generate_series(range_start, range_end - bucket_interval, bucket_interval)) AS bucket_start_epoch,
+            extract(epoch FROM generate_series(range_start + bucket_interval, range_end, bucket_interval)) AS bucket_end_epoch
```
elizabethengelman, line 27: This is very likely me just not understanding postgres, but I'm not quire sure I understand why we're generating series here, but then it seems like we're extracting just one value from each of those series - why generate the series at all?

elizabethengelman (approved):
LGTM! Just had one question. Also it may be nice to add a test case to check that there is a failure if the inputs result in more than 100 buckets. But happy to hold off on that test scenario and add it in later.

db/migrations/20201007080800_create_time_auction_totals_query.sql
paytonrules: It's not just one value, it's a series, it's just the postgres being complicated. 

We're actually extracting one entry from each entry in the series. So the result looks something like:

| bucket start | bucket start epoch | bucket end epoch |
| -------------|:-------------------:|-------------------:|
| "timestamp start | "timestamp start as epoch" | "timestamp end as epoch"|
| "timestamp 2 start | "timestamp 2 start as epoch" | "timestamp 2 end as epoch"|

Keeping this note here since I had to go run it in pgadmin to be sure myself.

PR #361, opened 2020-11-03
Additional updates to collateral generator

elizabethengelman:


generators/new_collateral/config/parser.go
```diff
@@ -45,23 +44,22 @@ func (parser) ParseCurrentConfig(configFilePath, configFileName string) (types.T
 	}, nil
 }
 
-// ParseExporterMetaData is exported for testing
-func ParseExporterMetaData(tomlConfig types.TransformersConfigForToml) (types.ExporterMetaData, error) {
+func parseExporterMetaData(tomlConfig types.TransformersConfigForToml) (types.ExporterMetaData, error) {
```
elizabethengelman, line 54: Updated the tests so that these pieces of functionality are tested through the one exported metohd, `ParseCurrentConfig`. However, this meant that for the error cases, I needed to create new toml files to make sure that the parser behaved properly in those error cases. Totally up for some other ideas on how to handle this!

generators/new_collateral/config/parser_test.go
```diff
@@ -34,107 +33,55 @@ var _ = Describe("Config Parser", func() {
...
-				},
-			}
-			_, parseErr := config.ParseTransformerExporters(tomlConfig)
+			// the exporter.cat_v1_1_0 will not properly decode into a types.TransformerExporter because the Path field doesn't match
+			configFile := "testConfigWithBadTransformerExporter"
+			_, parseErr := configParser.ParseCurrentConfig(testConfigFilePath, configFile)
 			Expect(parseErr).To(HaveOccurred())
-			Expect(parseErr.Error()).To(MatchRegexp("'Path' expected type 'string', got unconvertible type 'int'"))
+			Expect(parseErr).To(MatchError(
+				"error parsing transformer exporters from config file: error decoding transformerExporters:" +
```
elizabethengelman: I was struggling with how best to test these wrapped errors - would love some help figuring out a better way to test them!

rmulhol: This is maybe kind of cheating and idk if we may want to match against the nested error, but if not - maybe we could create an error to wrap the description and then match against that in the tests, e.g.

in the source:
```
var ErrParsingConfig = errors.Net("error parsing transformer exporters from config file")
fmt.Errorf("%w: %s", ErrParsingConfig, transformerExportersErr.Error())
```

in the tests:
```
Expect(parseErr).To(MatchError(ErrParsingConfig))
```

After some quick reading, it doesn't seem like there's a simple way to wrap multiple errors with `fmt.Errorf` and `%w` :/

generators/new_collateral/config/parser.go
yaoandrew: I think this makes sense and I can't think of a more elegant way to handle the tests. Still readable and makes sense with the explicit toml files.

yaoandrew (approved):
Looks good. Wrapped Errors for the win! 🏅

generators/new_collateral/config/parser_test.go
elizabethengelman: Good idea - made the change in ecd4f99

PR #362, opened 2020-11-05
Vdb 1657 test storage initializers

yaoandrew:
Removes some old collateral types - `dgd`, `gnt`, `omg`, and `rep`.


plugins/execute/transformerExporter.go
```diff
@@ -133,126 +133,126 @@ var Exporter exporter
...
+			vat_slip.EventTransformerInitializer,
+			vat_suck.EventTransformerInitializer,
+			vow_fess.EventTransformerInitializer,
+			vow_file_auction_address.EventTransformerInitializer,
+			vow_file_auction_attributes.EventTransformerInitializer,
+			vow_flog.EventTransformerInitializer,
+			vow_heal.EventTransformerInitializer,
+			yank.EventTransformerInitializer,
+		},
+		[]storage.TransformerInitializer{median_bat.StorageTransformerInitializer,
```
rmulhol: I _think_ this should be resolved in a rebase, but would be cool to have `median_bat` on a newline + alphabetized with the other storage transformers (and then hopefully the whitespace isn't a diff)

transformers/storage/median/initializers/median_comp/initializer.go
```diff
@@ -1,4 +1,4 @@
-package median_bat
+package median_comp
```
rmulhol: Niiiice catch 👍

rmulhol (approved):
LGTM after rebase 👍

PR #364, opened 2020-11-05
remove gomoderator files and command in travis

yaoandrew:
This should remove `gomoderator` from travis and remove the docker files. I don't think we can remove python as it seems like the `awscli` uses that.


rmulhol (approved):
~I think there might be a reference in the docs that we want to remove, otherwise~ LGTM!

(no need to remove references in the docs if we're keeping the script around to run outside of the context of CI)

elizabethengelman (approved):
👍

PR #365, opened 2020-11-05
Added time_ilk_snapshot and time_frob_totals

hexonaut:
This is a new PR to start from the `beta` branch, and which merges together these previous two PRs: https://github.com/makerdao/vdb-mcd-transformers/pull/295 and https://github.com/makerdao/vdb-mcd-transformers/pull/202


hexonaut:
Hey all. Could I request a prioritization of this PR? It is a very simple addition, and will provide me huge benefit to be able to switch over to the beta version of the data api for mkranalytics.com which is currently broken due to the chain split a couple weeks ago.

elizabethengelman:
Hey @hexonaut, I'm taking a look at this today.

db/migrations/20201007081009_create_time_frob_query.sql
```diff
@@ -0,0 +1,58 @@
...
+END;
+$$
+    LANGUAGE plpgsql
+    STRICT
+    STABLE;
+-- +goose StatementEnd
+
+-- +goose Down
+DROP FUNCTION api.time_frob_totals(TEXT, TIMESTAMP, TIMESTAMP, INTERVAL);
+DROP TYPE api.time_frob_total CASCADE;
```
elizabethengelman, line 70: would it be possible to also add a test for this frob query?

db/migrations/20201007080108_create_time_ilk_snapshot_query.sql
```diff
@@ -0,0 +1,86 @@
...
+);
+
+-- +goose StatementBegin
+CREATE FUNCTION api.time_ilk_snapshots(ilk_identifier TEXT, range_start TIMESTAMP, range_end TIMESTAMP, bucket_interval INTERVAL DEFAULT '1 day'::INTERVAL)
+    RETURNS SETOF api.time_ilk_snapshot AS
+$$
+DECLARE
+    r api.time_ilk_snapshot%rowtype;
+BEGIN
+    ASSERT EXTRACT(EPOCH FROM (range_end - range_start)) / EXTRACT(EPOCH FROM bucket_interval) <= 1000, 'Please limit requests to at most 1000 buckets.';
```
elizabethengelman, line 35: I believe that we're limiting the auction time series queries to 100 buckets - just curious as to why this is bumped up to 1000 buckets?

db/migrations/20201007081009_create_time_frob_query.sql
```diff
@@ -0,0 +1,58 @@
...
+);
+
+-- +goose StatementBegin
+CREATE FUNCTION api.time_frob_totals(ilk_identifier TEXT, range_start TIMESTAMP, range_end TIMESTAMP, bucket_interval INTERVAL DEFAULT '1 day'::INTERVAL)
+    RETURNS SETOF api.time_frob_total AS
+$$
+DECLARE
+    r api.time_frob_total%rowtype;
+BEGIN
+    ASSERT EXTRACT(EPOCH FROM (range_end - range_start)) / EXTRACT(EPOCH FROM bucket_interval) <= 1000, 'Please limit requests to at most 1000 buckets.';
```
elizabethengelman, line 23: Same question here w/r/t the max number of buckets.

db/migrations/20201007081009_create_time_frob_query.sql
```diff
@@ -0,0 +1,58 @@
...
+    
+    FOR r IN 
+        SELECT TO_TIMESTAMP(EXTRACT(EPOCH FROM range_start) + ROUND((headers.block_timestamp - EXTRACT(EPOCH FROM range_start)) / EXTRACT(EPOCH FROM bucket_interval)) * EXTRACT(EPOCH FROM bucket_interval)) AS bucket_start,
```
elizabethengelman: Just curious why you're using a different approach to getting `bucket_start` and `bucket_end` in this query vs in the ilk query. I find using `generate_series` a bit easier to read, but I also may be missing some nuanced difference between the two.

db/migrations/20201007081009_create_time_frob_query.sql
```diff
@@ -0,0 +1,58 @@
...
+            COALESCE(SUM(GREATEST(-dink, 0)), 0) AS free,
+            COALESCE(SUM(GREATEST(dart, 0)), 0) AS draw,
+            COALESCE(SUM(GREATEST(-dart, 0)), 0) AS wipe
+        FROM maker.vat_frob
+            LEFT JOIN public.headers ON (vat_frob.header_id = headers.id)
+            LEFT JOIN maker.urns ON (urns.id = urn_id)
+            LEFT JOIN maker.ilks ON (ilks.id = ilk_id)
+            WHERE ilks.identifier = time_frob_totals.ilk_identifier AND
+                headers.block_timestamp >= EXTRACT(EPOCH FROM range_start) and
+                headers.block_timestamp < EXTRACT(EPOCH FROM range_end)
```
elizabethengelman: Should these be comparing the `block_timestamp` to `bucket_start`/`bucket_end` instead of `range_start`/`range_end`?

elizabethengelman (commented):
These queries are both looking good! Just a few questions, also it would be great if we could add unit tests for the time frob query.

db/migrations/20201007080108_create_time_ilk_snapshot_query.sql
hexonaut: I actually want to increase the amount on the auctions as well. Originally I thought 100 was sufficient, but I've been getting more examples of situations where people are looking for more buckets on mkranalytics.com. I could just code this in on the front end, but the 100 I picked originally was kind of arbitrary, so I figured it would be easier overall to increase it instead. The performance still seems fine even at 1000.

db/migrations/20201007081009_create_time_frob_query.sql
hexonaut: The problem is that due to the joined filtering, the query became unusably slow for anything beyond a few buckets. I did some analysis, and it appeared to be having to scan the table over and over again for each bucket. I figured it was easier to just do 1 loop over the entire range and bucket the results as you go. The result is the same, but the query is much much faster.

db/migrations/20201007081009_create_time_frob_query.sql
hexonaut: No this is correct because it is bucketing the results as it moves along the whole range instead of joining the filtered results bucket by bucket.

hexonaut:
I added a test for the frob totals, and responded to your questions/comments. Let me know if there are any more issues. Thanks!

db/migrations/20201007080108_create_time_ilk_snapshot_query.sql
elizabethengelman: 👍  cool, that makes sense.

I'd like to take a peak at the current performance of the auction queries before we merge a an increase to the bucket limits, just so we can compare before and after that change is made. But we can get a PR up for that change - feel free to open one yourself, or I can open one later today/tomorrow.

hexonaut: Sure, I can take care of that. Just running some test queries on my own database I'm seeing ~200ms to return 100 results and ~2000ms to return 1000 results on the flip totals. I'm not sure what your targets are, but it looks like the performance scales approximately linearly with the number of buckets.

hexonaut: I've opened a PR: https://github.com/makerdao/vdb-mcd-transformers/pull/393

hexonaut:
Closing this as it was re-opened as another PR.

PR #366, opened 2020-11-05
Vdb 1672 port collateral generator over

elizabethengelman:
I think that this build should pass once https://github.com/makerdao/vulcanizedb/pull/148 gets merged in. Wondering if there's a better way to handle this sort of change.


rmulhol (approved):
started going through this line by line but then figured we probably don't want to make any big changes over here that we're not making in prod so that future changes are easy to apply to both branches - 👍 from me to merge, unless there are important differences between this and prod that I should dig into

paytonrules:
Does this need to go in Oasis too? Oasis doesn't have extract-diffs so I'm guessing not but....

elizabethengelman:
@rmulhol the only main change between what is in `prod` and this PR is removing the schema field in b439b4273fa2b54103a0fb0df471424cbe50f322. The rest of the PR should be the same as is in prod now.

@paytonrules yep, I don't think we need to add this to Oasis since we're not currently watching Oasis storage, also I don't think that the Oasis contracts are collateral specific.

PR #368, opened 2020-11-09
Add BAL_A and YFI_A collaterals

paytonrules:
Used the generator tool for this.


elizabethengelman (commented):
This is looking good! I think we'll also want to manually add a couple things:
- integration tests - I think we can probably do just rely events for flip, osm and median for now, since every contract should have a rely event, and that will be a good way to make sure that things are wired up correctly.
- add the new contracts to [FlipV110ABI](https://github.com/makerdao/vdb-mcd-transformers/blob/prod/transformers/shared/constants/method.go#L53), [MedianABI](https://github.com/makerdao/vdb-mcd-transformers/blob/prod/transformers/shared/constants/method.go#L79) and [OsmABI](https://github.com/makerdao/vdb-mcd-transformers/blob/prod/transformers/shared/constants/method.go#L93)
- add the new contracts to [test_data/config_values.go](https://github.com/makerdao/vdb-mcd-transformers/blob/prod/transformers/test_data/config_values.go)

paytonrules:
The ABI's here are wrong. Will close and reopen.

paytonrules:
Fixed up a bunch including adding integration tests.

PR #371, opened 2020-11-09
On Beta the paths for constants are different

paytonrules:
This should not be merged into prod - in prod the generator uses the new paths. But to use this in Beta we have to fix up the paths.


paytonrules:
I tested this manually but generating a collateral with it and making sure it could build the execute container.

elizabethengelman (approved):
👌

PR #372, opened 2020-11-09
Add YFI_A and BAL_A collateral to prod

paytonrules:


paytonrules:
This is littered with errors, will close and reopen.

environments/extractDiffs.toml
```diff
@@ -215,7 +227,59 @@
...
         deployed = 10322821
+    [contract.OSM_BAL]
```
rmulhol: Don't think we want these OSM contracts in here - not transforming storage from them, and they're fairly noisy since they're regularly providing price updates

environments/mcdTransformers.toml
```diff
@@ -3,167 +3,9 @@
...
...flop_v1_0_1", "flop_v1_0_9", "jug", "median_bat", "median_comp", "median_eth", "median_knc", "median_link", "median_lrc", "median_mana", "median_usdt", "median_wbtc", "median_zrx", "pot", "spot", "vat", "vow", "auction_file", "bite", "cat_claw", "cat_file_box", "cat_file_chop_lump_dunk", "cat_file_flip", "cat_file_vow", "deal", "dent", "deny", "flap_kick", "flip_file_cat", "flip_kick", "flop_kick", "jug_drip", "jug_file_base", "jug_file_ilk", "jug_file_vow", "jug_init", "log_median_price", "log_value", "median_diss_batch", "median_diss_single", "median_drop", "median_kiss_batch", "median_kiss_single", "median_lift", "new_cdp", "osm_change", "pot_cage", "pot_drip", "pot_exit", "pot_file_dsr", "pot_file_vow", "pot_join", "rely", "spot_file_mat", "spot_file_par", "spot_file_pip", "spot_poke", "tend", "tick", "vat_deny", "vat_file_debt_ceiling", "vat_file_ilk", "vat_flux", "vat_fold", "vat_fork", "vat_frob", "vat_grab", "vat_heal", "vat_hope", "vat_init", "vat_move", "vat_nope", "vat_rely", "vat_slip", "vat_suck", "vow_fess", "vow_file_auction_attributes", "vow_file_auction_address", "vow_flog", "vow_heal", "yank", "flip_bal_a_v1_1_14", "median_bal", "flip_yfi_a_v1_1_14", "median_yfi"]
```
rmulhol: I'm pretty torn about this diff. On the one hand, I like having these vals sorted/on new lines. On the other hand, that seems like a massive waste of time if we have to do it every time we run the generator. Wonder if it's worth creating a story for creating a script to format our toml (at a glance, doesn't seem like there's a widely used tool for doing this)

rmulhol (approved):
Would prob purge the OSM contracts from the extractDiffs config, but otherwise LGTM 👍

environments/extractDiffs.toml
paytonrules: Happy to take them out, but we should probably find a way to make this more intuitive in the code. There's no real way to tell which contracts go where.

environments/mcdTransformers.toml
paytonrules: I mentioned earlier - but I'd like to see them put in alphabetical order. Then you'd only have a diff on the new entries. I think that'd probably solve the issue.

In either case this should get fixed at the generator.

environments/extractDiffs.toml
rmulhol: Agreed - got any ideas? Best I can come up with is comparing against the storage transformers :/

environments/mcdTransformers.toml
rmulhol: Makes sense - do you know if we have a story for that? Are you thinking leave as is for now and then after we update and run the generator that it should take care of this?

environments/extractDiffs.toml
paytonrules: I don't know that I understand how these processes work enough to have clear suggestions. I have questions thuogh.

If I understand this right for execute (and backfills) we use the generated TransformerExporter to determine what to transform - it's a loop. So if a storage contract we don't care about is in the contracts section of the config, it won't matter.  This is why those are empty for the backfills unless we explicitly want to do a backfill.

But extract-diffs will extract diffs based on what's in the contract section. So for execute contracts is just a lookup for info, but for extract-diffs it actually determines what diffs to extract in addition. 

If I have that right we could have an explicit list of what to listen for in the config. It'll have some duplication, but then we could pull the contract section out.

environments/mcdTransformers.toml
paytonrules: I was thinking I'd leave it as-is for now and we'd fix it in the generator. Not sure if we have a story, I'll have to check. I don't really feel strongly about how we fix it, whatever is easiest to just make clean to read diffs.

environments/extractDiffs.toml
rmulhol: I like it! For reference - this is the code that grabs the addresses from this config file for the `extractDiffs` process: https://github.com/makerdao/vulcanizedb/blob/prod/libraries/shared/streamer/filter_query.go#L31

PR #375, opened 2020-11-12
Vdb 1700 use new geth patch

elizabethengelman:
To test out the extract-diffs docker image, we'll need to make sure to pass in the vdb branch `vdb-1700-update-geth-in-beta` as the `VDB_VERSION`.

~~Note: the last commit, dbf5335, should be removed before this PR is merged in. I manually changed the vdb version that the `execute` and `gomoderator` are using so that I could make sure that the build passed. But, once https://github.com/makerdao/vulcanizedb/pull/151 is merged in, that change will no longer be necessary.~~


.travis/deploy.sh
```diff
@@ -46,6 +49,9 @@ docker push makerdao/vdb-backfill-storage:$TAG
 message PUSHING BACKFILL-EVENTS DOCKER IMAGE
 docker push makerdao/vdb-backfill-events:$TAG
 
+message PUSHING EXTRACT-DIFFS DOCKER IMAGE
+docker push makerdao/vdb-extract-diffs:$TAG
```
rmulhol, line 53: I wonder if we might also need to add a deployment script to this file below

dockerfiles/extract_diffs/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+RUN apk --update --no-cache add g++ git linux-headers
+
+# Build migration tool
+WORKDIR /go
+RUN GO111MODULE=auto go get -u -d github.com/pressly/goose/cmd/goose
+WORKDIR /go/src/github.com/pressly/goose/cmd/goose
+RUN GO111MODULE=auto go build -a -ldflags '-s' -tags='no_mysql no_sqlite' -o goose
+
+ARG VDB_VERSION=staging
```
rmulhol: should this be `beta`?

dockerfiles/extract_diffs/startup_script.sh
```diff
@@ -0,0 +1,41 @@
...
+  if [ $? -ne 0 ]; then
+    exit 1
+  fi
+
+  # Construct the connection string for postgres
+  VDB_PG_CONNECT=postgresql://$DATABASE_USER:$DATABASE_PASSWORD@$DATABASE_HOSTNAME:$DATABASE_PORT/$DATABASE_NAME?sslmode=disable
+fi
+
+# Run the DB migrations
+echo "Connecting with: $VDB_PG_CONNECT"
```
rmulhol: prob want to remove this log

environments/extractDiffs.toml
```diff
@@ -0,0 +1,269 @@
...
...ame":"arg1","type":"bytes32"},{"indexed":true,"internalType":"bytes32","name":"arg2","type":"bytes32"},{"indexed":false,"internalType":"bytes","name":"data","type":"bytes"}],"name":"LogNote","type":"event"},{"constant":true,"inputs":[],"name":"age","outputs":[{"internalType":"uint32","name":"","type":"uint32"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[],"name":"bar","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":true,"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"bud","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},{"constant":false,"inputs":[{"internalType":"address","name":"usr","type":"address"}],"name":"deny","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":false,"inputs":[{"internalType":"address[]","name":"a","type":"address[]"}],"name":"diss","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":false,"inputs":[{"internalType":"address","name":"a","type":"addre
```
rmulhol: I think we can delete everything from here down

rmulhol (commented):
Looking good!

A few questions below

dockerfiles/extract_diffs/Dockerfile
elizabethengelman: yikes, nice catch!

.travis/deploy.sh
elizabethengelman: I'm also wondering if it would be good to remove the `if [ "$ENVIRONMENT" == "prod" ]; then` line here, since this is the beta branch. And then remove `elif [ "$ENVIRONMENT" == "staging" ]; then` from the prod branch?

rmulhol: I like this idea! 💯

elizabethengelman: great, i'll plan to do that new PRs

paytonrules:
The vulcanizedb patch has been merged, so you shouldn't need to set VDB_VERSION anymore when testing.

rmulhol (approved):

PR #376, opened 2020-11-16
[Prod] Version median contracts

paytonrules:
Tie the version of each median contract to the initial deployment of it's corresponding FLIP contract.


generators/new_collateral/test_data/test_data.go
```diff
@@ -312,7 +312,7 @@ var (
     repository = "github.com/makerdao/vdb-mcd-transformers"
     type = "eth_storage"
   [exporter.log_median_price]
-    contracts = ["MEDIAN_BAT", "MEDIAN_ETH_B"]
+    contracts = ["MEDIAN_BAT_1_0_0", "MEDIAN_ETH_B"]
```
rmulhol, line 315: Not added by this PR so no need to address, but since you brought it up earlier - do you think we should remove `"MEDIAN_ETH_B"`? Seems like it could be cool to say that after this PR, every median contract name includes a version/matches something that actually exists in a config file

transformers/integration_tests/median_lift_test.go
```diff
@@ -91,7 +91,7 @@ var _ = Describe("MedianLift EventTransformer", func() {
 		header, headerErr := persistHeader(db, blockNumber, blockChain)
 		Expect(headerErr).NotTo(HaveOccurred())
 
-		medianWbtcAddress := constants.GetContractAddress("MEDIAN_WBTC")
+		medianWbtcAddress := constants.GetContractAddress("MEDIAN_WBTC_1_0_6")
```
rmulhol: maybe worth using `test_data.MedianWbtcAddress()` so we don't need to make updates here to change the contract name?

rmulhol (approved):
LGTM! A couple questions but nothing merge blocking

generators/new_collateral/test_data/test_data.go
paytonrules: So we can't - yet - in test_data. I got burned by this.

The collateral generator currently creates MEDIAN and OSM contracts automatically - using the FLIP name - and it doesn't rip off the `_A` or `_B` because it can't be sure if that needs to come off. So if you rip this out you make a bunch of tests fail. Then you get confused and start swearing. Well that's what I did anyway.

My current thought is the collateral generator should ask you "the median contract will be named <blah> by default, would you like to change it?" Not sure if @elizabethengelman has better ideas.

transformers/integration_tests/median_lift_test.go
paytonrules: 👍 - should do more of that probably TBH. There's a lot of string duplication.

generators/new_collateral/test_data/test_data.go
paytonrules: I went ahead and added a story for this https://makerdao.atlassian.net/browse/VDB-1704

rmulhol:
Does the target branch needs to be switched over to `prod` now?

paytonrules:
> Does the target branch needs to be switched over to `prod` now?

It does - and it is. And it always has been. Pay no attention to the man behind the curtain....

PR #381, opened 2020-11-18
Remove "if prod" from travis deployment since this is the beta branch

elizabethengelman:


rmulhol:
I rolled this into #379 but happy to take it out of there if you think we should keep this separate?

rmulhol:
aaaand looks like I missed the line above the service deploy (the one about setting `$ENVIRONMENT`) 😅

elizabethengelman:
closing this since this change was merged into #379

PR #384, opened 2020-11-19
Vdb 1688 gh action extract diffs

yaoandrew:
This adds to Eric's github action and uses his python :snake: magic to verify that the `extract_diffs` container is 'healthy'. 
It runs as a separate job so the container clean up happens between the `execute` and` extract_diffs` tests and enables the jobs to run in parallel. A new secret was added to support a `wss` vs `http` path for geth when running `extract_diffs`.

as a side note, trying to run [github actions locally](https://github.com/nektos/act/) fails when using this repo due to this [issue](https://github.com/nektos/act/issues/418)

### DIFF
--- a/.github/workflows/check_vdb_command
+++ b/.github/workflows/check_vdb_command
@@ -9,7 +9,7 @@ parser.add_argument("--database_name", help="name of the postgres database", def
 parser.add_argument("--database_port", help="port of the postgres database", default="5432")
 parser.add_argument("--database_user", help="username of the postgres database user", default="postgres")
 parser.add_argument("--database_password", help="password of the postgres database user", default="postgres")
-parser.add_argument("--storagediffs_source", help="geth version", default="new-geth")
+parser.add_argument("--storagediffs_source", help="geth version", default="geth")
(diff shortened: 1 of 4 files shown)


.github/workflows/check_vdb_command
```diff
@@ -9,6 +9,7 @@ parser.add_argument("--database_name", help="name of the postgres database", def
 parser.add_argument("--database_port", help="port of the postgres database", default="5432")
 parser.add_argument("--database_user", help="username of the postgres database user", default="postgres")
 parser.add_argument("--database_password", help="password of the postgres database user", default="postgres")
+parser.add_argument("--storagediffs_source", help="geth version", default="new-geth")
```
paytonrules: I"m still kinda confused how this `new-geth` value works.  There's no chance we're inadvertently running the csv version of extract-diffs is there?

paytonrules (approved):
Just one real question.

.github/workflows/check_vdb_command
yaoandrew: You're right.. this was causing it to default to csv and the container would report that it was healthy and waiting for a csv file to appear. Changed back to `geth` and added a `wss://` path to the node.

PR #385, opened 2020-11-19
Log Errors when ValidateMigrations cannot parse the response

paytonrules:
Sometimes Github is erroring on unmarshal, so turn it into a straight string for debugging.

In addition, only run validatemigrations on a PR. This will help with occasional API throttling we run into with github.

### DIFF
--- a/cmd/checkMigrations.go
+++ b/cmd/checkMigrations.go
@@ -114,7 +114,7 @@ func getGithubFileNames() ([]string, error) {
 	jsonErr := json.Unmarshal(body, &files)
 
 	if jsonErr != nil {
-		return []string{}, fmt.Errorf("Failed to unmarshal %s, %w", string(body), jsonErr)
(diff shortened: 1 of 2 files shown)


cmd/checkMigrations.go
```diff
@@ -114,7 +114,7 @@ func getGithubFileNames() ([]string, error) {
 	jsonErr := json.Unmarshal(body, &files)
 
 	if jsonErr != nil {
-		return []string{}, fmt.Errorf(errorContext, jsonErr)
+		return []string{}, fmt.Errorf("Failed to unmarshal %s, %w", string(body), jsonErr)
```
rmulhol: super minor but I think technically we want to avoid capitalization in error messages: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

rmulhol (approved):

cmd/checkMigrations.go
paytonrules: Bah humbug. Fixed.

rmulhol (approved):

PR #386, opened 2020-11-23
Update docs with an image and some small corrections

paytonrules:


transformers/events/DOCUMENTATION.md
```diff
@@ -2,11 +2,13 @@
 
 ## Architecture
 
-Transformers fetch logs from Ethereum, convert/decode them into usable data, and then persist them in postgres. A transformer converts the raw chain data into a human friendly representation suitable for consumption in the API.
```
rmulhol, line 5: Wondering if we might want to keep something stating that configuring event transformers in execute causes us to query the ethereum node for those logs and write them to the `public.event_logs` table (in addition to scanning that table and transforming them)

transformers/events/DOCUMENTATION.md
```diff
@@ -29,28 +31,27 @@ If the contract isn't already present in the environment you'll need to add it b
...
 To create a custom transformer you will need to create a new `Transformer`
 struct with a `ToModels` method on it which converts a `core.EventLog` object
 (the raw, untransformed data) to an `event.InsertionModel` (the domain object).
 
-Note that for this step you may _not_ need to create the database
-migration, because to write the transformer you don't need to save the transformed
+Much of the time it is possible to complete this step without creating the database
+migration, because to write the transformer you don't actually save the transformed
 object to the database. You may be saving addresses, which
 go into the already existing address table and are referenced by a foreign key.
```
rmulhol, line 51: Hm, I'm confused by this line. I think every transformer writes to a new table? Even if the table is functionally just a join table of `log_id` + `address_id`, I think it's normally declared on a table matching the event name

transformers/events/DOCUMENTATION.md
```diff
@@ -168,3 +169,9 @@ In the event there are not logs for an event you're looking to transform in ethe
 ## Useful Documents
 
 [Ethereum Event ABI Specification](https://solidity.readthedocs.io/en/develop/abi-spec.html#events)
+
+## Diagram
+
+This shows the entire process, from the perspective of the `execute` process. Writing transformers means writing code for the little loop near the end.
+
+![Transformer Sequence Diagram](./Transforming Events.svg)
```
rmulhol: do we need to include this file in this PR?

transformers/events/DOCUMENTATION.md
paytonrules: I did not know it did that - yes it deserves a mention.

transformers/events/DOCUMENTATION.md
paytonrules: You don't have to do it _yet_ to make the test pass. You'll need it for the integration test.

transformers/events/DOCUMENTATION.md
paytonrules: Only if we want it to show up.

transformers/events/DOCUMENTATION.md
paytonrules: I'm not sure what happened with the top here, I basically restored the top paragraph as I like it better. Blame turkey.

transformers/events/DOCUMENTATION.md
rmulhol: I think maybe I'm thrown off by the line about "You may be saving an address" - the reason you don't need a migration isn't because of the content of the table, it's because the unit tests for individual transformers don't exercise writing the new record to the DB (that only happens in integration tests), right?

rmulhol (approved):
Good with this as is, definite improvement over what we had before 👍 

Still think it's a little confusing to imply folks don't need a migration to add a new table for the new transformed event, but we can always shore that up later

transformers/events/DOCUMENTATION.md
paytonrules: That's correct - also that's not new. It's how we've phrased it for a while.

PR #387, opened 2020-11-23
Remove eth b backfill

rmulhol:


dockerfiles/execute/startup_script.sh
```diff
@@ -7,6 +7,6 @@ if test -n "$DIFF_BLOCK_FROM_HEAD_OF_CHAIN"; then
     DIFF_STARTING_BLOCK=$DIFF_BLOCK_FROM_HEAD_OF_CHAIN
 fi
 
-echo "Starting transformer execution..."
+echo "Starting transformer execution from $DIFF_STARTING_BLOCK blocks back"
```
paytonrules, line 10: Is this gonna show the actual starting block?  Or will it show the string `$DIFF_STARTING_BLOCK`.

paytonrules (approved):

dockerfiles/execute/startup_script.sh
rmulhol: hopefully the actual number that's configured (which technically isn't the starting block, but rather the number of how far back from the starting block we should scan)

PR #388, opened 2020-11-30
[DO NOT MERGE] Rescan last 200000 blocks for diffs

rmulhol:
- diffs were not transformed between blocks 11,234,894 and 11,315,955
- these diffs aren't picked up by the existing execute process since it
  only looks at diffs within 1250 blocks of the head
- current head is ~11,360,000, so scanning within 200,000 or head
  should cover all that were missed
- only need storage transformers since existing execute process will
  handle events

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ before_install:
 - docker build -f dockerfiles/gomoderator/Dockerfile .
 script:
 - set -e
-- scripts/check_config.sh
 - env GO111MODULE=on make validatemigrationorder
 - env GO111MODULE=on make test
 - env GO111MODULE=on make integrationtest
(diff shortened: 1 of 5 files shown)


yaoandrew (approved):

elizabethengelman (approved):

.travis.yml
```diff
@@ -16,7 +16,6 @@ before_install:
 - docker build -f dockerfiles/gomoderator/Dockerfile .
 script:
 - set -e
-- scripts/check_config.sh
```
elizabethengelman, line 19: Just curious why we're removing this here? Not a blocker, just trying to remind myself what this does.

rmulhol: this verifies that the transformers in the exporters match the transformers in the config. In this case, there's a conflict because the config includes some event transformers that aren't in the exporter. The reason for the discrepancy is that those event transformers need to be in the config for the plugin integration tests to run/pass, but we only want to run storage transformers with this branch

rmulhol:
Closing because this branch should not be merged (it's mostly useful for transforming old diffs that fell out of the `$DIFF_BLOCK_FROM_HEAD_OF_CHAIN` window, but shouldn't be deployed to the main `execute` process that's tracking the head of the chain)

PR #389, opened 2020-12-02
(VDB-1712) Add GUSD transformers to execute [beta]

rmulhol:


environments/extractDiffs.toml
```diff
@@ -52,9 +52,13 @@
...
         deployed = 11048869
+    [contract.MCD_FLIP_GUSD_A_1_1_5]
```
rmulhol: Just remembered that this environment now uses this config to setup the diffs subscription with Geth, so maybe we want to remove this address until the backfill events process has completed (so that we can generate keys for diffs)?

environments/extractDiffs.toml
```diff
@@ -227,6 +231,10 @@
...
         deployed = 8925077
+    [contract.MEDIAN_YFI_1_1_14]
```
rmulhol: Also noticing that it appears this address wasn't in this config or the backfill configs - I wonder if we might want to include it in the backfill events transformers' config (and later add it to the backfill storage process)?

environments/backfillEvents.toml
```diff
@@ -2,8 +2,89 @@
...
+        "deal",
+        "dent",
+        "deny",
+        "flip_file_cat",
+        "flip_kick",
+        "rely",
+        "tend",
+        "tick",
+        "yank",
+    ]
```
rmulhol: Does adding this step need to happen in a separate PR? Just realized that the backfill events startup script expects an ending block number, but I guess we wouldn't have that until we've confirmed that we've deployed an execute process that is tracking the new address?

environments/backfillEvents.toml
```diff
@@ -2,8 +2,89 @@
     home     = "github.com/makerdao/vulcanizedb"
     name     = "transformerExporter"
     save     = true
-    transformerNames = []
+    transformerNames = [
```
elizabethengelman: I wonder if we should hold off updating backfillEvents yet, so that we can also update the dockerfile startup script's ending block number to be about when this is merged in?

elizabethengelman: Actually, now that I think about it, this is probably okay since the current ending block number is before the GUSD  flip contract's deployment block.

environments/extractDiffs.toml
elizabethengelman: Nice catch! I like that idea. Maybe we should create a story for GUSD storage backfill with a note about adding this address back to extractDiffs so we don't forget? I feel like I'm starting to have a hard time keeping straight which of the new collaterals need to be backfilled, etc. Especially since I think we may have taken care of one of the collateral's storage backfills when we spun up geth4?

environments/extractDiffs.toml
elizabethengelman: I think we have stories for backfilling [BAL, YFI](https://makerdao.atlassian.net/browse/VDB-1697) and [ETH-B.](https://makerdao.atlassian.net/browse/VDB-1696) I feel like prioritizing those would be good. 

Also I wonder if we could combine them all into one update?

elizabethengelman (approved):
LGTM! 

I think we've discussed most of the backfill-related comments during IPM, so feel free to ignore anything that we've already talked about.

environments/backfillEvents.toml
elizabethengelman: I missed this comment earlier somehow. Yeah, you're right that we wouldn't have the ending block number until we merge this PR in. I _think_ that if this config change stays in, then the backfill process would try to start up, but wouldn't do anything because it would try to backfill from the GUSD deployment block to the currently configured ending block, which is 11108609 and before GUSD is deployed. Not sure if that would cause an error or not though 🤔

PR #391, opened 2020-12-02
VDB-1409/VDB-1656 time series queries

elizabethengelman:
This was cut from #365. The only changes are that I updated the timestamp for the `create_time_ilk_snapshot_query` migration, so that the migrations would be run in order, and checked in the updated schema.


yaoandrew (approved):
LGTM. I took a minute to go through the original comments on the older PRs to get a little more context.

PR #393, opened 2020-12-03
Adjust time series auctions bucket limit to 1000

hexonaut:
The bucket limit was set kind of arbitrarily based on my first best guess, but I've been getting feedback that the limit is too low. Running my own tests shows an increase of about ~200ms query response with 100 buckets to ~2000ms with 1000 buckets. This seems reasonable to me, but I am uncertain about how you are measuring performance targets.


yaoandrew:
Cherry picked this commit onto https://github.com/makerdao/vdb-mcd-transformers/pull/391

PR #394, opened 2020-12-03
Don't log connection string on startup

rmulhol:


paytonrules (approved):
Haven't we fixed this like 10 times.

PR #396, opened 2020-12-04
(VDB-1726) Create urn IDs for src & dst in fork [prod]

rmulhol:
- fork creates/updates urns for the passed ilk + src/dst fields. If we
  don't write them to the urns table, they won't be available to the
  storage transformers parsing urn diffs - and diffs will go
  unrecognized.


rmulhol:
Closing because the storage transformer should detect diffs from urns based off of a fork and create an urn ID at that time

PR #397, opened 2020-12-04
(VDB-1726) Create urn IDs for src & dst in fork [beta]

rmulhol:
- fork creates/updates urns for the passed ilk + src/dst fields. If we
  don't write them to the urns table, they won't be available to the
  storage transformers parsing urn diffs - and diffs will go
  unrecognized.

### DIFF
--- a/transformers/integration_tests/vat_fork_test.go
+++ b/transformers/integration_tests/vat_fork_test.go
@@ -30,24 +30,27 @@ import (
 )
 
 var _ = Describe("Vat fork transformer", func() {
+	var (
+		ilkID         int64
+		src           = "0x7939E55BE6A8CB8fceE57F409543E25489C06aaC"
+		dst           = "0xd175Dfd88939eE702CB69e96f8bedAa2f93FBFfA"
+		vatForkConfig = event.TransformerConfig{
+			TransformerName:   constants.VatForkTable,
+			ContractAddresses: []string{test_data.VatAddress()},
+			ContractAbi:       constants.VatABI(),
+			Topic:             constants.VatForkSignature(),
+		}
+	)
+
 	BeforeEach(func() {
 		test_config.CleanTestDB(db)
-	})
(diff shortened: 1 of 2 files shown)


rmulhol:
It turns out that in the storage transformer process, we already lookup possible urns based on entries in the `maker.urns` table _plus_ entries in the `maker.vat_fork` table: https://github.com/makerdao/vdb-mcd-transformers/blob/prod/transformers/storage/repository.go#L217
And then we get or create an urn ID when we transform a diff matching any of those entries: https://github.com/makerdao/vdb-mcd-transformers/blob/prod/transformers/storage/vat/repository.go#L219

One could argue that it would still be good to create an urn as soon as we see the fork event, but that seems far less urgent if we know it'll happen as soon as we detect an associated diff - so considering closing this

transformers/integration_tests/vat_fork_test.go
```diff
@@ -56,28 +59,44 @@ var _ = Describe("Vat fork transformer", func() {
...
 		Expect(dbResult.Ilk).To(Equal(ilkID))
-		Expect(dbResult.Src).To(Equal("0x7939E55BE6A8CB8fceE57F409543E25489C06aaC"))
-		Expect(dbResult.Dst).To(Equal("0xd175Dfd88939eE702CB69e96f8bedAa2f93FBFfA"))
+		Expect(dbResult.Src).To(Equal(src))
+		Expect(dbResult.Dst).To(Equal(dst))
 		Expect(dbResult.Dink).To(Equal("3248431462049897973"))
 		Expect(dbResult.Dart).To(Equal("218121873079553101113"))
 	})
+
+	It("creates urn IDs for src and dst fields in fork payload", func() {
```
yaoandrew, line 89: 💯

yaoandrew (approved):
LGTM if we decide we want to keep this logic.

PR #398, opened 2020-12-08
Switch backfill events base image to buster

rmulhol:
- Avoid alpine since it can hang on plugin.Open

Can confirm this resolves hanging when running locally, though it's not clear whether that hanging is happening on the deploy.

May want to consider using buster uniformly across all images if we're going to make this change.

Probably also would be good to add additional workflows and health checks to verify every container runs.


rmulhol:
Closing because it looks our deployment _is_ running, so we don't need an urgent fix and can approach this more deliberately/uniformly across containers

PR #399, opened 2020-12-09
Query new state diffs

paytonrules:
After going in circles - primarily with the testing. I ended up going with just returning the block heights for every new diff, in ascending order. TotalCount is provided by postgraphile anyway.

image

### DIFF
--- a/db/migrations/20201207151120_create_new_storage_block_heights.sql
+++ b/db/migrations/20201207151120_create_new_storage_block_heights.sql
@@ -1,6 +1,6 @@
 -- +goose Up
 
-CREATE OR REPLACE FUNCTION api.storage_transformation_status()
+CREATE OR REPLACE FUNCTION api.new_storage_block_heights()
     RETURNS SETOF BIGINT
     LANGUAGE sql
     STABLE
@@ -12,4 +12,4 @@ $$;
 
 -- +goose Down
 
-DROP FUNCTION api.storage_transformation_status();
+DROP FUNCTION api.new_storage_block_heights();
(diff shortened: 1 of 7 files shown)


transformers/component_tests/queries/new_storage_block_heights_test.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+var _ = Describe("New Storage Block Heights Query", func() {
+	const newStorageBlockHeightsQuery = `SELECT * FROM api.new_storage_block_heights()`
+
+	BeforeEach(func() {
+		test_config.CleanTestDB(db)
+	})
+
+	It("has a row for ever new storage diff", func() {
```
rmulhol: :%s/ever/every

db/migrations/20201207151120_create_new_storage_block_heights.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+
+CREATE OR REPLACE FUNCTION api.new_storage_block_heights()
```
rmulhol: wondering if we might want to name this something that makes the purpose more obvious for API users - my mind is going somewhere very verbose like `get_block_heights_for_new_untransformed_diffs()` but maybe there's a better middle ground.

Also def nbd as long as it serves the purpose for Ed

rmulhol (approved):

transformers/component_tests/queries/new_storage_block_heights_test.go
paytonrules: 👍

db/migrations/20201207151120_create_new_storage_block_heights.sql
paytonrules: I've gone through several names and hate them all equally. Yours is probably the "least bad".

PR #402, opened 2020-12-14
Vdb 1739 fix data generation

paytonrules:
This fixes 2 issues with the data generation.

1) Storage Diffs require an eth_node_id. This worked in tests because we injected a pg that happened to have a matching EthNode id but it failed in the real code. I adjusted the tests to make sure they don't create a node in this one case (because the generator itself creates a node).

2) Headers hard coded their eth-node id to 1, which works if you blew away your database first, but not otherwise.


rmulhol (approved):
LGTM as long as the build passes after the rebase

yaoandrew (approved):
🚢 LGTM

PR #403, opened 2020-12-15
Vdb 1741 add 1.2.0 collaterals to prod

elizabethengelman:


environments/extractDiffs.toml
```diff
@@ -87,6 +87,10 @@
...
         deployed = 10790528
+    [contract.MCD_FLIP_RENBTC_A_1_2_1]
```
elizabethengelman: I just realized that maybe we don't want to include these in here until the backfillEvents process has been completed, because otherwise we'll probably end up with some diffs that we can't transform yet? But on the other hand, the sooner we get this in, the less backfill we'll have to do for storage diff. 🤔

rmulhol (approved):
:shipit:

PR #405, opened 2020-12-15
Vdb 1738 all flips

elizabethengelman:
The `all_flips` query function uses on the `get_flip` method, and calls it for each bid_id for the given ilk. So, the fundamental change for this was to allow `get_flip` to take in an address, and then filter the returned flip by that address. So that I could do this more iteratively, for now I just created a new `get_flip_with_address` function to use in place of `get_flip`. Since `get_flip` is being used by a couple other functions besides `all_flips`, I also switched those out to use `get_flip_with_address`. Those other functions are: `bite_event_computed_columns` and `flip_bid_event_computed_columns`.

I also needed to make sure that the types that each of these functions returns includes the flip_address, so that the user could know which flip contract the result was associated with, or so that other queries that used those functions could also filter on address. The types that changed were:
- `api.flip_bid_snapshot` got a flip_address attribute
- `api.bite_event` got a flip_address attribute, which also required that the function `api.urn_bites` be updated to return the flip_address so that its returning type could be coerced into a `api.bite_event`

Hope that helps make this a little bit less confusing 😅 .

- [x] update `get_flip` query to account for more than one address per ilk (created a temporary new query `get_flip_with_address` - planning to use this one instead of the old `get_flip` going forward)
- [x] use `get_flip_with_address` in `all_flips` query
- [x] use `get_flip_with_address` in `bite_event_computed_columns` query
- [x] use `get_flip_with_address` in `flip_bid_event_computed_column` query
- [x] remove the original `get_flip` query in lieu of the new `get_flip_with_address` query?
- [x] identify any other queries that need to be fixed (e.g. flop? flap?): added a [new story](https://makerdao.atlassian.net/browse/VDB-1783) for this investigation


db/migrations/20201214160608_update_flip_bid_snapshot.sql
```diff
@@ -0,0 +1,149 @@
...
+-- add a flip_address attribute to the flip_bid_snapshot type
+ALTER TYPE api.flip_bid_snapshot ADD ATTRIBUTE flip_address text;
+
+-- update the get_flip function to return a flip_address in it's flip_bid_snapshot
+CREATE OR REPLACE FUNCTION api.get_flip(bid_id NUMERIC, ilk TEXT, block_height BIGINT DEFAULT api.max_block())
+    RETURNS api.flip_bid_snapshot
+AS
+$$
+WITH ilk_ids AS (SELECT id FROM maker.ilks WHERE ilks.identifier = get_flip.ilk),
+     -- there should only ever be 1 address for a given ilk, which is why there's a LIMIT with no ORDER BY
```
rmulhol: Probably need to keep reading further down, but why wouldn't it be possible for there to be multiple addresses for a given ilk? Wouldn't like `MCD_FLIP_ETH_A_1_0_0` and `MCD_FLIP_ETH_A_1_0_9` both match on the `ilk` of `ETH_A` but exist as separate addresses on the `public.addresses` table?
Assuming that's the case, then I would think that the `address_id` we get back from `maker.flip_ilk` would be potentially non-deterministic

elizabethengelman: Oops, I think that this comment is copy-pasted from the initial implementation of `get_flip` - I believe that I changed `get_flip` to `get_flip_with_address`, and everywhere that was calling `get_flip` should now be updated to use the new function. I did it this way, with several migrations, so that I could make the change iteratively and so it would hopefully be a bit easier to review. But this is a good reminder that it would probably best to squash all (most?) of these migrations into one before merging.

db/migrations/20201214160609_update_get_flip_query.sql
```diff
@@ -0,0 +1,70 @@
+-- +goose Up
+
+-- add a new function that works the same as get_flip but takes into account a specific flip_address
+-- this is required because there are now more than one flip contract per ilk
+CREATE OR REPLACE FUNCTION api.get_flip_with_address(bid_id NUMERIC, flip_address TEXT, ilk TEXT, block_height BIGINT DEFAULT api.max_block())
+    RETURNS api.flip_bid_snapshot
+AS
+$$
+WITH ilk_ids AS (SELECT id FROM maker.ilks WHERE ilks.identifier = get_flip_with_address.ilk),
```
rmulhol: Wondering if it might make sense to name this `ilk_id`. Seems like we probably expect it to return only one value, and might be in trouble if it returns more

db/migrations/20201214160609_update_get_flip_query.sql
```diff
@@ -0,0 +1,70 @@
...
+     address_id AS (SELECT id FROM public.addresses WHERE address = get_flip_with_address.flip_address),
+     kicks AS (SELECT usr
```
rmulhol: Similar thing here, potentially renaming `kicks` to `kick`. Also - do we need the `LIMIT 1`? I imagine it might be good practice to limit to 1 if we can only handle there being 1 result (and we might want to do the same above for `ilk_id`/`address_id`), but I'm also wondering if there's a way we can fail fast if that assumption is broken 🤔

db/migrations/20201214160609_update_get_flip_query.sql
```diff
@@ -0,0 +1,70 @@
...
+         LIMIT 1
+     ),
+     deals AS (SELECT deal.bid_id
+               FROM maker.deal
+                        LEFT JOIN public.headers ON deal.header_id = headers.id
+               WHERE deal.bid_id = get_flip_with_address.bid_id
+                 AND deal.address_id = (SELECT * FROM address_id)
+                 AND headers.block_number <= block_height)
+
+SELECT get_flip_with_address.block_height,
```
rmulhol: I think we've gone back and forth on this in the past, so I have no problem with sticking to convention, but I'm wondering if it might make more sense to return the block height from the `maker.flip` record (in `storage_values`) as opposed to the block height passed by the user to the query - to clarify when we last saw a value assigned. Could go either way here though, not sure what's more clear

db/migrations/20201214160609_update_get_flip_query.sql
```diff
@@ -0,0 +1,70 @@
...
+       storage_values.lot,
+       storage_values.bid,
+       storage_values.gal,
+       CASE (SELECT COUNT(*) FROM deals)
+           WHEN 0 THEN FALSE
+           ELSE TRUE END,
+       storage_values.tab,
+       storage_values.created,
+       storage_values.updated,
+       flip_address AS flip_address
```
rmulhol: nbd but would it make sense to say `get_flip_with_address.flip_address AS flip_address` to line up with the stuff above?

db/migrations/20201214162833_update_all_flips_query.sql
```diff
@@ -0,0 +1,73 @@
...
+-- SQL in this section is executed when the migration is applied.
+
+-- update all_flips function to use get_flip_with_address
+CREATE OR REPLACE FUNCTION api.all_flips(ilk TEXT, max_results INTEGER DEFAULT -1, result_offset INTEGER DEFAULT 0)
+    RETURNS SETOF api.flip_bid_snapshot AS
+-- +goose StatementBegin
+$BODY$
+BEGIN
+    RETURN QUERY (
+        WITH ilk_ids AS (SELECT id
```
rmulhol: maybe `ilk_ids` should be singular?

db/migrations/20201214160609_update_get_flip_query.sql
```diff
@@ -0,0 +1,70 @@
+-- +goose Up
+
+-- add a new function that works the same as get_flip but takes into account a specific flip_address
+-- this is required because there are now more than one flip contract per ilk
+CREATE OR REPLACE FUNCTION api.get_flip_with_address(bid_id NUMERIC, flip_address TEXT, ilk TEXT, block_height BIGINT DEFAULT api.max_block())
```
rmulhol: Trying to think about whether it'd be possible to eliminate the `ilk` from the params here - seems kinda redundant if we're passing an address (for a flip that's associated with exactly one ilk). Nbd if it's complicated

db/migrations/20201217151117_update_flip_bid_event_computed_column.sql
```diff
@@ -0,0 +1,35 @@
+-- +goose Up
+CREATE OR REPLACE FUNCTION api.flip_bid_event_bid(event api.flip_bid_event)
+    RETURNS api.flip_bid_snapshot AS
+$$
+WITH ilks AS (
```
rmulhol: Sorry to keep nitpicking singular vs plural 😅 but maybe this should be `ilk`?

transformers/component_tests/queries/all_flips_query_test.go
```diff
@@ -151,52 +208,50 @@ var _ = Describe("All flips view", func() {
...
 
-		expectedBid2 := test_helpers.FlipBidFromValues(strconv.Itoa(bidIdTwo), strconv.Itoa(int(ilkIdTwo)),
+		expectedBid2 := test_helpers.FlipBidFromValues(strconv.Itoa(fakeBidIdTwo), strconv.Itoa(int(ilkIdTwo)),
 			strconv.Itoa(int(urnIdTwo)), "false", headerTwo.Timestamp, headerTwo.Timestamp, flipStorageValuesTwo)
+		expectedBid2.FlipAddress = anotherFlipAddress
 
 		var actualBid1 test_helpers.FlipBid
-		queryErr1 := db.Get(&actualBid1, `SELECT bid_id, ilk_id, urn_id, guy, tic, "end", lot, bid, gal, dealt, tab, created, updated FROM api.all_flips($1) WHERE bid_id = $2`,
-			ilkOne.Identifier, bidIdOne)
+		queryErr1 := db.Get(&actualBid1, `SELECT bid_id, ilk_id, urn_id, guy, tic, "end", lot, bid, gal, dealt, tab, flip_address, created, updated FROM api.all_flips($1) WHERE bid_id = $2`,
```
rmulhol: Wondering if it might make sense to do a `db.Select` into a `[]test_helpers.FlipBid` (i.e. omit the `where bid_id = $1`), and then verify that the results contain both addresses? Thinking that at least in theory we would expect the same bid ID to exist on multiple contracts for the same ilk, so that would more closely approximate actual usage

transformers/component_tests/queries/all_flips_query_test.go
```diff
@@ -126,22 +130,75 @@ var _ = Describe("All flips view", func() {
...
+		expectedBid2.FlipAddress = anotherContractAddress
+
+		var actualBid1 test_helpers.FlipBid
+		queryErr1 := db.Get(&actualBid1, `SELECT bid_id, ilk_id, urn_id, guy, tic, "end", lot, bid, gal, dealt, tab, flip_address, created, updated FROM api.all_flips($1) WHERE bid_id = $2`,
+			ilkOne.Identifier, fakeBidIdOne)
+		Expect(queryErr1).NotTo(HaveOccurred())
+		Expect(expectedBid1).To(Equal(actualBid1))
+
+		var actualBid2 test_helpers.FlipBid
+		queryErr2 := db.Get(&actualBid2, `SELECT bid_id, ilk_id, urn_id, guy, tic, "end", lot, bid, gal, dealt, tab, flip_address, created, updated FROM api.all_flips($1) WHERE bid_id = $2`,
```
rmulhol: Thinking maybe we can omit the `where bid_id = $1` here?

transformers/component_tests/queries/all_flips_query_test.go
```diff
@@ -151,52 +208,50 @@ var _ = Describe("All flips view", func() {
...
-		test_helpers.CreateFlip(db, headerTwo, flipStorageValuesTwo, test_helpers.GetFlipMetadatas(strconv.Itoa(bidIdTwo)), anotherFlipAddress)
+		flipStorageValuesTwo := test_helpers.GetFlipStorageValues(2, ilkOne.Hex, fakeBidIdTwo)
+		test_helpers.CreateFlip(db, headerTwo, flipStorageValuesTwo, test_helpers.GetFlipMetadatas(strconv.Itoa(fakeBidIdTwo)), anotherFlipAddress)
 
 		// insert a new bid associated with a different flip contract address
 		ilkIdTwo, urnIdTwo, setupErr := test_helpers.SetUpFlipBidContext(test_helpers.FlipBidContextInput{
 			DealCreationInput: test_helpers.DealCreationInput{
 				DB:              db,
-				BidId:           bidIdTwo,
+				BidId:           fakeBidIdTwo,
```
rmulhol: As a matter of fact, I wonder if these two records should have the same bid ID, so that they only differ on the address 🤔

transformers/component_tests/queries/test_helpers/test_helpers.go
```diff
@@ -488,11 +489,12 @@ func FlapBidFromValues(bidId, dealt, updated, created string, bidValues map[stri
...
-		IlkId:     ilkId,
-		UrnId:     urnId,
-		Gal:       bidValues[storage.BidGal].(string),
-		Tab:       bidValues[storage.BidTab].(string),
+		commonBid:   commonBidFromValues(bidId, dealt, updated, created, bidValues),
+		IlkId:       ilkId,
+		UrnId:       urnId,
+		Gal:         bidValues[storage.BidGal].(string),
+		Tab:         bidValues[storage.BidTab].(string),
+		FlipAddress: fakes.FakeAddress.Hex(),
```
rmulhol: Does the fact that we're using this value here mean that we need to use the same one when we're setting up/asserting against flip records in the tests calling this function? Wondering if we might want to add a param to inject it to avoid coupling to this specific string

rmulhol (approved):
LGTM - thanks for doing this work!

I dropped a bunch of nitpicks but I think really the only things I'd zero in on are:
1) Do we want to modify the tests to make sure we're exercising a scenario where we're getting two flips for the same ilk _and the same bid ID_ from different contracts?
2) Are we building in an assumption in the component tests that the setup for flip records will always use `fakes.FakeAddress`? Might we want to allow those tests to use random strings without failing?

db/migrations/20201214160608_update_flip_bid_snapshot.sql
rmulhol: Definitely on board with squashing. Also I suppose we're back in the realm of verifying that migrations timestamps are always only increasing from the most recent in the DB. Hopefully CI will save us there but probably worth keeping in the back of our minds as we're adding new ones

elizabethengelman: That's a good call - I think that I ran into a failed build when the timestamps were out of order, so hopefully we'll be okay on that front.

db/migrations/20201217151117_update_flip_bid_event_computed_column.sql
elizabethengelman: Thanks for catching these, I meant to change some of them to singular and forgot! 🙏

db/migrations/20201214160609_update_get_flip_query.sql
elizabethengelman: Good call - I think that we're okay with not including `LIMIT 1` for the ilk and addresses queries because they both have uniqueness constraints on the fields we're querying by (ilk identifier and address). 

I like the idea of failing fast if there is more than one kick - I was trying to get that to work with an `ASSERT` in there, but haven't quite figured out how to get that to work. I'm wondering if folks think that'd be a useful addition?

rmulhol: I'm fine with abandoning this line of thought if it's introducing too much complexity. I think in theory we should only have one kick per bid ID per address, but don't think business rules should creep into the schema unless we can back it up with constraints.

db/migrations/20201214160609_update_get_flip_query.sql
elizabethengelman: Yeah, I like this idea a lot. 👍

transformers/component_tests/queries/all_flips_query_test.go
elizabethengelman: Good call! I ended up adding a new test for this, which may be overkill, but it seems like it'd be nice to test the scenario where the bids have the same id, and when they have different ids.

transformers/component_tests/queries/test_helpers/test_helpers.go
elizabethengelman: I like the idea of passing in the address here, and think I was avoiding doing that earlier because this method is used in LOTS of places! I had been setting the FlipAddress on the returned FlipBid for tests where it matters, but I think that passing in the address is definitely the cleaner way to handle it.

rmulhol: Cool, no worries from my perspective if we want to make that a separate story, if it would blow up the diff

elizabethengelman: I think I was being dramatic, it wasn't _that_ many places 😜

db/migrations/20201214160609_update_get_flip_query.sql
elizabethengelman: That's a good point about avoiding business rules in the schema without constraints. Are you thinking that including an `ASSERT` without a constraint on the flip_kick table would be an example of business logic seeping into the schema?

I think you're right that we should only ever have one kick per bid ID per address, since the flipper contract increments the bid id when `kick` is called: https://github.com/makerdao/dss/blob/master/src/flip.sol#L124. But we don't have a constraint for that in the db. 🤔

elizabethengelman:
Once this is merged, we can run the following query in the API to ensure that it has taken affect, and has fixed the issue:
```
{
  allFlips(ilk: "ETH-A", first: 50) {
    nodes {
      bidId
      guy
    }
  }
}

```

Currently, this query returns a list of null nodes:
```
  "data": {
    "allFlips": {
      "nodes": [
        null,
        null,
        null,
        ...
        null
      ]
    }
  },
```

ecmendenhall (approved):

PR #406, opened 2020-12-15
Turn on recheck headers

paytonrules:
Seems too easy, I probably flubbed it.


rmulhol (commented):
I _think_ this should do it 👍 

Prob worth doing the same thing for other plugins - maybe even starting there

paytonrules:
That's a good point - want to do Oasis first? https://github.com/makerdao/vdb-oasis-transformers/pull/12

rmulhol:
Would like to do this in a separate container so it doesn't slow down the main sync

ecmendenhall:
We're gonna run this as a parallel container.

PR #408, opened 2020-12-18
Update collateral generator

elizabethengelman:
This updates the collateral generator to account for versioned Median contracts. I think that @paytonrules created a story to fix the Median names, and I think that still warrants some attention in case we want to allow users to update the generated collateral name. But in the mean time, this should make adding new collaterals a bit easier.


generators/new_collateral/types/collateral.go
```diff
@@ -54,9 +59,10 @@ func (c Collateral) GetFlipContractName() string {
 }
 
 func (c Collateral) GetMedianContractName() string {
-	// example: MEDIAN_ETH_B
+	// example: MEDIAN_ETH_B_1_1_3
 	name := strings.ToUpper(strings.Replace(c.Name, "-", "_", -1))
-	return fmt.Sprintf("MEDIAN_%s", name)
+	version := fmt.Sprintf("%s", strings.Replace(c.Version, ".", "_", -1))
```
rmulhol, line 76: A little confused by this addition - might we want to panic or something if we detect versions with a `.` separator? Thinking we probably don't want to start to allow them if they might still blow up other parts of the code

generators/new_collateral/test_data/test_data.go
```diff
@@ -312,7 +312,7 @@ var (
     repository = "github.com/makerdao/vdb-mcd-transformers"
     type = "eth_storage"
   [exporter.log_median_price]
-    contracts = ["MEDIAN_BAT_1_0_0", "MEDIAN_ETH_B"]
+    contracts = ["MEDIAN_BAT_1_0_0", "MEDIAN_ETH_B_1_1_3"]
```
rmulhol: I think having `ETH_B` as an example collateral here is a little confusing, because we don't have distinct median contracts for the same collateral based on the letter version (e.g. "A" vs "B"), right? Wondering if it might make sense to eliminate that part altogether so that these tests reflect the actual config (and/or if we want to validate the assumption that collateral with different letter version will always share the same medianizer).

rmulhol (approved):
These changes lgtm but I'd love for @paytonrules to chime in because I remember him mentioning that we couldn't have versions for ETH_B and I don't remember why that was :(

generators/new_collateral/test_data/test_data.go
paytonrules: > These changes lgtm but I'd love for @paytonrules to chime in because I remember him mentioning that we couldn't have versions for ETH_B and I don't remember why that was :(

It had something to do with there not actually being a MEDIAN_ETH_B..I think. It caused test failures though so if it's working here it may be fine. It's possible I fixed the cause but this is probably worth revisiting when we get back. I might be able to track down the commit and refresh both our memories.

elizabethengelman: Oh good point - I think I was conflating adding numbered version and lettered versioning. I think we started adding the numbered versioning because Eric discovered that the median contract had changed. But I think we can probably safely  remove the lettered version for now. I'll make that update.

generators/new_collateral/types/collateral.go
elizabethengelman: This change should be adding a version number delimited with "\_" to the end of the median contract name, since we have more than one median contract versions now. This piece should be changing the version that the user enters into the script from "." delimited to "\_" delimited so that we don't run into issues with other pieces of the code. If the user enters the version in with "\_", then i think this line shouldn't make any changes.

generators/new_collateral/types/collateral.go
```diff
@@ -17,6 +17,24 @@ func NewCollateral(name, version string) Collateral {
...
 
+func (c Collateral) formattedLowerCaseName() string {
+	return strings.ToLower(strings.Replace(c.Name, "-", "_", -1))
+}
+
+func (c Collateral) formattedUpperCaseName() string {
+	return strings.ToUpper(strings.Replace(c.Name, "-", "_", -1))
+}
+
+func (c Collateral) FormattedLowerCaseNameWithoutVersionLetter() string {
```
elizabethengelman, line 28: These two methods probably aren't super future proof since they are expecting that all collateral names be "NAME" or "NAME_LETTER" and just chop the letter piece off, and this probably isn't guaranteed. But I think using this method will it easier to add new collaterals to the system in the short term, and then maybe we can re-think this change as part of https://makerdao.atlassian.net/browse/VDB-1704.

rmulhol: Makes sense to me. Maybe worth updating that story? Thinking these changes prevent us from running into cases where it generates `median_eth_b`, but the new risk would be chopping a multi-word collateral name - is that right? Like, `collateral_number_one_a` would end up parsing to `median_collateral` instead of `median_collateral_number_one` - if I follow.

generators/new_collateral/types/collateral.go
```diff
@@ -17,50 +17,67 @@ func NewCollateral(name, version string) Collateral {
...
+	// example: ETH_B => ETH
```
rmulhol: Should this example be `ETH_B => eth`?

generators/new_collateral/types/collateral.go
elizabethengelman: Yep, you got it - this wouldn't work if the collateral has a multi-word name. Updated VDB-1704 to mention this ✔️

generators/new_collateral/types/collateral.go
elizabethengelman: Yep, good catch! Addressed in 328817f

PR #409, opened 2020-12-18
Vdb 1752 add 1.2.2 collaterals to prod

elizabethengelman:


transformers/shared/constants/method.go
```diff
@@ -102,13 +104,15 @@ func MedianV100ABI() string {
 
 func MedianV114ABI() string {
```
paytonrules, line 105: The function can probably be renamed since there's some V1_2_2 in here as well.

paytonrules (commented):
🎉

transformers/shared/constants/method.go
elizabethengelman: That's a good call. I'm not sure what the best naming convention should be for these sorts of functions that have multiple versions in them. I think previously I was thinking about it as this function would return all Median contract from version 1.1.4 onward, but that may be confusing. Any ideas?

paytonrules: I don't have a great one. For other contracts, like flip, there's a distinct V100 and V11....

This might be the best for now, in context.

paytonrules (approved):

PR #410, opened 2021-01-05
(VDB-1713) ilk snapshot dunk

rmulhol:


elizabethengelman (approved):
Looks good!

If I'm remembering correctly, this change will fix all new ilk snapshots, but we'll still need to make a change to update the old ones with missing dunk values - is that right?

rmulhol:
> Looks good!
> 
> If I'm remembering correctly, this change will fix all new ilk snapshots, but we'll still need to make a change to update the old ones with missing dunk values - is that right?

I believe that is correct, I don't _think_ the triggers will update intermediate values automatically. I think that we might be able to fix those values by deleting the old `vat_ilk_dunk` values and switching their diff's status back to `new` - if those diffs are reprocessed and the `insert` triggers fire, then that should update all subsequent snapshots (until the next `dunk`)

ecmendenhall (approved):
Looks good. For my understanding, this was just an accidentally missed column in the SQL?

rmulhol:
> Looks good. For my understanding, this was just an accidentally missed column in the SQL?

Basically yes - this column was added to ilks as a replacement for `lump`, and we updated the ilk snapshot to include the new column but accidentally did not update the triggers for every other field on ilks to carry forward the `dunk` value. This PR fixes that.

PR #411, opened 2021-01-05
(VDB-1723) unrecognized diffs window

rmulhol:


paytonrules (approved):

elizabethengelman (approved):
This LGTM! I think we talked about this a couple weeks ago - but we think that the `UNRECOGNIZED_DIFF_STARTING_BLOCK` env var isn't currently being set, so it's defaulting to -1, right?

rmulhol:
> This LGTM! I think we talked about this a couple weeks ago - but we think that the `UNRECOGNIZED_DIFF_STARTING_BLOCK` env var isn't currently being set, so it's defaulting to -1, right?

I believe so, yes

PR #412, opened 2021-01-05
(VDB-1723) Increase unrecognized diffs window

rmulhol:
- Capture diffs that are unrecognized but further back


ecmendenhall (approved):
I'm in!

PR #414, opened 2021-01-08
Add index on api.urn_snapshot block_height

ecmendenhall:
- Add `urn_snapshot_block_height_index` on `api.urn_snapshot` `block_height` column. (The completeness checker relies on a query filter that uses this index).
- Create index concurrently. This command can't be run in a transaction block.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 11.2
--- Dumped by pg_dump version 11.10
+-- Dumped from database version 11.6
+-- Dumped by pg_dump version 11.6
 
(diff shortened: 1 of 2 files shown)


db/schema.sql
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 11.6
--- Dumped by pg_dump version 11.6
+-- Dumped from database version 11.2
+-- Dumped by pg_dump version 11.10
```
rmulhol: Would be cool to leave these versions intact because I think we're running 11.6 on our deploy

ecmendenhall: Updated.

rmulhol (approved):

PR #415, opened 2021-01-11
Add min_block_height arg to api.get_urns_by_ilk

ecmendenhall:
- Add an optional `min_block_height` argument to `api.get_urns_by_ilk`
- Update tests to exercise new argument. Feedback welcome here—I hope the behavior is clear.
- Postgres considers functions unique based on name and argument signature. That means we can use `CREATE OR REPLACE` to modify functions with the same signature, but here we need to `DROP` and `CREATE` a function with the same name, but an extra argument. The potential impact here is that this query will be briefly unavailable when the migration is running.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -1492,7 +1492,7 @@ FROM (SELECT DISTINCT ON (urn_identifier, urn_snapshot.ilk_identifier) urn_ident
                                                                        updated
       FROM api.urn_snapshot
       WHERE urn_snapshot.block_height <= get_urns_by_ilk.block_height
-        AND urn_snapshot.block_height > get_urns_by_ilk.min_block_height
+        AND urn_snapshot.block_height >= get_urns_by_ilk.min_block_height
         AND urn_snapshot.ilk_identifier = get_urns_by_ilk.ilk_identifier
       ORDER BY urn_identifier, ilk_identifier, updated DESC) AS latest_urns
 ORDER BY updated DESC
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -1476,10 +1476,10 @@ $$;
 
 
 --
--- Name: get_urns_by_ilk(text, bigint, integer, integer); Type: FUNCTION; Schema: api; Owner: -
+-- Name: get_urns_by_ilk(text, bigint, bigint, integer, integer); Type: FUNCTION; Schema: api; Owner: -
 --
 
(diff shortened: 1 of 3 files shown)


db/schema.sql
```diff
@@ -1492,6 +1492,7 @@ FROM (SELECT DISTINCT ON (urn_identifier, urn_snapshot.ilk_identifier) urn_ident
                                                                        updated
       FROM api.urn_snapshot
       WHERE urn_snapshot.block_height <= get_urns_by_ilk.block_height
+        AND urn_snapshot.block_height > get_urns_by_ilk.min_block_height
```
ecmendenhall: Probably want `>=` here.

transformers/component_tests/queries/get_urns_by_ilk_query_test.go
```diff
@@ -248,20 +262,40 @@ var _ = Describe("All Urns function", func() {
...
 				BlockHeight:   blockTwo,
-				Ink:           strconv.Itoa(updatedInk),
+				Ink:           strconv.Itoa(blockTwoInk),
 				Art:           strconv.Itoa(setupDataOne[vat.UrnArt].(int)), // Not changed
 				Created:       helper.GetValidNullString(createdTimestamp),
 				Updated:       helper.GetValidNullString(updatedTimestampTwo),
 			}
 
-			err = db.Get(&actualUrn, urnsByIlkQuery, helper.FakeIlk.Identifier, blockTwo)
+			urnsByIlkMinHeightQuery := `SELECT urn_identifier, ilk_identifier, block_height, ink, art, created, updated FROM api.get_urns_by_ilk($1, $2, $3)`
```
rmulhol, line 297: I'm thinking it'd be cool to write this test such that we were passing a min but no max, since that's the use case behind this PR - right? But I also get that it may be difficult to do that because of argument order, which has me wondering - how does postgraphile handle a case where we only pass a min (but no max)? Maybe worth firing up postgraphile against a local DB with these changes to verify we can actually execute that call, if you haven't already

ecmendenhall: Good call. I'll explicitly verify this, but since there are defaults for all args, I believe Postgraphile inherits the default (`api.max_block()`) when you only pass a min.

rmulhol: If so, then this LGTM 👍

rmulhol (approved):

PR #416, opened 2021-01-12
Creates query to get block heights of transformed diffs

yaoandrew:
This creates a new query to get the block height for transformed diffs in order to be used in [VDB status page](https://statuspage.makerdao.com). It replaces:

`allStorageDiffs(condition: { checked: true }, last: 1) {
			nodes {
				blockHeight
			}`


db/migrations/20210111183822_get_block_heights_for_transformed_diffs.sql
```diff
@@ -0,0 +1,14 @@
+-- +goose Up
+
+CREATE OR REPLACE FUNCTION api.get_block_heights_for_transformed_diffs()
+    RETURNS SETOF BIGINT
+    LANGUAGE sql
+    STABLE
+AS
+$$
+SELECT block_height FROM public.storage_diff WHERE status = 'transformed' ORDER BY block_height ASC
```
ecmendenhall: Is it useful to get back a whole set here, or should we just limit to the latest? I don't feel super strongly either way—if the past few block heights are useful in some way, keep them.

ecmendenhall (dismissed):
Looks good! One general comment: I want to raise again the hunch that knowing the latest block height for the other statuses (new, pending, non-canonical, etc.) might be useful, too. If we can get that data easily through other queries, maybe this isn't the right place for it, but it strikes me as potentially helpful in a debugging scenario.

db/migrations/20210111183822_get_block_heights_for_transformed_diffs.sql
rmulhol: Yeah I think we probably only need the maximum block height of any transformed diff. I believe that this query would return millions of rows

yaoandrew: Ah makes sense. I think I misunderstood how the query would be used with `last` argument to filter down the collection.

rmulhol: Yeah, that's a common gotcha with postgraphile - the `first`/`last` params aren't very helpful if being used on a function call: https://www.graphile.org/postgraphile/custom-queries/#advice

yaoandrew:
> Looks good! One general comment: I want to raise again the hunch that knowing the latest block height for the other statuses (new, pending, non-canonical, etc.) might be useful, too. If we can get that data easily through other queries, maybe this isn't the right place for it, but it strikes me as potentially helpful in a debugging scenario.

Sounds good. We've added this [story](https://makerdao.atlassian.net/browse/VDB-1770?atlOrigin=eyJpIjoiMWQ0NTliY2E2MDk5NDMwMzgzMThkZWQ2NzMwYTg1ODUiLCJwIjoiaiJ9) to the board.

transformers/component_tests/queries/get_block_heights_for_transformed_diffs_test.go
```diff
@@ -0,0 +1,56 @@
...
+	const blockHeightsForTransformedStorageDiffs = `SELECT * FROM api.get_block_heights_for_transformed_diffs()`
+
+	BeforeEach(func() {
+		test_config.CleanTestDB(db)
+	})
+
+	It("ignores untransformed diffs", func() {
+		test_helpers.CreateFakeDiffRecord(db)
+
+		var diff int
```
yaoandrew: Maybe this should be a `uint64`?

db/migrations/20210111183822_get_block_heights_for_transformed_diffs.sql
```diff
@@ -0,0 +1,14 @@
+-- +goose Up
+
+CREATE OR REPLACE FUNCTION api.get_block_heights_for_transformed_diffs()
+    RETURNS SETOF BIGINT
+    LANGUAGE sql
+    STABLE
+AS
+$$
+SELECT MAX (block_height) FROM public.storage_diff WHERE status = 'transformed'
```
rmulhol: Sorry to be super annoying but I think one cool thing to do here could be to actually execute this query against the db with `EXPLAIN ANALYZE` and then compare the results to `SELECT block_height FROM public.storage_diff WHERE status = 'transformed' ORDER BY block_height DESC LIMIT 1`.

Also, wondering if we want a separate PR/migration in vdb to add a partial index for diffs where status = transformed - could be worth doing that first to see if the index affects the explain analyze output.

transformers/component_tests/queries/get_block_heights_for_transformed_diffs_test.go
rmulhol: Could you say more here? I'd be fine either way but am unsure of the costs/benefits

transformers/component_tests/queries/get_block_heights_for_transformed_diffs_test.go
```diff
@@ -0,0 +1,56 @@
...
+		Expect(diff).To(Equal(0))
```
rmulhol: Does this test need to invoke the query before making this assertion?

rmulhol (commented):
Super close to an approve - my comments about profiling + indexes are not merge blocking, but it seems like the `"ignores untransformed diffs"` test is missing the step that calls the function?

db/migrations/20210111183822_get_block_heights_for_transformed_diffs.sql
```diff
@@ -0,0 +1,14 @@
+-- +goose Up
+
+CREATE OR REPLACE FUNCTION api.get_block_heights_for_transformed_diffs()
```
rmulhol: Pretty minor but maybe we should change the name of this function and the test file if we're only querying for one record

db/migrations/20210111183822_get_block_heights_for_transformed_diffs.sql
```diff
@@ -0,0 +1,14 @@
...
+    RETURNS SETOF BIGINT
```
rmulhol: also probably don't need SETOF here

transformers/component_tests/queries/get_block_heights_for_transformed_diffs_test.go
yaoandrew: `int` seems to be 32 or 64 bits depending on platform. Since blocks will never be negative and potentially large numbers does it make sense to explicitly have them be uint and 64 bits in size?

yaoandrew: My brain is mush. This is the test. I was thinking about real world block heights that could need larger size.

db/migrations/20210111183822_get_block_heights_for_transformed_diffs.sql
yaoandrew: using MAX
```
Planning Time: 0.165 ms
Execution Time: 0.084 ms
```

using ORDER BY
```
Planning Time: 0.146 ms
Execution Time: 0.094 ms
```

Currently there are 3741377 `transformed` diffs in prod

rmulhol: 👍 🎉 

sounds like we may not need an index 😅

db/migrations/20210111183822_get_max_block_height_of_transformed_diffs.sql
```diff
@@ -0,0 +1,14 @@
+-- +goose Up
+
+CREATE OR REPLACE FUNCTION api.get_max_block_height_of_transformed_diffs()
```
rmulhol: Now I'm just being picky but I wonder if I could convince you to rename this to `get_max_transformed_diff_block`, and then I can update #420 to be `get_max_transformed_event_block` and they'll look similar (reminder to update the migration and test names if you make the change)

rmulhol (approved):

PR #419, opened 2021-01-12
VDB-1756: Update urn snapshot trigger functions

elizabethengelman:
This PR is attempting to fix an issue we noticed in deleting/updating urn snapshot records. Whenever a header is removed (due to a reorg), the associated transformed diff record is also deleted through a cascade delete, which in turn triggers the urn snapshot at that block number to be removed. 

However, as part of the investigation into [VDB-1756](https://makerdao.atlassian.net/browse/VDB-1756), we found that urn snapshots weren't being removed. This is because the snapshot was being queried for by block number, and the block number was obtained by looking at the vat_urn_ink/vat_urn_art's header_id. But, since this sequence of events is started off when a header is deleted, that meant that the header record no longer existed in the database by the time the urn snapshot triggers were run, so there was no header to query for a block number. So we were getting a null block number value, and then not finding any urn snapshots with null block numbers to delete (or alter). A bulk of this PR is changing the query for getting the urn snapshots to delete (or alter) by looking  at their vat_urn_ink/vat_urn_art record's underlying diff record, and getting the block number that way instead of through the header record.

Another thing that we still probably need to consider is how to identify and update/fix old urn snapshots?


db/schema.sql
```diff
@@ -2211,13 +2211,6 @@ END
 $$;
 
 
---
--- Name: FUNCTION delete_obsolete_urn_snapshot(urn_id integer, header_id integer); Type: COMMENT; Schema: maker; Owner: -
---
-
-COMMENT ON FUNCTION maker.delete_obsolete_urn_snapshot(urn_id integer, header_id integer) IS '@omit';
```
elizabethengelman, line 2218: TODO: make sure that this comment is not removed.

transformers/component_tests/queries/all_urns_query_test.go
```diff
@@ -268,7 +268,8 @@ var _ = Describe("All Urns function", func() {
 			fakeHeaderTwoID, err := headerRepo.CreateOrUpdateHeader(fakeHeaderTwo)
 			Expect(err).NotTo(HaveOccurred())
 
-			err = vatRepo.Create(diffID, fakeHeaderTwoID, metadata.UrnInk, strconv.Itoa(updatedInk))
+			diffIdTwo := test_helpers.CreateFakeDiffRecordWithHeader(db, fakeHeaderTwo)
```
paytonrules, line 271: Is this just making sure there was an underlying diff, and when that happened would an error be returned?

db/migrations/20210114161425_update_urn_snapshot_trigger.sql
```diff
@@ -0,0 +1,506 @@
...
+WHERE vat_urn_art.urn_id = urn_art_before_block.urn_id
+  AND headers.block_number < (SELECT block_number FROM passed_block_number)
+ORDER BY block_number DESC
+LIMIT 1
+$$
+    LANGUAGE sql;
+
+
+
+-- drop new delete_obsolete_urn_snapshot and replace with old function
```
paytonrules, line 291: These comments are very helpful in this long migration.

transformers/component_tests/queries/all_urns_query_test.go
paytonrules, line 271: Why did you have to add these?

transformers/storage/vat/repository_test.go
```diff
@@ -1051,10 +1074,10 @@ var _ = Describe("Vat storage repository", func() {
 
 				It("deletes urn state associated with diff if it's the earliest state in the table", func() {
 					initialInk := rand.Int()
-					setupErrOne := repo.Create(diffID, headerOne.Id, urnInkMetadata, strconv.Itoa(initialInk))
+					setupErrOne := repo.Create(diffIdOne, headerOne.Id, urnInkMetadata, strconv.Itoa(initialInk))
```
paytonrules, line 1148: Here, with the creation of diffIdOne you've ensured the urnSnapshot is deleted right? Because before using the generic diffID hid that error?

paytonrules (approved):
I'm pretty sure this is good. I'm not 100% sure the case (urnSnapshot not being deleted) is actually tested but I think it is.

transformers/component_tests/queries/all_urns_query_test.go
elizabethengelman: I added these storage_diff records for each urn that was created since `delete_obsolete_urn_snapshot` (and other functions associated with this trigger) are now getting block numbers from the diff record, rather than the header. And so, we needed to have a new diff so that the functions could find the block number that we're expecting, so the correct snapshot would be deleted.

transformers/storage/vat/repository_test.go
elizabethengelman: Yep, that's right. The key piece here is that the diff's `block_height` is the same as the `block_number` on the header that is deleted. So, we can make sure that the `urn_snapshot_block_number` that is being queried for in the `delete_obsolete_urn_snapshot` function is the correct block number's snapshot to delete.

elizabethengelman:
I think you're right @paytonrules - I think that the tests implicitly are testing that the urn snapshot in question was deleted, because we're testing the number of snapshots before and after a header is removed. But, I just wrote a test that more closely mimics the issue we saw from VDB-1756, I'll plan to clean that up and push it up shortly.

transformers/storage/vat/repository_test.go
```diff
@@ -781,6 +781,58 @@ var _ = Describe("Vat storage repository", func() {
 					Expect(queryErr).NotTo(HaveOccurred())
 					Expect(len(urnStates)).To(Equal(0))
 				})
+
+				It("handles removing snapshots when there's a three-part reorg", func() {
+					// this test case mimics the issue that is detailed in VDB-1756 where we saw a reorg that resulted in
+					// receiving three headers - canonical, noncanonical, canonical
+
+					// set up the header, diff and snapshot before the scenario being tested
```
paytonrules: Why do you need all the "before" stuff?

paytonrules (approved):

transformers/component_tests/queries/all_urns_query_test.go
paytonrules: 👍

db/migrations/20210114161425_update_urn_snapshot_trigger.sql
```diff
@@ -0,0 +1,506 @@
...
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION maker.update_urn_arts_until_next_diff(start_at_diff maker.vat_urn_art, new_art NUMERIC) RETURNS maker.vat_urn_art
+AS
+$$
+DECLARE
+    diff_block_number    BIGINT := (
+        SELECT block_height
+        FROM public.storage_diff
+        WHERE id = start_at_diff.diff_id);
+    next_rate_diff_block BIGINT := (
```
rmulhol: nbd but this could probably be `next_art_diff_block`, I think?

transformers/storage/vat/repository_test.go
```diff
@@ -756,17 +770,69 @@ var _ = Describe("Vat storage repository", func() {
...
+					art := artBefore + 1
+					block := blockBefore + 1
+					blockTimestamp := blockBeforeTimestamp + 1
+
+					// HEADER WITH HASH ONE
+					headerWithHashOne := CreateHeaderWithHash(hashOne.String(), blockTimestamp, block, db)
+					diffId1 := CreateFakeDiffRecordWithHeader(db, headerWithHashOne)
+					setupErrOne := repo.Create(diffId1, headerWithHashOne.Id, urnArtMetadata, strconv.Itoa(art))
+					Expect(setupErrOne).NotTo(HaveOccurred())
+
```
rmulhol, line 813: Is it accurate that we'd expect the number of urn states to be 2 here as well, before the next header/diff for the same block is inserted - implying that the next insert basically replaces the one that was created here?

transformers/storage/vat/repository_test.go
```diff
@@ -756,17 +770,69 @@ var _ = Describe("Vat storage repository", func() {
...
 
 					var urnStates []test_helpers.UrnState
 					queryErr := db.Select(&urnStates, getArtQuery)
 					Expect(queryErr).NotTo(HaveOccurred())
 					Expect(len(urnStates)).To(Equal(0))
 				})
+
+				It("handles removing snapshots when there's a three-part reorg", func() {
+					// this test case mimics the issue that is detailed in VDB-1756 where we saw a reorg that resulted in
+					// receiving three headers - canonical, noncanonical, canonical
```
rmulhol: Seems like this test is saying we get:
- canonical for first block
- canonical for second block
- non-canonical for second block

Is that right? The comment makes me think the test would deal with everything at the same header, but then I don't understand incrementing the block number on line 803 🤔

transformers/storage/vat/repository_test.go
```diff
@@ -1086,6 +1161,58 @@ var _ = Describe("Vat storage repository", func() {
...
+					// third header event for the same block, which turns out to be canonical. But we are not reprocessing
+					// the associated diffs at that block and are therefore not reinserting the associated
+					// vat_urn_ink/vat_urn_art records
+					CreateHeaderWithHash(hashOne.String(), blockTimestamp, block, db)
+					var urnStates2 []test_helpers.UrnState
+					queryErr2 := db.Select(&urnStates2, getInkQuery)
+					Expect(queryErr2).NotTo(HaveOccurred())
+					// TODO: once we fix the issue where we're not reprocessing the canonical diff, this assertion should
+					// be change to: Expect(len(urnStates2)).To(Equal(2))
+					Expect(len(urnStates2)).To(Equal(1))
```
rmulhol: Maybe it makes sense to include an assert that the `ink` state matches `inkBefore`?

db/migrations/20210114161425_update_urn_snapshot_trigger.sql
elizabethengelman: Nice catch!

transformers/storage/vat/repository_test.go
elizabethengelman: That's a good question - I probably don't need the before stuff, but I was adding it in there to more closely model what happened in prod. I wanted to prove that this trigger wouldn't manipulate data that was already in the database before the scenario in question, if that makes sense. Probably would be more clear if I was asserting against the art/ink values in the "before" snapshot to make sure that it remains unchanged. 

I'm also happy to remove all that before stuff, if it adds more confusion than is worth is.

transformers/storage/vat/repository_test.go
elizabethengelman: Yeah, this assumption is correct. I added a "before" block, so that we'd make sure to have some data in the db before testing this scenario out to make sure that we weren't unintentionally manipulating snapshots that were already in the db, and shouldn't have been changed. Going to add some additional assertions and comments to hopefully clear up that confusion some.

rmulhol: I'm totally on board with including the before block to verify we're only manipulating the snapshot in question 👍

transformers/storage/vat/repository_test.go
elizabethengelman: Yep, that's right. I'll add an assertion here too, just to be explicit.

PR #421, opened 2021-01-15
Assign env in deploy before checking for env

rmulhol:
- deploy fails on no env if we don't grab it from the args before
  checking that it's assigned

### DIFF
--- a/.travis/deploy.sh
+++ b/.travis/deploy.sh
@@ -10,6 +10,7 @@ function message() {
     echo
 }
 
+ENVIRONMENT=$1
 if [ -z "$ENVIRONMENT" ]; then
     echo 'You must specify an environment (bash deploy.sh <ENVIRONMENT>).'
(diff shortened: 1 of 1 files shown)


.travis/deploy.sh
```diff
@@ -10,6 +10,7 @@ function message() {
     echo
 }
 
+ENVIRONMENT=$1
```
paytonrules, line 13: Sure does need to be moved up. How did this happen?

paytonrules (approved):

.travis/deploy.sh
jeannettemcd: > Sure does need to be moved up. How did this happen?

You are correct, it does need to move up. It gets passed in through the travis.yml file when the script is called.

elizabethengelman (approved):

ecmendenhall (approved):
Good catch! Thanks for the quick fix.

PR #424, opened 2021-01-20
Vdb 1756 update trigger approach

elizabethengelman:
This PR adds a function to the `update_urn_ink` and `update_urn_art` triggers so that on delete the underlying `storage_diff` record is changed from `transformed` to `pending`. I initially was thinking of doing this change only for diffs within 15 of the head of the chain, but I think that this is safe to do for older diffs as well, because this will ensure that they get one more check before they are marked as `noncanonical` - happy to rethink this though.

Also, hoping to get some feedback on this approach before potentially making the change to ilk snapshot triggers as well.


db/migrations/20210120091230_update_urn_snapshot_trigger_to_mark_diff_as_pending.sql
```diff
@@ -0,0 +1,97 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION maker.mark_noncanonical_transformed_diff_as_pending(diff_id BIGINT) RETURNS VOID
+AS
+    $$
```
paytonrules: Formatting is inconsistent here. You don't indent the `$$` section on the function below.

db/migrations/20210120091230_update_urn_snapshot_trigger_to_mark_diff_as_pending.sql
```diff
@@ -0,0 +1,97 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION maker.mark_noncanonical_transformed_diff_as_pending(diff_id BIGINT) RETURNS VOID
```
paytonrules: Might be better to just say `mark_transformed_diff_as_pending`. I know the use case is for diffs that are actually noncanonical, but the function itself doesn't make that distinction.

go.mod
```diff
@@ -10,12 +10,12 @@ require (
...
 	github.com/lib/pq v1.0.0
 	github.com/makerdao/vdb-transformer-utilities v0.0.0-20201021185605-2af87288ff1b
-	github.com/makerdao/vulcanizedb v0.1.0
+	github.com/makerdao/vulcanizedb v0.1.1-0.20210120155730-fed8d6aead03
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/onsi/ginkgo v1.14.1
 	github.com/onsi/gomega v1.10.2
 	github.com/sirupsen/logrus v1.7.0
-	github.com/spf13/cobra v0.0.3
+	github.com/spf13/cobra v0.0.5
```
paytonrules, line 18: Were these updates intentional? go.mod likes to update itself for me, so I'm paranoid.

test_data/vulcanize_schema.sql
```diff
@@ -36,6 +36,7 @@ COMMENT ON SCHEMA public IS 'standard public schema';
 
 CREATE TYPE public.diff_status AS ENUM (
     'new',
+    'pending',
```
paytonrules, line 39: Thanks for remembering to update this!

paytonrules (approved):
Minor comments - I really like the test changes.

go.mod
elizabethengelman: I intentionally updated the vulcanizedb version so that we'd get the new pending status from https://github.com/makerdao/vulcanizedb/blob/prod/libraries/shared/storage/diff_repository.go#L43, and then I think that the cobra update followed from that.

Updating vdb probably wasn't totally necessary because the only place this PR relies on the new pending status in code (vs in the db) is in tests, so I'd be happy to not check this in/

PR #425, opened 2021-01-21
Remove storage backfill

rmulhol:
- include changes from running `make test`

Should wait to merge until the backfill is complete


rmulhol:
Normally it's a problem if we convert from draft to ready for review and then immediately merge, but I don't think that should be the case here as long as this PR is merged before any other PRs that do build/deploy and should not kick off a new backfill.

Also, it make sense to merge ahead of and rebase this PR if we want to deploy changes before the backfill complete, in which case we should update the starting block number of the storage backfill to account for already scanned headers.

db/schema.sql
```diff
@@ -2233,13 +2233,6 @@ END
 $$;
 
 
---
```
yaoandrew, line 2236: Not sure what we decided on this but I just removed it from my PR.

yaoandrew (approved):

elizabethengelman (approved):

PR #426, opened 2021-01-21
Announce a release with each deployment

paytonrules:
This announces one release for deploying all five executables, which fits.

I also refactored the deploy script a bit to highlight the bits that change.


rmulhol:
Maybe worth tagging @jeannettemcd for review?

jeannettemcd (commented):
The deploy.sh looks good.
The announce.sh is using a sentry-cli, that needs to be installed, this deploy uses ubuntu trusty as it's base and does have sentry-cli installed by default, see travis.yml. 
Also, to run sentry-cli it will require a sentry auth token, and github integration.

paytonrules:
@jeannettemcd the announce script installs the sentry-cli as part of it (see lines 12-14).
You're right about the AUTH token, we've put it in travis's secrets on vulcanizedb and will have to do that on this repo as well. We're already running this on vulcanizedb and it works in a limited fashion without the GIthub integration. I wasn't planning on asking for that to be turned on right now, you've got enough on your plate.

jeannettemcd (approved):
LGTM

PR #427, opened 2021-01-22
Clean up panics and return errors instead

yaoandrew:
Reduce or remove use of `panic` and return an `error` instead to allow Sentry to capture errors since Sentry may not capture panics[ occurring in goroutines](https://github.com/getsentry/sentry-go/issues/115#issuecomment-564032753). Most changes occurred in storage repository when the metadata is not recognized.

### DIFF
--- a/transformers/storage/cat/repository.go
+++ b/transformers/storage/cat/repository.go
@@ -64,7 +64,7 @@ func (repo *StorageRepository) Create(diffID, headerID int64, metadata types.Val
 	case IlkDunk:
 		return repo.insertIlkDunk(diffID, headerID, metadata, value.(string))
 	default:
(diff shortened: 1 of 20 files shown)


db/schema.sql
```diff
@@ -2233,13 +2233,6 @@ END
 $$;
 
 
---
```
paytonrules: Needs a rebase maybe? I don't think you meant to check in a schema change.

paytonrules (approved):

db/schema.sql
rmulhol: I think this might be fine - got the same thing in #425 - looks like maybe we forgot to run `make test` before our last merge. Though reminder that we should either merge #425 first or update this PR to also restart the ongoing storage backfill at a more recent block (rather than causing it to restart at the beginning)

PR #428, opened 2021-01-23
(VDB-1802) Handle multiple flop contracts

rmulhol:
- All Flops now returns each instance of a flop bid ID when it exists
  on multiple flop contracts

Worth noting: there were some interesting failures along the way from the queries tests initializing the contract address as a random string of length 42. It should probably "0x" + random string of length 40 if anything - otherwise you get test failures because the first two characters of the address don't match after it goes through `GetOrCreateAddress`. Ended up following Elizabeth's lead and using `fakes.FakeAddress.Hex()`. Happy to make the change elsewhere in the queries tests but didn't want to expand the diff beyond what's already here between core changes + modifications from running `make test`


db/migrations/20210122160549_update_get_flop_query.sql
```diff
@@ -0,0 +1,187 @@
...
+                lot,
+                bid,
+                created,
+                updated
+         FROM maker.flop
+         WHERE flop.bid_id = get_flop_with_address.bid_id
+           AND flop.address_id = (SELECT * FROM address_id)
+           AND block_number <= block_height
+         ORDER BY block_number DESC
+         LIMIT 1
```
paytonrules, line 27: Isn't it still dangerous to use LIMIT 1 here?

rmulhol: I don't think so, because any records that share the same bid ID and address should be at different blocks - so this should just give us the most recent snapshot of a given flop auction. Let me know if there's a scenario I'm missing

paytonrules: The order is probably fixing it, I think you're right.

ecmendenhall (approved):

PR #429, opened 2021-01-25
Restart storage backfill

elizabethengelman:
based on #425

### DIFF
--- a/dockerfiles/backfill_storage/startup_script.sh
+++ b/dockerfiles/backfill_storage/startup_script.sh
@@ -29,5 +29,5 @@ fi
 
 # TODO: update starting/ending block number to match extract diffs deploy block
 # Run getStorageValue
-echo "Running backfillStorage from block 0 to 0"
-./vulcanizedb backfillStorage -s=0 -e=0 --config config.toml
+echo "Running backfillStorage from block 11326755 to 11679900"
+./vulcanizedb backfillStorage -s=11326755 -e=11679900 --config config.toml
--- a/environments/backfillStorage.toml
+++ b/environments/backfillStorage.toml
@@ -3,7 +3,93 @@
     name     = "transformerExporter"
     save     = true
     schema   = "maker"
-    transformerNames = []
+    transformerNames = [
+        "flip_aave_a_v1_2_2",
+        "flip_bal_a_v1_1_14",
+        "flip_eth_b_v1_1_3",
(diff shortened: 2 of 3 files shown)


db/schema.sql
```diff
@@ -2233,13 +2233,6 @@ END
 $$;
 
 
---
--- Name: FUNCTION delete_obsolete_urn_snapshot(urn_id integer, header_id integer); Type: COMMENT; Schema: maker; Owner: -
---
-
-COMMENT ON FUNCTION maker.delete_obsolete_urn_snapshot(urn_id integer, header_id integer) IS '@omit';
```
ecmendenhall: What's up with this schema change? Could you give me the quick summary of how we use Postgraphile magic comments?

ecmendenhall (approved):

db/schema.sql
elizabethengelman: I think we're primarily using comments to omit certain functions and tables from the postgraphile endpoint. I think that when I changed the function signature of `delete_obsolete_urn_snapshot` this comment should have been removed and `COMMENT ON FUNCTION  maker.delete_obsolete_urn_snapshot(urn_id integer, header_id integer, diff_id bigint) IS @omit` should have been added. 

The thing I'm a bit confused about is why/how ` maker.delete_obsolete_urn_snapshot(urn_id integer, header_id integer, diff_id bigint)` isn't showing up in the api, since there isn't a comment on that function in the schema at the moment. @rmulhol do you have any hunches?

rmulhol: Best guess is that something we have in place to disable mutations is preventing this from bring exposed in the API but not sure 🤔

rmulhol: Also I think this schema change just came from running `make test` - possible that a migration change wasn't reflected in the schema at the time the migration PR was merged. I think it should go away if you rebase at this point

elizabethengelman: Yeah, I think you're right that this schema change is probably already in the code base, and this diff will go away with a rebase. 👍

paytonrules: It looks like this is outdated but I think I can fill in - delete_obsolete_urn_snapshot is presumably VOLATILE, and because we disable the mutation plugin in postgraphile it's not going to show up. The '@omit' shouldn't be necessary.

I'd generally prefer we use the JSON in postgraphile for omitting functions instead of the SQL comments, which I think is the same view as the rest of the team.

PR #430, opened 2021-01-25
(VDB-1807) Add new collaterals to execute

rmulhol:


transformers/integration_tests/deny_test.go
```diff
@@ -116,6 +116,16 @@ var _ = Describe("Deny transformer", func() {
 		denyIntegrationTest(int64(11420948), test_data.FlipUniAV121Address(), usrAddress, usrAddress)
 	})
 
+	Context("Flip UNIV2USDCETH_A deny events", func() {
+		usrAddress := "0xda0fab060e6cc7b1C0AA105d29Bd50D71f036711"
+		denyIntegrationTest(11700547, test_data.FlipUniV2UsdcEthAddress(), usrAddress, usrAddress)
```
ecmendenhall, line 121: It's hard to tell from test context what these numbers are—I'm guessing this is the block at which these events begin?

ecmendenhall: (and does it matter that all the rest are `int64`?)

ecmendenhall (approved):

transformers/integration_tests/deny_test.go
rmulhol: That's correct - the block at which the events happened. It's indicated by the argument name for the `denyIntegrationTest` function, but we could consider ways to make the function name more expressive. 🤔 

I think all of these castings to `int64` are unnecessary since the literals will be parsed as an `int64` in this context. May push up a separate commit to remove them.

ecmendenhall: The name is good! It just wasn't immediately visible in this diff.

PR #431, opened 2021-01-25
VDB 1801 fix all flaps query for multiple contracts

paytonrules:
Can I get a check on these tests?

### DIFF
--- a/db/migrations/20210126150526_update_get_flap_query.sql
+++ b/db/migrations/20210126150526_update_get_flap_query.sql
@@ -152,7 +152,7 @@ WITH address_id AS (
          WHERE deal.bid_id = get_flap.bid_id
            AND deal.address_id = (SELECT * FROM address_id)
            AND headers.block_number <= block_height
-         ORDER BY bid_id, block_number DESC
+         ORDER BY block_number DESC
          LIMIT 1
      )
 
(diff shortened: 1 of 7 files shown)


db/migrations/20210126150526_update_get_flap_query.sql
```diff
@@ -0,0 +1,187 @@
...
+         LIMIT 1
+     ),
+     deal AS (
+         SELECT deal, bid_id
+         FROM maker.deal
+                  LEFT JOIN public.headers ON deal.header_id = headers.id
+         WHERE deal.bid_id = get_flap_with_address.bid_id
+           AND deal.address_id = (SELECT * FROM address_id)
+           AND headers.block_number <= block_height
+         ORDER BY bid_id, block_number DESC
```
rmulhol, line 34: nbd but prob don't need to order by bid_id? I think there should only be one bid_id in the results set

rmulhol (approved):

db/migrations/20210126150526_update_get_flap_query.sql
paytonrules: 🤔 that was present in the original query so I didn't think too much about it, but I'm pretty sure you're right.

PR #436, opened 2021-01-27
(VDB-1803) Add indexes to places required for API

rmulhol:
- When Postgraphile is configured to not ignore indexes, it will not
  expose columns that are not indexed. These changes add indexes to
  columns that contain core data for the tables in question, so that
  they can be accessed via the API


paytonrules (approved):
Make sure you check the timestamp before merging.

paytonrules (approved):

ecmendenhall (approved):

PR #439, opened 2021-01-29
Add time series queries

ecmendenhall:
- Copy over time series queries from https://github.com/makerdao/vdb-mcd-transformers/pull/391


paytonrules:
Are there not tests for the time auction totals?

ecmendenhall:
Good catch @paytonrules. There are, they were merged in a separate PR. I just pulled them in.

paytonrules (approved):
I'm assuming not much, if any, code had to be changed so this has been reviewed before and we just needed to make sure everything got over.

PR #445, opened 2021-02-04
Add get_storage_tranformations query

paytonrules:
This query returns the minimum block and maximum block for a storage diff in a given state. So if you ask for
api.get_storage_transformations_for_status("new") you'll see the earliest storage diff (including its block number) and the latest (ergo most recent).

image


db/migrations/20210209134315_create_storage_transformations_for_status_query.sql
```diff
@@ -0,0 +1,44 @@
...
+	block_height 	BIGINT,
+	from_backfill 	BOOLEAN,
+	status 			public.diff_status,
+	storage_key 	BYTEA,
+	storage_value 	BYTEA,
+	created 		TIMESTAMP WITHOUT TIME ZONE,
+	updated 		TIMESTAMP WITHOUT TIME ZONE
+);
+
+CREATE OR REPLACE FUNCTION api.get_storage_transformations_for_status(status TEXT)
```
elizabethengelman: I wonder if we could call this function `get_min_and_max_diffs_by_status` or something to tell the user that this will be return a min and a max diff?

Also not sure if we should include "transformations" in the function name since it's possible that the diffs haven't been transformed, like for new diffs.

elizabethengelman (approved):

db/migrations/20210209134315_create_storage_transformations_for_status_query.sql
paytonrules: I'm competely fine with changing the name, as I've hated all my options.

db/migrations/20210209134315_create_min_and_max_diff_by_status_query.sql
```diff
@@ -0,0 +1,43 @@
...
+CREATE TYPE api.min_or_max AS ENUM (
+	'min',
+	'max'
+);
+
+CREATE TYPE api.storage_diff AS (
+	min_or_max 		api.min_or_max,
+	address 		BYTEA,
+	block_hash 		BYTEA,
+	block_height 	BIGINT,
```
ecmendenhall: Sorry to be a style stickler, but can we line up indentation here?

ecmendenhall (approved):
Looks good! Just left one style nitpick.

PR #448, opened 2021-02-10
New collaterals: UNIV2DAIUSDC-A, UNIV2ETHUSDT-A

yaoandrew:
This generates new configs for the UNIV2 LP collaterals for ETHUSDT and DAIUSDC


elizabethengelman:
We should make sure to add the new Flip contracts to the [FlipV110ABI method](https://github.com/makerdao/vdb-mcd-transformers/blob/prod/transformers/shared/constants/method.go#L57) - it's unfortunate that we don't have a failing test to indicate that this change should happen at the moment, so maybe we can figure out a way to add one.

The important reason to add the new contracts to this method is that we have a [check in GetABIFromContractsWithMatchingABI](https://github.com/makerdao/vdb-transformer-utilities/blob/main/pkg/shared/constants/external.go#L68) that makes sure that all of the contracts in that list have the same ABI. So, if any of the new Flip contracts have a different ABI that we didn't know about, this will catch it.

plugins/execute/transformerExporter.go
```diff
@@ -146,140 +148,141 @@ var Exporter exporter
 
 func (e exporter) Export() ([]event.TransformerInitializer, []storage.TransformerInitializer, []interface1.ContractTransformerInitializer) {
 	return []event.TransformerInitializer{
-			auction_file.EventTransformerInitializer,
```
elizabethengelman: This diff is a bit challenging to read, not sure why things were re-sorted so that they're not in alphabetical order. Not a big deal, but just want to verify that the only change in here is that the collateral generator added `flip_univ2ethusdt_a_v1_2_5.StorageTransformerInitializer` and `flip_univ2daiusdc_a_v1_2_5.StorageTransformerInitializer` to this the `storage.TransformerInitializer` slice. The `event.TransformerInitializer` slice shouldn't have changed.

elizabethengelman (commented):
We also need to add the new collateral's flip contracts to the [FlipV110ABI method](https://github.com/makerdao/vdb-mcd-transformers/blob/prod/transformers/shared/constants/method.go#L57) - unfortunately we don't have a failing test to indicate that this is necessary, but maybe we can find a way to add one.

The reason we need to add the new Flip contracts to that method is that it uses [GetABIFromContractsWithMatchingABI](https://github.com/makerdao/vdb-transformer-utilities/blob/main/pkg/shared/constants/external.go#L68), which checks to make sure that all of the contracts have the same ABI. This is helpful in case one of the contract's ABIs changed without us realizing, so we'd be able to ask about and handle the new ABI.

plugins/execute/transformerExporter.go
yaoandrew: 👍 yes the changes are to the `storage.TransformerInitializer ` slice. Good idea on adding a test on the  `FlipV110ABI()` method. That's super easy to miss.

elizabethengelman (approved):

ecmendenhall (approved):

PR #455, opened 2021-02-17
Configures backfillStorage for usdc usdt univ2

yaoandrew:


ecmendenhall (approved):

elizabethengelman (dismissed):
We'll also need to update the `dockerfiles/backfill_storage/startup_script.sh` to indicate the starting and ending block for the back file, as well as add the initializers to `plugins/backfill/storage/transformerExporter.go`.

PR #464, opened 2021-02-23
Add backfill storage config

yaoandrew:


elizabethengelman (changes requested):
`backfillStorage.toml` looks good. We'll also need to update the `./plugins/backfill/storage/transformerExporter.go` and
`./dockerfiles/backfill_storage/startup_script.sh` files.

elizabethengelman (approved):

PR #468, opened 2021-02-25
VDB-1845: dog bark transformer (Liq2.0)

elizabethengelman:
Opened this PR against a `liq-2.0` branch. 

We probably don't want to merge these liq 2.0 changes into `prod` until the contracts are deploy on mainnet since the config values are dummy data for now.  So I was thinking that having a `liq-2.0` feature branch may make this workflow easier - we could merge reviewed changes into `liq-2.0`, and then make sure to keep that branch up-to-date with `prod`. Then, once the mainnet deployment happens, we can update the dummy config with the real values and merge it into `prod`. Totally cool with a different approach too!


db/migrations/20210224083353_create_dog_bark.sql
```diff
@@ -0,0 +1,32 @@
+-- +goose Up
+CREATE TABLE maker.dog_bark
+(
+    id         SERIAL PRIMARY KEY,
+    log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    address_id BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    ilk_id     INTEGER NOT NULL REFERENCES maker.ilks (id) ON DELETE CASCADE,
+    urn_id     INTEGER NOT NULL REFERENCES maker.urns (id) ON DELETE CASCADE,
+    clip       BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
```
elizabethengelman, line 10: Was going back and forth on whether we wanted this to be a FK to the address table, or if we'd want to persist the clip address as a string. Curious what others think!

environments/mcdTransformers.toml
```diff
@@ -92,6 +92,13 @@
     rank = "0"
     repository = "github.com/makerdao/vdb-mcd-transformers"
     type = "eth_event"
+  [exporter.dog_bark]
+    contracts = ["MCD_DOG_1_x_x"]
```
elizabethengelman, line 96: Update this contract name once the Dog contract has been deployed.

environments/mcdTransformers.toml
```diff
@@ -910,6 +917,10 @@
...
     deployed = 10742907
+  [contract.MCD_DOG_1_x_x]
```
elizabethengelman, line 920: 🚧 We'll need to update these contract details once the Dog contract is deployed. To update:
- the contract name MCD_DOG_1_x_x
- the contract address
- the ABI - though this ABI is likely correct, we should use the ABI that is posted when the contract is deployed to make sure that the right complier version was used,etc
- deployed block number

transformers/integration_tests/dog_bark_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/makerdao/vdb-mcd-transformers/transformers/shared/constants"
+	"github.com/makerdao/vdb-mcd-transformers/transformers/test_data"
+	"github.com/makerdao/vulcanizedb/libraries/shared/factories/event"
+	"github.com/makerdao/vulcanizedb/pkg/core"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func fetchDogBarkLogsFromChain(config event.TransformerConfig, header core.Header) []core.EventLog {
+	//TODO: when there are real Dog Bark events on chain, use the following code to fetch real events from the chain
```
elizabethengelman, line 16: 🚧

transformers/shared/constants/method.go
```diff
@@ -26,6 +26,7 @@ import (
 func Cat100ABI() string     { return constants.GetContractABI("MCD_CAT_1_0_0") }
 func Cat110ABI() string     { return constants.GetContractABI("MCD_CAT_1_1_0") }
 func CdpManagerABI() string { return constants.GetContractABI("CDP_MANAGER") }
+func DogABI() string        { return constants.GetContractABI("MCD_DOG_1_x_x") }
```
elizabethengelman, line 29: We'll probably want to include the version in this method name, so that if a new Dog contract is deployed in the future, we'll already have the versioning in place.

transformers/test_data/config_values.go
```diff
@@ -9,6 +9,7 @@ import (
 
 func Cat100Address() string   { return checksum(constants.GetContractAddress("MCD_CAT_1_0_0")) }
 func Cat110Address() string   { return checksum(constants.GetContractAddress("MCD_CAT_1_1_0")) }
+func Dog1xxAddress() string   { return checksum(constants.GetContractAddress("MCD_DOG_1_x_x")) }
```
elizabethengelman, line 12: 🚧 Update this method name with the version when the contract is deployed.

transformers/test_data/dog_bark.go
```diff
@@ -0,0 +1,81 @@
...
+var (
+	ilkIdentifier   = "4554482d41"
+	urnAddress      = "0x7fa9ef6609ca7921112231f8f195138ebba2977"
+	id              = big.NewInt(32)
+	ink             = big.NewInt(123)
+	art             = big.NewInt(456)
+	due             = big.NewInt(789)
+	ClipAddress     = "0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB"
+	testBlockNumber = uint64(4)
+	RawDogBarkLog   = types.Log{
```
elizabethengelman, line 24: 🚧 Once the Dog contract is deployed, it's probably a good idea to grab a real Bark event for this test data instead of this fake data.

db/migrations/20210224083353_create_dog_bark.sql
yaoandrew: I think its more useful to see the actual address as a user, unless we're missing something else in postgraphile by not having it be a FK.

yaoandrew (approved):

ecmendenhall (approved):
Thanks for the parting gift, @elizabethengelman! 🎁

PR #479, opened 2021-03-08
Vdb 1855 dog deny event

yaoandrew:


transformers/integration_tests/dog_deny_test.go
```diff
@@ -0,0 +1,4 @@
+package integration_tests
+
+//TODO: add integration test when mainnet has data
```
ecmendenhall: Could you add a pending test here that will make some noise in the build, but won't fail? Not a huge deal, but better to have a reminder that runs than a comment.

transformers/test_data/dog_deny.go
```diff
@@ -0,0 +1,54 @@
...
+		OrderedColumns: []event.ColumnName{
+			event.HeaderFK,
+			event.LogFK,
+			event.AddressFK,
+			constants.UsrColumn,
+		},
+		ColumnValues: event.ColumnValues{
+			event.HeaderFK: DogDenyEventLog.HeaderID,
+			event.LogFK:    DogDenyEventLog.ID,
+			//event.AddressFK,
```
ecmendenhall, line 48: What's up with these? Not known until we have better test data?

ecmendenhall (approved):

transformers/test_data/dog_deny.go
yaoandrew: No, we don't know them since they are database values (addressID) so we don't know them until the test runs.

transformers/integration_tests/dog_deny_test.go
yaoandrew: Sure... I think you can add the framework for it and us `XIt` to mark it pending. As long as it compiles.

PR #482, opened 2021-03-17
Update api.flip_bid_event_bid to use event block

ecmendenhall:
Embedded `bid`s inside `bidEvents` are showing the most recent state snapshot instead of the historical snapshot at the time of the event. Passing the event's block height to `api.get_flip_with_address` should fix this up.


db/migrations/20210317095927_add_block_height_to_flip_bid_event_bid.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+CREATE OR REPLACE FUNCTION api.flip_bid_event_bid(event api.flip_bid_event)
+    RETURNS api.flip_bid_snapshot AS
+$$
+SELECT * FROM api.get_flip_with_address(event.bid_id, event.contract_address, event.block_height)
```
yaoandrew, line 5: 🎉

yaoandrew (approved):

PR #484, opened 2021-03-19
Vdb 1861 clip kick

yaoandrew:


transformers/component_tests/initializers/events/event_initializers_test.go
```diff
@@ -131,6 +132,14 @@ var _ = Describe("Event transformer initializers", func() {
 		assertCorrectInit(transformer, name, topic, 8928165, catContracts())
 	})
 
+	It("configures clip kick", func() {
+		transformer := clip_kick.EventTransformerInitializer(db)
+		name := "clip_kick"
+		topic := "0x898eb267e6ff99efebaf458079f659c812f378fe8cf2a33ab4028717672bd021"
```
ecmendenhall, line 140: Just for my own understanding, which components here are real and which are fake? Looks like this is a real topic, discovered by running the contracts locally, but the address and block are faked?

ecmendenhall (approved):
📎 🦵 ⚽

transformers/component_tests/initializers/events/event_initializers_test.go
yaoandrew: That's right.. Those are to be replaced with real values when available on `mainnet`.  I was trying to make them look really fake 🤣

db/migrations/20210318214325_create_clip_kick.sql
```diff
@@ -0,0 +1,33 @@
...
+(
+    id         SERIAL PRIMARY KEY,
+    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    address_id BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    bid_id     NUMERIC NOT NULL,
+    top        NUMERIC,
+    tab        NUMERIC,
+    lot        NUMERIC,
+    usr        TEXT,
```
yaoandrew: `usr` and `kpr` should probably be `BIGINT` here in keeping with the convention of having addresses stored as IDs.

PR #488, opened 2021-03-26
Vdb 1862 clip take

yaoandrew:


db/migrations/20210325223733_create_clip_take.sql
```diff
@@ -0,0 +1,33 @@
+-- +goose Up
+CREATE TABLE maker.clip_take
+(
+    id         SERIAL PRIMARY KEY,
+    header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
+    address_id BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
+    log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
+    clip_id    NUMERIC,
```
ecmendenhall: Is this the `id` argument from the take event? If so, I think this is the equivalent of what we call `bidId` in Liq-1.x. If that's the case, I think we'll want to call this `sale_id`: it's the ID of this `Sale` in the array [here](https://github.com/makerdao/dss/blob/4cc80e9682debb6de047235d3d69c117d7536d45/src/clip.sol#L86).

ecmendenhall: (I don't think I caught this on earlier events, but we should check those out too)

ecmendenhall (approved):

PR #495, opened 2021-04-21
Add mainnet contract for link_a clip and dog

yaoandrew:


transformers/test_data/config_values.go
```diff
@@ -9,8 +9,8 @@ import (
 
 func Cat100Address() string   { return checksum(constants.GetContractAddress("MCD_CAT_1_0_0")) }
 func Cat110Address() string   { return checksum(constants.GetContractAddress("MCD_CAT_1_1_0")) }
-func Clip1xxAddress() string  { return checksum(constants.GetContractAddress("MCD_CLIP_1_x_x")) }
-func Dog1xxAddress() string   { return checksum(constants.GetContractAddress("MCD_DOG_1_x_x")) }
+func Clip130Address() string  { return checksum(constants.GetContractAddress("MCD_CLIP_LINK_A_1_3_0")) }
```
ecmendenhall, line 12: I'm cool with this for now, but this will have to become plural pretty soon, right? Something like `FlipV100Addresses` below.

ecmendenhall (approved):

transformers/test_data/config_values.go
yaoandrew: yes.. we'll also need to refactor it to use the `getContractAddresses` helper function as we add more collateral types

PR #497, opened 2021-05-03
Add integration tests dog clip events

yaoandrew:


db/migrations/20210324093804_create_dog_file_vow.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE maker.dog_file_vow
     log_id     BIGINT  NOT NULL REFERENCES public.event_logs (id) ON DELETE CASCADE,
     address_id BIGINT  NOT NULL REFERENCES public.addresses (id) ON DELETE CASCADE,
     header_id  INTEGER NOT NULL REFERENCES public.headers (id) ON DELETE CASCADE,
-    what       BIGINT,
+    what       TEXT,
```
ecmendenhall: Has this already been applied in QA? If so, you probably need a migration to change it. If there's no data in there, you should be able to modify the column.

ecmendenhall (approved):
Not sure you can get away with modifying that migration, but otherwise looks good.

db/migrations/20210324093804_create_dog_file_vow.sql
yaoandrew: I don't think there was ever data in there since the test wasn't working

PR #499, opened 2021-05-05
Backfill dog clip events

yaoandrew:


plugins/backfill/transformerExporter.go
```diff
@@ -13,7 +24,19 @@ type exporter string
...
+			clip_kick.EventTransformerInitializer,
+			clip_redo.EventTransformerInitializer,
+			clip_take.EventTransformerInitializer,
+			clip_yank.EventTransformerInitializer,
+			dog_deny.EventTransformerInitializer,
+			dog_digs.EventTransformerInitializer,
+			dog_rely.EventTransformerInitializer,
+			dog_file_ilk_clip.EventTransformerInitializer,
+			dog_file_hole.EventTransformerInitializer,
+			dog_file_ilk_chop_hole.EventTransformerInitializer,
```
ecmendenhall, line 37: Weird formatting here...

ecmendenhall (changes requested):
Make sure you also update `dockerfiles/backfill_events/startup_script.sh` with the ending block for the backfill. Otherwise looks good.

plugins/backfill/transformerExporter.go
yaoandrew: yeah.. strange. In goland its lined up under the `[ ]`. I'll see if its my tab setting.

ecmendenhall (approved):

dockerfiles/backfill_events/startup_script.sh
```diff
@@ -4,4 +4,4 @@
 # In other words, just after the block number where the new collateral(s) was added to execute.
 echo "Starting event backfill to block [0]"
```
ecmendenhall, line 5: I missed this earlier, but it's nice to update this so that the end point shows up in the logs. Would be best to extract this to a variable and update in a single place.

PR #513, opened 2021-05-26
Removes vat from keys loader

yaoandrew:


ecmendenhall (approved):
Looks good—this is all because `vat` is immutable, right?

yaoandrew:
> Looks good—this is all because `vat` is immutable, right?

Yes, so everything shifts by one

PR #515, opened 2021-06-09
Clip storage static values

yaoandrew:
Adds necessary keys and repository logic to store static clip storage values

Needs to have https://github.com/makerdao/vulcanizedb/pull/189 merged in first.


yaoandrew:
Merging to support backfilling new data and collateral

PR #517, opened 2021-06-11
New clip collateral configs

yaoandrew:
- Adds storage transformer initializers and tests for all clip collateral
- Adds new collateral to clip events config
- Adds all clip collateral contract details to extractDiffs config
- Adds storage transformer initializers to plugin execute transformerExporter


plugins/backfill/transformerExporter.go
```diff
@@ -15,8 +14,6 @@ var Exporter exporter
 
 func (e exporter) Export() ([]event.TransformerInitializer, []storage.TransformerInitializer, []interface1.ContractTransformerInitializer) {
 	return []event.TransformerInitializer{},
-		[]storage.TransformerInitializer{
-			dog_v1_3_0.StorageTransformerInitializer,
```
ecmendenhall, line 19: This backfill also bumped into the throttling issue, right? Let's make sure we're tracking all the outstanding backfill work somewhere, so we don't miss it.

ecmendenhall (approved):
Looks good. Time to update the generator :smile:

PR #521, opened 2021-06-16
Remove active from storage

yaoandrew:
- Clean up migrations
- Remove active field from key loader

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -8217,39 +8217,6 @@ ALTER SEQUENCE maker.clip_dog_id_seq OWNED BY maker.clip_dog.id;
 
 
 --
--- Name: clip_ilk; Type: TABLE; Schema: maker; Owner: -
---
-
-CREATE TABLE maker.clip_ilk (
-    id integer NOT NULL,
-    diff_id bigint NOT NULL,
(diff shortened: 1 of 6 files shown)


db/schema.sql
```diff
@@ -8216,39 +8216,6 @@ CREATE SEQUENCE maker.clip_dog_id_seq
 ALTER SEQUENCE maker.clip_dog_id_seq OWNED BY maker.clip_dog.id;
 
 
---
--- Name: clip_ilk; Type: TABLE; Schema: maker; Owner: -
---
-
-CREATE TABLE maker.clip_ilk (
```
ecmendenhall, line 8223: Looks like `clip_ilk` disappeared from the schema. Is that expected? I don't see any changes in here related to that table.

ecmendenhall (approved):
One question on why `clip_ilk` is no longer in the schema file. Otherwise looks good.

db/schema.sql
yaoandrew: That might be the result of my change from last time around. vat and ilk are immutable so not necessary to keep track of. I had them in the schema with earlier static fields.

PR #527, opened 2021-08-12
Clips storage table

yaoandrew:
- Adds clip event types
- Adds clip event snapshot 
- Adds computed columns
- Adds `all_clips` query using addresses


db/migrations/20210803131922_create_clip_sale_event_computed_columns.sql
```diff
@@ -0,0 +1,102 @@
+-- +goose Up
+
+CREATE FUNCTION api.clip_sale_event_tx(event api.clip_sale_event) RETURNS SETOF api.tx
```
ecmendenhall, line 3: Don't need to fix this now, but we should look into whether we can drop the `SETOF` here to prevent Postgraphile from paginating this computed column. I'm always annoyed when I have to ask for the `first: 1` transaction in GraphQL in a context where there's only a single associated tx.

db/migrations/20210803131922_create_clip_sale_event_computed_columns.sql
```diff
@@ -0,0 +1,102 @@
...
+       storage_values.tic,
+       storage_values.top,
+       storage_values.created,
+       storage_values.updated,
+       get_clip_with_address.clip_address
+FROM storage_values
+$$;
+
+--
+-- Name: clip_sale_event_bid(api.clip_sale_event); Type: FUNCTION; Schema: api; Owner: -
```
ecmendenhall: Looks like a copy-paste  from the schema here—can you give all the migrations a pass and remove any of these stray comments?

transformers/component_tests/queries/get_clip_query_test.go
```diff
@@ -0,0 +1,66 @@
...
+		clipKickEvent.ColumnValues[event.HeaderFK] = headerOne.Id
+		clipKickEvent.ColumnValues[event.LogFK] = clipKickLog.ID
+		clipKickEvent.ColumnValues[event.AddressFK] = addressId
+		clipKickEvent.ColumnValues[constants.SaleIDColumn] = strconv.Itoa(fakeSaleId)
+		clipKickErr := event.PersistModels([]event.InsertionModel{clipKickEvent}, db)
+		Expect(clipKickErr).NotTo(HaveOccurred())
+	})
+
+	Context("get_clip_with_address", func() {
+		It("gets only the specified clip", func() {
```
ecmendenhall, line 77: Seems like this test is missing a second clip with the same sale ID from another address. Should we create a second clip as part of this test and verify that we're only retrieving the one we expect?

ecmendenhall (changes requested):
Looking good! A couple minor comments.

transformers/component_tests/queries/all_clip_sale_events_test.go
```diff
@@ -0,0 +1,374 @@
...
+				test_helpers.SaleEvent{
+					SaleId: strconv.Itoa(saleId),
+					Act:    "kick"},
+				test_helpers.SaleEvent{
+					SaleId: clipKickEventTwo.ColumnValues["sale_id"].(string),
+					Act:    "kick"},
+			))
+		})
+
+		It("returns sale events from different kinds of clippers (clips with different contract addresses", func() {
```
ecmendenhall: Tiny nitpick :)

```suggestion
		It("returns sale events from different kinds of clippers (clips with different contract addresses)", func() {
```

ecmendenhall (approved):

PR #537, opened 2021-09-29
Bugfix clip active array

yaoandrew:
- Fixes a previous bug in decoding storage diffs on the active field 
- Adds logic to create keys for active array elements

### DIFF
--- a/transformers/storage/repository.go
+++ b/transformers/storage/repository.go
@@ -17,6 +17,7 @@
 package storage
 
 import (
+	"database/sql"
 	"math"
 	"strconv"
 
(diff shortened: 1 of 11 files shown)


transformers/storage/repository.go
```diff
@@ -66,6 +68,19 @@ func (repository *MakerStorageRepository) GetClipSalesIDs(contractAddress string
 	return saleIDs, err
 }
 
+func (repository *MakerStorageRepository) GetActiveLength(contractAddress string) (string, error) {
+	var activeLength sql.NullString
+	addressID, addressErr := repository.GetOrCreateAddress(contractAddress)
+	if addressErr != nil {
+		return "", addressErr
```
yaoandrew: Not sure how I feel about returning the empty string here in the case of an address error.

ecmendenhall: Does zero make more sense?

transformers/storage/repository.go
```diff
@@ -66,6 +68,19 @@ func (repository *MakerStorageRepository) GetClipSalesIDs(contractAddress string
 	return saleIDs, err
 }
 
+func (repository *MakerStorageRepository) GetActiveLength(contractAddress string) (string, error) {
```
ecmendenhall: Nitpick: This is actually `GetMaxActiveLength`

ecmendenhall (approved):

PR #550, opened 2021-10-14
Add config for storage backfill

yaoandrew:


ecmendenhall (approved):

dockerfiles/backfill_storage/startup_script.sh
```diff
@@ -28,5 +28,5 @@ if test -z "$VDB_PG_CONNECT"; then
 fi
 
 # Run backfillStorage
-echo "Running backfillStorage from block 0 to 0"
-./vulcanizedb backfillStorage -s=0  -e=0 --config config.toml
+echo "Running backfillStorage from block 12464984 to 13379859"
+./vulcanizedb backfillStorage -s=12464984 -e=13379859 --config config.toml
```
ecmendenhall, line 32: Huh, I see that this restarted, but it looks like the previous backfill ended at `12394815`. I think we should be OK to start there.

PR #558, opened 2021-11-22
Adds configuration for wbtc-b

yaoandrew:
- Needs deployment block updated when available
- Potentially needs version updated when released


environments/backfillStorage.toml
```diff
@@ -142,6 +142,12 @@
         rank = "0"
         repository = "github.com/makerdao/vdb-mcd-transformers"
         type = "eth_storage"
+    [exporter.clip_wbtc_a_v1_5_0]
```
ecmendenhall: Looks like a mismatch here.

ecmendenhall (approved):

PR #564, opened 2021-12-02
Add WBTC-C support

dizzy:


dockerfiles/backfill_storage/startup_script.sh
```diff
@@ -28,5 +28,5 @@ if test -z "$VDB_PG_CONNECT"; then
 fi
 
 # Run backfillStorage
-echo "Running backfillStorage from block 12349311 to 13425398"
-./vulcanizedb backfillStorage -s=12349311 -e=13425398 --config config.toml
+echo "Running backfillStorage from block 12353406 to 13425398"
```
yaoandrew: Should we just wait on this and use the [parameter based config](https://github.com/makerdao/vdb-mcd-transformers/commit/bfeeb7bb4828b035f1aa2ccb1aa24f749bdeea9a) once we've merged all these new collateral PRs?

yaoandrew (approved):
Looks good to me besides the question below and might need a rebase.

dockerfiles/backfill_storage/startup_script.sh
dizzy: Yeah, well actually after rebase this should be replaced with the parameter based config, so I will merge all new collateral support and then we can restart backfill from a common block

PR #570, opened 2021-12-08
Update prod

dizzy:


yaoandrew (approved):
Looking good. I'm assuming we can have the configs in the backfills since we're controlling them via ENV vars in AWS?

dizzy:
> Looking good. I'm assuming we can have the configs in the backfills since we're controlling them via ENV vars in AWS?

yep, that's right, backfills won't start on merge, instead we're going to manually start them from the desired block

ecmendenhall (approved):

PR #580, opened 2021-12-22
Use Vat storage transformer without urn and dai keys 

yaoandrew:
- Run the backfill with a reduced set of keys in the Vat contract
- The key count prior to removal was over 120K
- We can reduce this to 13K by removing urn keys
- Further reduced to 7K with dai keys removed

Urns that have frobs we care about were already updated with a separate Urn backfill process.


plugins/backfill/storage/transformerExporter.go
```diff
@@ -3,60 +3,7 @@
...
-	median_uni_v1_2_1 "github.com/makerdao/vdb-mcd-transformers/transformers/storage/median/initializers/median_uni/v1_2_1"
-	median_usdt_v1_0_4 "github.com/makerdao/vdb-mcd-transformers/transformers/storage/median/initializers/median_usdt/v1_0_4"
-	median_wbtc_v1_0_6 "github.com/makerdao/vdb-mcd-transformers/transformers/storage/median/initializers/median_wbtc/v1_0_6"
-	median_wsteth_v1_9_8 "github.com/makerdao/vdb-mcd-transformers/transformers/storage/median/initializers/median_wsteth/v1_9_8"
-	median_yfi_v1_1_14 "github.com/makerdao/vdb-mcd-transformers/transformers/storage/median/initializers/median_yfi/v1_1_14"
-	median_zrx_v1_0_8 "github.com/makerdao/vdb-mcd-transformers/transformers/storage/median/initializers/median_zrx/v1_0_8"
-	pot "github.com/makerdao/vdb-mcd-transformers/transformers/storage/pot/initializer"
-	spot "github.com/makerdao/vdb-mcd-transformers/transformers/storage/spot/initializer"
-	vow "github.com/makerdao/vdb-mcd-transformers/transformers/storage/vow/initializer"
+	vat "github.com/makerdao/vdb-mcd-transformers/transformers/storage/vat/initializer"
```
ecmendenhall, line 6: Can we create this as a new and separate transformer rather than modifying the existing one? Essentially, copy-paste the existing `vat` transformer to `vat_lite`, remove the storage keys, and treat it as a separate module. It could exist in parallel to the standard `vat` transformer and would only be used in this specific backfill situation.

